PR #151, opened 2015-01-24
! all: improve the metric recorders infrastructure

ivantopo:
This branch contains all the work I have been doing lately with regards to making the metric recorders easier to create and use as well as trying to keep the Kamon public API simple and non Akka dependant. More enhancements are in the oven :).


kamon-core/src/main/resources/reference.conf
```diff
@@ -142,7 +144,7 @@ kamon {
     }
 
     # Default dispatcher for all trace module operations
-    dispatcher = ${kamon.default-dispatcher}
+    dispatcher = "akka.actor.default-dispatcher"
```
dpsoft: I think that a good idea could be use kamon.default-dispatcher for some modules

kamon-core/src/main/scala/kamon/ModuleSupervisor.scala
```diff
@@ -0,0 +1,49 @@
...
+
+trait ModuleSupervisorExtension extends actor.Extension {
+  def createModule(name: String, props: Props): Future[ActorRef]
+}
+
+class ModuleSupervisorExtensionImpl(system: ExtendedActorSystem) extends ModuleSupervisorExtension {
+  import system.dispatcher
+  private val supervisor = system.actorOf(Props[ModuleSupervisor], "kamon")
+
+  def createModule(name: String, props: Props): Future[ActorRef] = Future {} flatMap { _: Unit ⇒
```
dpsoft: why :Future[ActorRef] = Future{} flatMap { _: Unit => instead of Future[ActorRef] ?

kamon-core/src/main/scala/kamon/util/FastDispatch.scala
```diff
@@ -0,0 +1,22 @@
...
+import akka.actor.ActorRef
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+/**
+ *  Extension for Future[ActorRef]. Try to dispatch a message to a Future[ActorRef] in the same thread if it has already
+ *  completed or do the regular scheduling otherwise. Specially useful when using the ModuleSupervisor extension to
+ *  create actors.
+ */
+object FastDispatch {
```
dpsoft: I really like the idea!!!!!

kamon-system-metrics/src/main/scala/kamon/system/sigar/FileSystemMetrics.scala
```diff
@@ -0,0 +1,25 @@
...
+import kamon.metric.GenericEntityRecorder
+import kamon.metric.instrument.{ Memory, InstrumentFactory }
+import org.hyperic.sigar.{ DiskUsage, FileSystem, Sigar }
+
+class FileSystemMetrics(instrumentFactory: InstrumentFactory) extends GenericEntityRecorder(instrumentFactory) with SigarMetric {
+  val reads = DiffRecordingHistogram(histogram("file-system-reads", Memory.Bytes))
+  val writes = DiffRecordingHistogram(histogram("file-system-writes", Memory.Bytes))
+
+  def sumOfAllFileSystems(sigar: Sigar, thunk: DiskUsage ⇒ Long): Long = {
+    val fileSystems = sigar.getFileSystemList.filter(_.getType == FileSystem.TYPE_LOCAL_DISK).map(_.getDevName).toSet
```
dpsoft: I dont know if is necessary call always to sigar.getFileSystem List, I mean maybe with the first call is enough, unless we want to update all the time the info of the devices, just in case that appear new one in runtime. what you think?

kamon-system-metrics/src/main/scala/kamon/system/sigar/NetworkMetrics.scala
```diff
@@ -0,0 +1,33 @@
...
+class NetworkMetrics(instrumentFactory: InstrumentFactory) extends GenericEntityRecorder(instrumentFactory) with SigarMetric {
+  val receivedBytes = DiffRecordingHistogram(histogram("rx-bytes", Memory.Bytes))
+  val transmittedBytes = DiffRecordingHistogram(histogram("tx-bytes", Memory.Bytes))
+  val receiveErrors = DiffRecordingHistogram(histogram("rx-errors"))
+  val transmitErrors = DiffRecordingHistogram(histogram("tx-errors"))
+  val receiveDrops = DiffRecordingHistogram(histogram("rx-dropped"))
+  val transmitDrops = DiffRecordingHistogram(histogram("tx-dropped"))
+
+  def sumOfAllInterfaces(sigar: Sigar, thunk: NetInterfaceStat ⇒ Long): Long = {
+    val interfaces = sigar.getNetInterfaceList.toList.filter(_ != "lo")
```
dpsoft: same as above(FileSystemMetrics)

dpsoft:
Man BTW: 

Groso

kamon-core/src/main/resources/reference.conf
ivantopo: Right, I think we already discussed that before and I just forgot.. will introduce a small dispatcher config just for Kamon.

kamon-core/src/main/scala/kamon/ModuleSupervisor.scala
ivantopo: I need to review this part again. There is a problem with the initialization order when using the Akka instrumentation, because the Dispatchers instrumentation might want to lookup the metrics extension before the ActorSystem is fully constructed and, the metrics extension wants the ModuleSupervisor to create the subscriptions actor, which can't be created because the actor system isn't fully constructed yet, causing a sad deadlock :(... 

I guess that we will need to separate the creation of the Metrics extension from the creation of the subscriptions dispatcher actor, so that the metrics extension is "akka free" and we don't have issues creating it at any point.

kamon-core/src/main/scala/kamon/util/FastDispatch.scala
ivantopo: (y)

kamon-system-metrics/src/main/scala/kamon/system/sigar/FileSystemMetrics.scala
ivantopo: yeah, probably the right way would be to change the Sigar-based metric recorders to receive the sigar instance in their constructor arguments, initialize whatever they might need to initialize and record the values con calls to .update.. WDYT?

kamon-system-metrics/src/main/scala/kamon/system/sigar/NetworkMetrics.scala
ivantopo: nice, same solution would apply here.

dpsoft:
LGTM!!!!!

Issue #126, opened 2014-12-04
create some sugar to start/finish segments


ivantopo:
We should provide some syntax sugar for creating blocking and async segments, so that people don't need to handle and finish the segment, at least not for the happy case. The following comes to mi mind:

``` scala
TraceRecorder.blockingSegment[T](name, category, library)(f: => T): T
TraceRecorder.asyncSegment[T](name, category, library)(f: => Future[T]): Future[T]
```

Maybe some builders for segments recorders could be nice, allowing the user to do something along the lines of:

``` scala
object Segments {
  val GetUsers = TraceRecorder.blockingSegmentBuilder("get-users", "jdbc", "xxx")
}

// and in the actual call site
Segments.GetUsers.measure {
  ....
}
```

open to suggestions!

ivantopo:
I'm closing this issue because it is either too old to be relevant or stale. Stop be our [Github Discussions](https://github.com/kamon-io/Kamon/discussions) section and drop a message if you need any additional help with the topic.

Have a nice day!

PR #156, opened 2015-02-14
Single kamon instance per jvm

ivantopo:
After this PR, there will no longer be a Kamon instance anywhere. There will be only one instance wrapped in the `kamon.Kamon` companion object and such instance needs to be `.start(..)`ed to work properly.

By having a single place to look for a Kamon instance we can simplify our current instrumentation a bit, as well as opening the door for more interesting possibilities, such as the `kamon-annotation` module, richer APIs for our tracer (see #126).


dpsoft:
LGTM!!!!!

PR #158, opened 2015-02-20
Kamon annotation

dpsoft:
Almost complete implementation of kamon-annotation module.


kamon-annotation/src/main/java/kamon/annotation/Metrics.java
```diff
@@ -0,0 +1,12 @@
+package kamon.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE})
+public @interface Metrics {
```
dpsoft: We really need change this name!!!!!!

kamon-annotation/src/main/scala/kamon/annotation/instrumentation/BaseAnnotationInstrumentation.scala
```diff
@@ -0,0 +1,180 @@
...
+  private[this] def methodName(method: Method): String = method.toString.replace(" ", "-").toLowerCase
+}
+
+@Aspect
+class ClassToProfiledMixin {
+  @DeclareMixin("(@kamon.annotation.Metrics *)")
+  def mixinClassToProfiled: Profiled = new Profiled {}
+}
+
+trait Profiled {
```
dpsoft: We really need change this name!!!!!!

kamon-annotation/src/main/scala/kamon/annotation/instrumentation/BaseAnnotationInstrumentation.scala
```diff
@@ -0,0 +1,180 @@
...
+    }
+  }
+
+  @inline final def registerSegment(method: Method, segments: TrieMap[String, SegmentInfo])(implicit eval: StringEvaluator, tags: TagsEvaluator): Unit = {
+    if (method.isAnnotationPresent(classOf[Segment])) {
+      val segment = method.getAnnotation(classOf[Segment])
+      val name = eval(segment.name())
+      val category = eval(segment.category())
+      val library = eval(segment.library())
+      val metadata = tags(segment.tags())
```
dpsoft: we need use 'tag' instead of 'metadata'

kamon-annotation/src/main/java/kamon/annotation/Metrics.java
ivantopo: What about `@EnableKamonAnnotations`?

kamon-annotation/src/main/scala/kamon/annotation/instrumentation/BaseAnnotationInstrumentation.scala
```diff
@@ -0,0 +1,176 @@
...
+      counters.put(methodName(method), counter)
+    }
+  }
+
+  @inline final def registerTrace(method: Method, traces: TrieMap[String, TraceContext])(implicit eval: StringEvaluator): Unit = {
+    if (method.isAnnotationPresent(classOf[Trace])) {
+      val count = method.getAnnotation(classOf[Trace])
+      val name = eval(count.value())
+      val tags = evalTags(count.tags())
+      val trace = Kamon.tracer.newContext(name)
```
ivantopo: Here we are registering a single TraceContext for all invocations of the `@Trace`'d method and that breaks the principle of having a single TraceContext per each separate invocation of a given functionality. Instead of doing this, we should be always creating a new TraceContext when calling `processTrace`.

kamon-annotation/src/main/scala/kamon/annotation/instrumentation/BaseAnnotationInstrumentation.scala
ivantopo: What about `AnnotationInstruments`? Also, as pointed out above, the traces member should not map to a TraceContext. I guess the reason why this works is that we are not guarding against finishing the same TraceContext multiple times, do you think we should?

Finally, taking processHistogram as an example: it is calling histograms.get(name).map(...), but since we are registering the required instruments upon the creation of the annotated class instance, we could just do histograms.apply(..) and avoid mapping over the Option[Histogram]. If the required instrument is not available in the map it means we have a bug to fix or that the class wasn't annotated with `@EnableKamonAnnotations`, both cases would need manual intervention to get fixed.

kamon-annotation/src/main/scala/kamon/annotation/util/ELProcessorPool.scala
```diff
@@ -0,0 +1,60 @@
...
+import kamon.Kamon
+import kamon.annotation.Annotation
+import kamon.annotation.resolver.PrivateFieldELResolver
+import kamon.annotation.util.FastObjectPool.PoolFactory
+
+/**
+ * Convenient pool of @see ELProcessor, since it is not thread safe and the creation is expensive.
+ */
+object ELProcessorPool {
+  private val pool = new FastObjectPool[ELProcessor](ELPoolFactory(), Kamon(Annotation).poolSize)
```
ivantopo: What about just having a ThreadLocal[ELProcessor]? Having a too conservative size for the object pool can generate unnecessary blocks in the code... I'm not entirely sure about introducing the object pool. what do you thing? what would be the pros/cons of using a simple ThreadLocal vs FastObjectPool?

kamon-annotation/src/main/java/kamon/annotation/Histogram.java
```diff
@@ -0,0 +1,90 @@
...
+    int  precision() default 2;
+
+
+    /**
+     * Tags are a way of adding dimensions to metrics,
+     * these are constructed using EL syntax e.g. "${{'algorithm':'1','env':'production'}}"
+     *
+     * @return the tags associated to the histogram
+     */
+    String tags() default "";
```
ivantopo: I don't find myself very attracted by the `'key':'value'` syntax.. is that something specific to EL? can we just use a plain `key=value,otherkey=othervalue` string? in addition to the EL version, of course.

kamon-annotation/src/main/scala/kamon/annotation/instrumentation/AnnotationInstrumentation.scala
```diff
@@ -0,0 +1,63 @@
...
+
+    profiled.getClass.getDeclaredMethods.filterNot(method ⇒ Modifier.isStatic(method.getModifiers)).foreach {
+      method ⇒
+        registerTrace(method, profiled.traces)
+        registerSegment(method, profiled.segments)
+        registerCounter(method, profiled.counters)
+        registerMinMaxCounter(method, profiled.minMaxCounters)
+        registerHistogram(method, profiled.histograms)
+        registerTime(method, profiled.histograms)
+    }
```
ivantopo: How do registrations work with overloaded methods?

kamon-annotation/src/test/java/kamon/annotation/AnnotatedJavaClass.java
```diff
@@ -0,0 +1,65 @@
...
+public class AnnotatedJavaClass {
+
+    public static String ID = "10";
+
+    @Trace("trace")
+    public static void trace() {}
+
+    @Trace("trace-with-segment")
+    @Segment(name = "segment", category = "segments", library = "segment")
+    public static void segment() {
```
ivantopo: I'm not sure of what is the case of having both `@Trace` and `@Segment` in the same method, can you tell me a bit more about it?

ivantopo:
@dpsoft nice work bro! There are a few things to improve but it's taking a very interesting shape :D. Looking forward to your comments.

kamon-annotation/src/main/java/kamon/annotation/Metrics.java
dpsoft: Yes i like it!!!

kamon-annotation/src/main/scala/kamon/annotation/instrumentation/BaseAnnotationInstrumentation.scala
dpsoft: Ok, let's go with `Annotation Instruments`. and I agree completely with avoid the Option.

kamon-annotation/src/main/scala/kamon/annotation/instrumentation/BaseAnnotationInstrumentation.scala
dpsoft: yes bro, you are right!!!

kamon-annotation/src/main/scala/kamon/annotation/instrumentation/AnnotationInstrumentation.scala
dpsoft: I take the absolute method name, for example: `public-static-long-kamon.annotation.annotatedjavaclass.histogram(java.lang.long)` then an overloaded version could be  `public-static-long-kamon.annotation.annotatedjavaclass.histogram(java.lang.long, java.lang.String)` or `public-static-double-kamon.annotation.annotatedjavaclass.histogram(java.lang.long, java.lang.String)` this is done inside of register* methods.

kamon-annotation/src/main/java/kamon/annotation/Histogram.java
dpsoft: yes, the sintax to create a Map in EL, maybe we can accept  `key=value,otherkey=othervalue` and then parse the expression to EL signature?

kamon-annotation/src/test/java/kamon/annotation/AnnotatedJavaClass.java
dpsoft: That code is not a valid use case, I will remove it.

kamon-annotation/src/main/scala/kamon/annotation/util/ELProcessorPool.scala
dpsoft: I thought the pool was a good idea in the previous approach where always the invocation of a method were shooting the creation of a new ELContext. In the current context is not necessary, creating a new ELContext when the object is created is enough since the lifecycle is really short.

kamon-annotation/src/main/java/kamon/annotation/Metrics.java
ivantopo: let's move to the shorter `@EnableKamon`.

kamon-annotation/src/main/scala/kamon/annotation/instrumentation/BaseAnnotationInstrumentation.scala
```diff
@@ -0,0 +1,180 @@
...
+      segments.put(methodName(method), SegmentInfo(name, category, library, tags))
+    }
+  }
+
+  @inline final def processTrace(traces: TrieMap[String, TraceContextInfo], pjp: ProceedingJoinPoint): AnyRef = {
+    val name = methodName(pjp.getStaticPart.getSignature.asInstanceOf[MethodSignature].getMethod)
+    val traceInfo = traces(name)
+
+    Tracer.withContext(Kamon.tracer.newContext(traceInfo.name)) {
+      traceInfo.tags.foreach { case (key, value) ⇒ Tracer.currentContext.addMetadata(key, value) }
```
ivantopo: We need a few changes in the TraceContext for this to work properly. Currently we only have the concept of metadata but the TC should allow you to specify both tags and metadata. The tags are used for the metrics side and should be stable values that will be repeated in many traces, on the other side, metadata should be used to gather specific information about a single trace context and will be of interest for people currently using simple trace subscriptions and when we actually get to distributed traces.

This needs to wait a bit until the actual support for tags gets merged to master.

kamon-annotation/src/main/java/kamon/annotation/Histogram.java
ivantopo: how would that look like?

kamon-annotation/src/main/scala/kamon/annotation/instrumentation/AnnotationInstrumentation.scala
ivantopo: nice!

dpsoft:
Merged!!!!!!

PR #159, opened 2015-02-25
= play: Compile pattern for path normalization only once

magro:
Just a tiny performance optimization for play RequestInstrumentation.

### DIFF
--- a/kamon-play/src/main/scala/kamon/play/instrumentation/RequestInstrumentation.scala
+++ b/kamon-play/src/main/scala/kamon/play/instrumentation/RequestInstrumentation.scala
@@ -106,12 +106,14 @@ object RequestInstrumentation {
 
   private val cache = TrieMap.empty[String, String]
 
+  private val normalizePattern = """\$([^<]+)<[^>]+>""".r
+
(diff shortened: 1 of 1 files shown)


ivantopo:
Hey Martin, thanks for proposing this! I just need to ask you to change the commit message to something like `= play: compile pattern for path normalization only once`, just to match our commit message conventions. 

Also, we were discussing the possibility of not giving any default name to traces, other than "UnnamedTrace" to force people into provide their own trace names and as a consequence, we would remove the precise piece of code you are modifying now.. what are your thoughts as a Play and Kamon user on that approach?

The main motivation for that was initiated in our Spray module, because there we don't have the route patterns to deduce a decent trace name and we are currently just using `METHOD - request.path` as the default trace name. This of course can be problematic if you have ids in your path and can generate an unbounded number of useless traces unless users properly rename their traces.

magro:
Hi Ivan, I'm currently on vacation, I'll update the msg once I'm back. 

Re default trace names: It would be great if there would be useful strategies available out of the box for play that could be configured, so that starting with kamon-play doesn't force you to write code.

dpsoft:
Hi @magro,  can you please update the message in order to merge de PR.

magro:
@dpsoft done, thanks for the reminder!

dpsoft:
merged!!

magro:
Thanks!

PR #162, opened 2015-03-09
! all: introduced support for metric tags.

ivantopo:
This PR brings metric tags to Kamon! Tags are associated with entities, not with individual metrics in a entity recorder as we initially thought. To achieve this some important changes were introduced as well:
  - the API for registering metrics is simple and does not give you back an Option[EntityRegistration[Recorder]] but rather returns a plain recorder. If you want to make use of filters then use `Kamon.metrics.shouldTrack(..)`.
  - the segment metrics are no longer stored inside the trace metrics recorder but rather in their own segment metrics recorder and they will always have a tag named `trace` with the value of the trace in which they were created.
  - since Datadog already supports tags the kamon-datadog module will send them right away :D.
  - the idea of using MetricKey implementations to store additional information about the metrics has been discarded in favor of metric tags.


dpsoft:
LGTM. I think that we can merge and work in master for the next release.

Issue #172, opened 2015-03-27
Resolve the StatsD hostname upon every flush


ivantopo:
See: https://groups.google.com/forum/#!topic/kamon-user/fokzLslVN0A

dpsoft:
fixed with #174

PR #174, opened 2015-03-27
StatsD extension now defers the creation of the InetSocketAddress instance until needed

amnesiac:
StatsD extension now defers the creation of the InetSocketAddress instance against the provided UDP host/port until it's required for sending a metric payload. (https://github.com/kamon-io/Kamon/issues/172)

### DIFF
--- a/kamon-statsd/src/main/scala/kamon/statsd/StatsDMetricsSender.scala
+++ b/kamon-statsd/src/main/scala/kamon/statsd/StatsDMetricsSender.scala
@@ -26,7 +26,7 @@ import java.util.Locale
 
 import kamon.metric.instrument.{ Counter, Histogram }
 
-class StatsDMetricsSender(remote: InetSocketAddress, maxPacketSizeInBytes: Long, metricKeyGenerator: MetricKeyGenerator)
+class StatsDMetricsSender(statsDHost: String, statsDPort: Int, maxPacketSizeInBytes: Long, metricKeyGenerator: MetricKeyGenerator)
     extends Actor with UdpExtensionProvider {
   import context.system
 
@@ -38,6 +38,8 @@ class StatsDMetricsSender(remote: InetSocketAddress, maxPacketSizeInBytes: Long,
 
   udpExtension ! Udp.SimpleSender
 
+  def newSocketAddress = new InetSocketAddress(statsDHost, statsDPort)
+
   def receive = {
     case Udp.SimpleSenderReady ⇒
       context.become(ready(sender))
@@ -48,7 +50,7 @@ class StatsDMetricsSender(remote: InetSocketAddress, maxPacketSizeInBytes: Long,
   }
 
   def writeMetricsToRemote(tick: TickMetricSnapshot, udpSender: ActorRef): Unit = {
-    val packetBuilder = new MetricDataPacketBuilder(maxPacketSizeInBytes, udpSender, remote)
+    val packetBuilder = new MetricDataPacketBuilder(maxPacketSizeInBytes, udpSender, newSocketAddress)
 
     for (
       (entity, snapshot) ← tick.metrics;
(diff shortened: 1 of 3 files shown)


kamon-statsd/src/main/scala/kamon/statsd/StatsDMetricsSender.scala
```diff
@@ -48,7 +48,7 @@ class StatsDMetricsSender(remote: InetSocketAddress, maxPacketSizeInBytes: Long,
   }
 
   def writeMetricsToRemote(tick: TickMetricSnapshot, udpSender: ActorRef): Unit = {
-    val packetBuilder = new MetricDataPacketBuilder(maxPacketSizeInBytes, udpSender, remote)
+    val packetBuilder = new MetricDataPacketBuilder(maxPacketSizeInBytes, udpSender, statsDHost, statsDPort)
```
ivantopo: You could simply make remote a def in this actor using the provided host/port internally and keep the MetricDataPacketBuilder intact.

kamon-statsd/src/main/scala/kamon/statsd/StatsDMetricsSender.scala
```diff
@@ -80,15 +80,15 @@ class StatsDMetricsSender(remote: InetSocketAddress, maxPacketSizeInBytes: Long,
...
+  def props(statsDHost: String, statsDPort: Int, maxPacketSize: Long, metricKeyGenerator: MetricKeyGenerator): Props =
+    Props(new StatsDMetricsSender(statsDHost, statsDPort, maxPacketSize, metricKeyGenerator))
 }
 
 trait UdpExtensionProvider {
   def udpExtension(implicit system: ActorSystem): ActorRef = IO(Udp)
 }
 
-class MetricDataPacketBuilder(maxPacketSizeInBytes: Long, udpSender: ActorRef, remote: InetSocketAddress) {
+class MetricDataPacketBuilder(maxPacketSizeInBytes: Long, udpSender: ActorRef, statsDHost: String, statsDPort: Int) {
```
ivantopo: I think there is no need to change the MetricDataPacketBuilder, it can receive `remote: InetSocketAddress` as it did before because typically all the UDP packets are dispatched within a few milliseconds and there is no chance that the cached address will change in that time given the minimum cache time is 1 second and the host is already resolved in line 51 (after the suggested change). Only one TickMetricSnapshot is received on every flush so, by default, this guy will only be created once every 10 seconds and live for a few milliseconds.

Another option would be to still receive the host and port and have a val with them inside MetricDataPacketBuilder, both approaches seem fine to me.

ivantopo:
@amnesiac thanks for proposing this! Please review the comments that I left, update your commit message and electronically sign the CLA as described in our [contributing guide]. After that we are good to go :).

[contributing guide]: https://github.com/kamon-io/Kamon/blob/master/CONTRIBUTING.md#if-you-want-to-make-a-code-contribution-to-the-project

kamon-statsd/src/main/scala/kamon/statsd/StatsDMetricsSender.scala
amnesiac: That's right, I will do that. To be honest, I thought about some sort of InetSocketAddresProvider instance that will allow you to either create one an instance on demand or a single one depending on how you guys wanted to go in the future, but honestly I don't think there's a need to just create one instance, it's cheap to create them always on-the fly since they get discarded right away.

amnesiac:
@ivantopo done! I fixed the commit, a single one using Spray's commit message layout. I also signed the CLA. Let me know if there's anything else we need.

ivantopo:
@amnesiac thanks for the efforts! I just need to ask you for one more thing: please include the name of the module being changed in the commit message, it should be something like `= statsd: creation of....` and we are good to go :).

amnesiac:
Oh yeah, I forgot about that, done!

ivantopo:
thanks Marco, we just published the snapshot version `0.3.6-5456e7ca8f215b2a3a5bdbd3177a6ae61f268eef` including this change in case you want to give it a try, regards!

amnesiac:
@ivantopo will definitely do, and thanks a lot for the quick follow up!

PR #186, opened 2015-04-23
+ Add possibility to send Akka metrics to newrelic

slavaschmidt:

Example of the relevant configuration file part:

```
kamon.newrelic {
	app-name = "App Name"
	license-key = LicenseKey
	metrics {
	  user-metrics = "*"
	  trace = "*"
	  akka-dispatcher = "**"
	  akka-actor = "**"
	}
}
```

dashboards in following format:

AkkaActor/{ActorSystemName}/[user|system|...]/{ActorName}/{MetricName}

`AkkaActor/MyActor/user/*DatabaseWorker/ProcessingTime`

### DIFF
--- a/kamon-newrelic/src/main/resources/reference.conf
+++ b/kamon-newrelic/src/main/resources/reference.conf
@@ -23,6 +23,11 @@ kamon {
 
     # delay between connection attempts to NewRelic collector
     connect-retry-delay = 30 seconds
+
+    metrics {
+      user-metrics = "*"
+      trace  = "*"
+    }
   }
 
   modules {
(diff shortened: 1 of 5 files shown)


ivantopo:
Hello there! Thanks for proposing this, it certainly can be of help to many New Relic users! I like the idea of pushing some additional stuff as custom metrics, what I don't like a lot is that the efforts are kind of tied to pushing akka-specific metrics to NR but this could be taken in a slightly different direction allow any custom metric to be reported. 

I think we could simply generalize and have a simple metric naming structure that would work for any metric under the "custom metrics" section but still keep transaction and segments metrics the way they are today, as NR expect them in their current "places" to fill the default dashboards and so. 

What about something like "Custom/${entity.category}/${entity.name}/${metric.name}" where the entity name of course needs to be normalized to avoid introducing more `/` characters?

slavaschmidt:
Hello!

I've changed metrics naming as you suggested in #192

Thanks a lot for your feedback!

PR #188, opened 2015-04-25
register: Use the Entity.name in addition to the Entity.category

henrymai:
See:
https://groups.google.com/d/msg/kamon-user/0ZZ4IMu9t-g/4uPXHbGqg3MJ

### DIFF
--- a/kamon-annotation/src/test/scala/kamon/annotation/AnnotationInstrumentationSpec.scala
+++ b/kamon-annotation/src/test/scala/kamon/annotation/AnnotationInstrumentationSpec.scala
@@ -73,17 +73,17 @@ class AnnotationInstrumentationSpec extends BaseKamonSpec("annotation-instrument
       for (id ← 1 to 10) Annotated(id).count()
 
       val snapshot = takeSnapshotOf("count", "counter")
-      snapshot.counter("counter").get.count should be(10)
+      snapshot.counter("counter.count").get.count should be(10)
     }
 
     "count the invocations of a method annotated with @Count and evaluate EL expressions" in {
       for (id ← 1 to 2) Annotated(id).countWithEL()
 
       val counter1Snapshot = takeSnapshotOf("count:1", "counter", Map("counter" -> "1", "env" -> "prod"))
-      counter1Snapshot.counter("counter").get.count should be(1)
+      counter1Snapshot.counter("counter.count:1").get.count should be(1)
 
       val counter2Snapshot = takeSnapshotOf("count:2", "counter", Map("counter" -> "1", "env" -> "prod"))
-      counter2Snapshot.counter("counter").get.count should be(1)
+      counter2Snapshot.counter("counter.count:2").get.count should be(1)
     }
 
     "count the current invocations of a method annotated with @MinMaxCount" in {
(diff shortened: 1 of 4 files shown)


ivantopo:
Hello @henrymai, thanks for proposing this! I appreciate the efforts you already put to provide a solution to the problem you are seeing, but I think we should move the efforts in a different direction: The changes you need are specific to how the kamon-datadog module is sending data to the Datadog agent and we shouldn't touch the core internals to achieve this. 

All that is really needed to achieve what you want is to modify the [buildMetricName] function in the datadog module to include `entity.name` for single metric entities and also [buildIdentificationTag] to avoid including the name tag for these entities.

That said, I'm not sure if we should just apply the change or provide both the current and new options and allow users to select one via a config setting. Also, I remember that we started including entity names as tags because that way it was simpler to slice and dice with metrics in the Datadog dashboards, seems like it favors general metric names and get to the specifics using tags.. what are your thoughts on that matter? how is the counter name being a tag affecting the ability to create dashboards?

As you might guess, I'm not a Datadog user so any info provided by real users to understand the common usage patterns is greatly appreciated :).


[buildMetricName]: https://github.com/kamon-io/Kamon/blob/master/kamon-datadog/src/main/scala/kamon/datadog/DatadogMetricsSender.scala#L92
[buildIdentificationTag]: https://github.com/kamon-io/Kamon/blob/master/kamon-datadog/src/main/scala/kamon/datadog/DatadogMetricsSender.scala#L98

henrymai:
> Also, I remember that we started including entity names as tags because that way it was simpler to slice and dice with metrics in the Datadog dashboards, seems like it favors general metric names and get to the specifics using tags.. what are your thoughts on that matter? 

From my usage of datadog, we need something more specific for the counter name before we break down by tags.

So imagine in your application you have the following apis:
- /api/a
- /api/b
And we have three nodes that the apis are served on.
What is normally done is we would have the following counters:
- counter.api.a
- counter.api.b
And then tag the counters by node names.
With this structure, in datadog, we can graph counter.api.a, where we break down the data by the node tags. This way we could have individual lines graphed per node for '/api/a'. Meaning you would have a line for the number of times '/api/a' was hit on 'node1' and another line for the number of times '/api/a' was hit on 'node2', and etc.

Now if instead if we go with what is currently being done in kamon, you would have everything just under 'counter' and then have the following tags: 'api.a', 'api.b', 'node1', 'node2', 'node3', now when you break down the graph via the tags, it doesn't really mean anything. For example, the 'node1' line on the graph would contain the counter data from both 'api.a' and 'api.b' on a single line, which is not useful, since the aggregate statistic of both a and b combined doesn't tell us anything. You could break down the graph by the 'api.a' tag or the 'api.b' tag, but that has the same issue, where you would get the counts for the aggregate hits for the specific api, but no breakdown on the specific node counts for the api.

This is why we need to be able to actually have tags under a specific name rather than a general overarching 'counter' name.

ivantopo:
Thanks for the explanation, I totally see your point now. So, the change is definitely a positive thing to do, but we should move them to the kamon-datadog module as I mentioned before.

Please let us know if you still want to make the changes or if not we will do it anyway :), thanks for the insight!

henrymai:
Hey Ivan,

Thanks for the update, I'll close this pull request and let you guys make the right changes.

ivantopo:
Hey @henrymai! Applied the changes on f62f099989eb615d12291c546ade2f66fa7e63e2, that will be part of 0.4.0, thanks again for the input!

henrymai:
Hey @ivantopo,

I just tested out the latest snapshot and the counter tags are working as expected now.
Trace metrics though don't seem to be working the way we would expect.
Right now they show kind of like:
<application_name>.trace.elapsed_time.*
    tagged with: [<trace_name>, ...]
<application_name>.trace_segment.elapsed_time.*
   tagged with: [<segment_name>, ...]

It would be nice to get the behavior for counters from f62f099 applied to traces and segments as well.
Where it would look like instead:
<application_name>.trace.<trace_name>.elapsed_time.*
    tagged with: [...]
<application_name>.trace_segment.<segment_name>.elapsed_time.*
   tagged with: [...]

Thanks.

ivantopo:
hey @henrymai! I see what you mean and I just filled #210 to make it configurable, please provide any further feedback in that issue and I'll ping you when we have something working so that you can give it a try :), regards!

PR #192, opened 2015-05-06
+ Add possibility to send Akka metrics to newrelic

slavaschmidt:

Example of the relevant configuration file part:

```
kamon.newrelic {
	app-name = "App Name"
	license-key = LicenseKey
	metrics {
	  user-metrics = "*"
	  trace = "*"
	  akka-dispatcher = "**"
	  akka-actor = "**"
	}
}
```

dashboards in following format:

Cusom/AkkaActor/{ActorSystemName\[user|system|...]\{ActorName}/{MetricName}

for actors and

Custom/AkkaThreadPoolExecutor/Some-Service\akka.io.pinned-dispatcher/ProcessedTasks

for thread pools.

`Cusom/AkkaActor/MyActor\user\*DatabaseWorker/ProcessingTime`


ivantopo:
Hello @slavaschmidt! Finally getting back to this :D, I see that you applied some conventions as we discussed before but still the PR is very akka-specific.

What I was trying to get out of this is the possibility to report any kind of metric based on some rules, pretty much in the same way our StatsD and Datadog modules do: you specify some subscriptions and Kamon follows some naming conventions specific to sent them to StatsD/Datadog, but that has to be generic!

We put a lot of effort on making things generic in 0.4.0 precisely to allow for this kind of behavior.. so, instead of focusing on akka-specific metrics we could take advantage to include anything we want :). 

All of this, of course, only applies to custom metrics. Trace and Segment metrics still need a very specific format to be included correctly in New Relic's dashboard, but we can do something special for Custom Metrics. Would you like to chat a bit about it?

PR #200, opened 2015-05-13
+ newrelic: add possibility to send akka metrics to the newrelic

slavaschmidt:

dashboards in some format similar to:

Cusom/akka-actor/{ActorSystemName[user|system|...]\{ActorName}/{MetricName}

for actors and

Custom/akka-thread-pool-executor\Some-Service\akka.io.pinned-dispatcher/ProcessedTasks

for thread pools.

Cusom/akka-actor\MyActor\user\*DatabaseWorker/ProcessingTime

### DIFF
--- a/kamon-newrelic/src/main/resources/reference.conf
+++ b/kamon-newrelic/src/main/resources/reference.conf
@@ -24,7 +24,7 @@ kamon {
     # delay between connection attempts to NewRelic collector
     connect-retry-delay = 30 seconds
 
-    subscriptions {
+    custom-metric-subscriptions {
       counter         = [ "**" ]
       histogram       = [ "**" ]
       min-max-counter = [ "**" ]
--- a/kamon-newrelic/src/main/resources/reference.conf
+++ b/kamon-newrelic/src/main/resources/reference.conf
@@ -23,6 +23,16 @@ kamon {
 
     # delay between connection attempts to NewRelic collector
     connect-retry-delay = 30 seconds
+
+    subscriptions {
+      counter         = [ "**" ]
+      histogram       = [ "**" ]
+      min-max-counter = [ "**" ]
+      gauge           = [ "**" ]
+      akka-actor      = [ "**" ]
+      akka-dispatcher = [ "**" ]
+      akka-router     = [ "**" ]
+    }
   }
 
   modules {
--- a/kamon-newrelic/src/main/scala/kamon/newrelic/Agent.scala
+++ b/kamon-newrelic/src/main/scala/kamon/newrelic/Agent.scala
@@ -22,7 +22,7 @@ import akka.io.IO
 import akka.util.Timeout
 import com.typesafe.config.Config
 import kamon.Kamon
-import kamon.metric.{SegmentMetrics, TraceMetrics, MetricsModule, TickMetricSnapshotBuffer}
+import kamon.metric.{ SegmentMetrics, TraceMetrics, MetricsModule, TickMetricSnapshotBuffer }
 import spray.can.Http
 import spray.json._
 import scala.concurrent.Future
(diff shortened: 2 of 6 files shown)


dpsoft:
@slavaschmidt  thanks for this PR!!! and LGTM. What you think @ivantopo?.

kamon-newrelic/src/main/resources/reference.conf
```diff
@@ -23,6 +23,18 @@ kamon {
 
     # delay between connection attempts to NewRelic collector
     connect-retry-delay = 30 seconds
+
+    subscriptions {
+      trace           = [ "**" ]
+      trace-segment   = [ "**" ]
```
ivantopo: I think that trace and trace-segment subscriptions should not be configurable or at least not in the same place. Those two categories are the very foundation of how the New Relic dashboard works and should not be treated the same as we treat custom metrics.

We could always subscribe in code to all traces and segments and change this key to `custom-metric-subscriptions` to hold all the customized additions. That would also feel more natural to the users as they will know for sure that whatever they put in those keys will be reported as custom metrics. 

If you want to go the extra mile, the trace and trace-segment subscriptions could be also configurable via a separate config key, but at the moment I don't find that to be particularly useful.

kamon-newrelic/src/main/scala/kamon/newrelic/CustomMetricExtractor.scala
```diff
@@ -16,22 +16,31 @@
...
-import kamon.metric.{ EntitySnapshot, Entity }
+import kamon.metric.{ MetricKey, EntitySnapshot, Entity }
 import kamon.metric.instrument.CollectionContext
 
 object CustomMetricExtractor extends MetricExtractor {
 
   def extract(settings: AgentSettings, collectionContext: CollectionContext, metrics: Map[Entity, EntitySnapshot]): Map[MetricID, MetricData] = {
-    def onlySimpleMetrics(kv: (Entity, EntitySnapshot)): Boolean =
-      kamon.metric.SingleInstrumentEntityRecorder.AllCategories.contains(kv._1.category)
+    val (simple, complex) = metrics.partition(simpleMetrics)
```
ivantopo: Here the `metrics` map will contain all metrics, included trace and trace-segment metrics, they should be filtered out as they will already be processed by the web transaction metrics extractor.

ivantopo:
Left a couple notes there! Looking forward to your feedback.

slavaschmidt:
Hi @ivantopo 
Many thanks for quick feedback. Sorry, i missed these SegmentMetrics and TraceMetrics subscriptions.
I've removed them from the config and added an additional filter in the subscription code.

ivantopo:
There is nothing to be sorry about! Now, I see you removed trace and trace-segment from the config file, but I don't see the code that is actually subscribing to those categories, am I missing something? Also, filtering them from the custom metrics extractor is also missing.. maybe you forgot to push some of those changes?

kamon-newrelic/src/main/scala/kamon/newrelic/Agent.scala
```diff
@@ -143,4 +158,46 @@ object AgentSettings {
...
+
+}
+
+object MetricsSubscription {
+
+  private val defaultPattern = "**"
+
+  private val traceAndSegmentMetrics = Seq(TraceMetrics.category, SegmentMetrics.category)
+
+  def traceOrSegmentMetric(name: String) = traceAndSegmentMetrics contains name
```
ivantopo: `isTraceOrSegmentEntity` might be more appropriate, and pass in the `Entity` instead of just the category name. And please, add explicit return type! We know it is not necessary, but in general it makes code more readable :D.

ivantopo:
Closing this, as #228 was already merged and contains this changes.

PR #201, opened 2015-05-19
= core: fix ordered-sampler

Elyrixia:
According to [kamon.io](http://kamon.io/core/tracing/core-concepts/), the ordered-sampler config should be named *kamon.trace.ordered-sampler.sample-interval*, but TracerSettings was loading *interval* instead of *sample-interval*.

### DIFF
--- a/kamon-core/src/main/scala/kamon/trace/TracingContext.scala
+++ b/kamon-core/src/main/scala/kamon/trace/TracingContext.scala
@@ -66,7 +66,7 @@ private[trace] class TracingContext(traceName: String, token: String, izOpen: Bo
       val segment = currentSegments.next()
       if (segment.isClosed)
         segmentsInfo += segment.createSegmentInfo(_startTimestamp, startTimestamp)
-      else
+      else if (log != null)
         log.warning("Segment [{}] will be left out of TraceInfo because it was still open.", segment.name)
     }
 
--- a/kamon-core/src/main/scala/kamon/trace/Sampler.scala
+++ b/kamon-core/src/main/scala/kamon/trace/Sampler.scala
@@ -45,8 +45,8 @@ class RandomSampler(chance: Int) extends Sampler {
 class OrderedSampler(interval: Int) extends Sampler {
   import OrderedSampler._
 
-  require(interval > 0, "kamon.trace.ordered-sampler.interval cannot be <= 0")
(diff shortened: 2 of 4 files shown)


kamon-core/src/main/scala/kamon/trace/TracingContext.scala
```diff
@@ -66,7 +66,7 @@ private[trace] class TracingContext(traceName: String, token: String, izOpen: Bo
       val segment = currentSegments.next()
       if (segment.isClosed)
         segmentsInfo += segment.createSegmentInfo(_startTimestamp, startTimestamp)
-      else
+      else if (log != null)
```
ivantopo: has this ever been null in the tests that you have performed? I',m not sure of why the check is being made.

ivantopo:
Hello @Elyrixia! Thanks for fixing this :). Please remember to electronically sign our CLA so we can merge this: https://docs.google.com/forms/d/1G_IDrBTFzOMwHvhxfKRBwNtpRelSa_MZ6jecH8lpTlc/viewform

kamon-core/src/main/scala/kamon/trace/TracingContext.scala
Elyrixia: Yeah, I had some NullPointerException when working with simple-trace and some futures : **TracerModule** currently instantiate **TraceContext** with a null **LoggingAdapter** in function [*createTraceContext*](https://github.com/kamon-io/Kamon/blob/master/kamon-core/src/main/scala/kamon/trace/TracerModule.scala#L121).  
I didn't want to change the behavior of **TracerModule** so I added this check to avoid any exception with what I tested, but I think it could throw others errors, for example in the case of a [**MetricsOnlySegment**](https://github.com/kamon-io/Kamon/blob/master/kamon-core/src/main/scala/kamon/trace/MetricsOnlyContext.scala#L105). Maybe the *log* parameter should be an **Option[LoggingAdapter]** if you still agree to instantiate contexts with no logger, to avoid any problems ?

Another thing to note is that [**MetricsOnlySegment**](https://github.com/kamon-io/Kamon/blob/master/kamon-core/src/main/scala/kamon/trace/MetricsOnlyContext.scala#L104) use a check before trying to send a warning where as [**TracingContext**](https://github.com/kamon-io/Kamon/blob/master/kamon-core/src/main/scala/kamon/trace/TracingContext.scala#L69) does not.

Elyrixia:
CLA signed :)

PR #208, opened 2015-05-27
+ core: add a clock-sampler, fix ordered-sampler, add unit tests for samplers.

Elyrixia:
I wanted a steady sample of traces in my project, but the current ordered sampler depends on the number of requests, which varies a lot during the day. With a clock sampler, I can for example have an idea of the state of the application each 5 seconds.

This PR also fix the ordered sampler problem as described in https://github.com/kamon-io/Kamon/pull/201.

### DIFF
--- a/kamon-core/src/main/scala/kamon/trace/Sampler.scala
+++ b/kamon-core/src/main/scala/kamon/trace/Sampler.scala
@@ -18,8 +18,7 @@ package kamon.trace
 
 import java.net.InetAddress
 import java.util.concurrent.atomic.AtomicLong
-
-import kamon.util.{ NanoInterval, Sequencer }
+import kamon.util.{ NanoTimestamp, NanoInterval, Sequencer }
 import scala.concurrent.forkjoin.ThreadLocalRandom
 
 import scala.util.Try
@@ -50,8 +49,8 @@ class RandomSampler(chance: Int) extends Sampler {
 class OrderedSampler(interval: Int) extends Sampler {
   import OrderedSampler._
 
-  require(interval > 0, "kamon.trace.ordered-sampler.interval cannot be <= 0")
-  assume(interval isPowerOfTwo, "kamon.trace.ordered-sampler.interval must be power of two")
+  require(interval > 0, "kamon.trace.ordered-sampler.sample-interval cannot be <= 0")
+  assume(interval isPowerOfTwo, "kamon.trace.ordered-sampler.sample-interval must be power of two")
 
   private val sequencer = Sequencer()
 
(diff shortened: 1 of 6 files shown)


dpsoft:
El may. 27, 2015 6:47 AM, "Sallé Tristan" <notifications@github.com>
escribió:
>
> I wanted a steady sample of traces in my project, but the current ordered
sampler depends on the number of requests, which varies a lot during the
day. With a clock sampler, I can for example have an idea of the state of
the application each 5 seconds.

kamon-core/src/main/scala/kamon/trace/Sampler.scala
```diff
@@ -71,3 +71,16 @@ class ThresholdSampler(thresholdInNanoseconds: Long) extends Sampler {
   def shouldReport(traceElapsedTime: NanoInterval): Boolean = traceElapsedTime.nanos >= thresholdInNanoseconds
 }
 
+class ClockSampler(pauseInNanoseconds: Long) extends Sampler {
```
ivantopo: The parameter should be also a `NanoInterval` to keep it consistent across samplers. Also, the current implementation is not thread-safe because of the `timer` var bellow which needs to be safely published and coordinated. This code is going to be called from multiple threads so some changes need to be made.

kamon-core/src/main/scala/kamon/trace/TracerModule.scala
```diff
@@ -122,12 +122,12 @@ private[kamon] class TracerModuleImpl(metricsExtension: MetricsModule, config: C
...
-    else {
-      if (!_settings.sampler.shouldTrace)
+    _settings.levelOfDetail match {
+      case LevelOfDetail.MetricsOnly ⇒
         newMetricsOnlyContext(traceToken)
-      else
+      case _ if !isLocal || !_settings.sampler.shouldTrace ⇒
+        newMetricsOnlyContext(traceToken)
+      case _ ⇒
         new TracingContext(traceName, traceToken, true, _settings.levelOfDetail, isLocal, startTimestamp, null, dispatchTracingContext)
```
ivantopo: That null being passed there is surely a leftover from some refactored I did, and I apologize for it! But now that you are there, maybe you can help clean it out anyway :D.

In the previous PR you included a null check for the log instance which I don't see in this one. Maybe we should simply provide the TracerModule with a logging adapter via constructor param, WDYT?

ivantopo: Leaving the null there or adding the null check are definitely no options :)

Elyrixia: Yeah, I thought it would be better to fix logging problems in another PR, but I didn't know the best way to do it. Maybe a LoggingAdapter could be instantiated when [TracerModule start method](https://github.com/kamon-io/Kamon/blob/master/kamon-core/src/main/scala/kamon/trace/TracerModule.scala#L158) is called ? Or just provide a logging when instantiating TracerModuleImpl like you said.

ivantopo: Yeap, that start call seem to be the right place and it is a fairly simple change, can you please include it? although the thread safety issue would be much more important to correct in order to merge this PR :).

Elyrixia:
I just fixed the clock sampler but I don't know if this is the best way to do it.

dpsoft:
@Elyrixia  thanks for this PR! Please review the comments that I left and just need to ask you please squash your commits and amend the message to something like `"+ core: add a clock-sampler, fix ordered-sampler, add unit tests for samplers"` and will merge right away, just to match our commit message conventions.

Elyrixia:
Hello @dpsoft, I squash the commits as you asked but I don't see any comments left to review.

ivantopo:
@Elyrixia thanks a lot for this contribution, and sorry that it took us (or more precisely, me) too long to merge it. We really appreciate all the effort you are doing around Kamon, keep up with the awesome work you are doing! :D

jtjeferreira:
@ivantopo This got lost in the 1.x rewrite right?

Issue #134, opened 2014-12-24
akka-http module


dpsoft:
This issue should be used for obtain ideas that could be useful as metrics.

- [ ] propagate the TraceContext over requests.
- [ ] metrics about requests.

pjan:
at the same time, metrics about the server would make sense as well, I believe (total number of requests, request rate, number of open connections, ...)

rabzu:
fyi:
https://github.com/kamon-io/kamon-akka-http

dpsoft:
related to https://github.com/kamon-io/Kamon/pull/216

PR #216, opened 2015-06-10
+ akka-http-server: module for akka-http-server

pjan:
As akka-http is close to release, I put some time in making a module for Kamon. Closely follows and extends #134.

- Depends on 1.0-RC3 of `akka-http-core` and `akka-http`
- Purely focussed on the *server-side* of akka-http for now
- Currently only works with the **[high-level server-side api](http://doc.akka.io/docs/akka-stream-and-http-experimental/1.0-RC3/scala/http/routing-dsl/index.html)**
- Close to feature-parity with the Spray module for the server side (I have not copied the `TraceLocal storeDiagnosticData` as I haven't completely figured out where that data is used. Implementing this would be easy though)
- The `AkkaHttpServerMetrics` extends the `HttpServerMetrics` and adds:
  - connection counts per server binding
  - an open connection MinMaxCounter that tracks the number of outstanding/open requests
- Updated the `LogReporter` to also report both of the above metrics

I have some tests ready, but they're currently a bit rough around the edges, and would like to get some feedback on this before completely fleshing them out.

Hope this one gets some love
@dpsoft & @ivantopo

### DIFF
--- a/kamon-akka-http-server/src/main/scala/kamon/akka/http/server/AkkaHttpServerMetrics.scala
+++ b/kamon-akka-http-server/src/main/scala/kamon/akka/http/server/AkkaHttpServerMetrics.scala
@@ -1,0 +1,33 @@
+/* ===================================================
+ * Copyright © 2013 the kamon project <http://kamon.io/>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ========================================================== */
+package kamon.akka.http.server
+
+import kamon.http.HttpServerMetrics
+import kamon.metric.EntityRecorderFactory
+import kamon.metric.instrument.InstrumentFactory
+
+class AkkaHttpServerMetrics(instrumentFactory: InstrumentFactory) extends HttpServerMetrics(instrumentFactory) {
+  def recordConnection(bindingAddress: String): Unit =
+    counter(bindingAddress).increment()
+
+  def openConnections = minMaxCounter("open-connections")
+}
+
+object AkkaHttpServerMetrics extends EntityRecorderFactory[AkkaHttpServerMetrics] {
+  def category: String = "akka-http-server"
+  def createRecorder(instrumentFactory: InstrumentFactory): AkkaHttpServerMetrics =
+    new AkkaHttpServerMetrics(instrumentFactory)
+}
--- a/kamon-akka-http-server/src/main/scala/kamon/akka/http/server/instrumentation/ServerInstrumentation.scala
+++ b/kamon-akka-http-server/src/main/scala/kamon/akka/http/server/instrumentation/ServerInstrumentation.scala
@@ -150,7 +150,7 @@ trait ServerInstrumentationUtils {
   }
 
   def attachTraceTokenHeader(response: HttpResponse, traceTokenHeaderName: String, traceToken: String): HttpResponse =
-    response.withHeaders(response.headers :+ RawHeader(traceTokenHeaderName, traceToken))
+    response.withHeaders(response.headers.filterNot(_.name == traceTokenHeaderName) :+ RawHeader(traceTokenHeaderName, traceToken))
 }
 
 object ServerInstrumentationUtils extends ServerInstrumentationUtils
--- a/kamon-akka-http-server/src/main/scala/kamon/akka/http/server/instrumentation/ServerInstrumentation.scala
+++ b/kamon-akka-http-server/src/main/scala/kamon/akka/http/server/instrumentation/ServerInstrumentation.scala
@@ -38,13 +38,15 @@ class HttpRequestInstrumentation extends ServerInstrumentationUtils {
   @After("httpRequestCreation(request)")
   def afterHttpRequestCreation(request: HttpRequest with TraceContextAware): Unit = {
     val akkaHttpServerExtension = Kamon(AkkaHttpServer)
-    val serverMetrics = akkaHttpServerExtension.serverMetrics
+    val incomingContext = Tracer.currentContext
 
-    val defaultTraceName = akkaHttpServerExtension.generateTraceName(request)
-    val token = request.headers.find(_.name == akkaHttpServerExtension.settings.traceTokenHeaderName).map(_.value)
+    if (incomingContext.isEmpty) {
+      val defaultTraceName = akkaHttpServerExtension.generateTraceName(request)
+      val token = request.headers.find(_.name == akkaHttpServerExtension.settings.traceTokenHeaderName).map(_.value)
 
-    val newContext = Kamon.tracer.newContext(defaultTraceName, token)
-    Tracer.setCurrentContext(newContext)
+      val newContext = Kamon.tracer.newContext(defaultTraceName, token)
+      Tracer.setCurrentContext(newContext)
+    }
 
     request.traceContext
 
--- a/kamon-akka-http-server/src/main/scala/kamon/akka/http/server/instrumentation/ServerInstrumentation.scala
+++ b/kamon-akka-http-server/src/main/scala/kamon/akka/http/server/instrumentation/ServerInstrumentation.scala
@@ -1,0 +1,154 @@
+/* ===================================================
+ * Copyright © 2013 the kamon project <http://kamon.io/>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ========================================================== */
+package kamon.akka.http.server.instrumentation
+
+import akka.http.scaladsl.Http.{ IncomingConnection, ServerBinding }
+import akka.http.scaladsl.model.headers.RawHeader
+import akka.http.scaladsl.model.{ HttpResponse, HttpRequest }
+import akka.http.scaladsl.server.RequestContext
+import kamon.Kamon
+import kamon.akka.http.server.{ AkkaHttpServerExtension, AkkaHttpServer, AkkaHttpServerMetrics }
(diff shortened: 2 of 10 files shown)


dpsoft:
@pjan it is really awesome!!!, please give us a few days to review and come back with some comments.

kamon-akka-http-server/src/main/scala/kamon/akka/http/server/AkkaHttpServerMetrics.scala
```diff
@@ -0,0 +1,33 @@
...
+
+import kamon.http.HttpServerMetrics
+import kamon.metric.EntityRecorderFactory
+import kamon.metric.instrument.InstrumentFactory
+
+class AkkaHttpServerMetrics(instrumentFactory: InstrumentFactory) extends HttpServerMetrics(instrumentFactory) {
+  def recordConnection(bindingAddress: String): Unit =
+    counter(bindingAddress).increment()
+
+  def openConnections = minMaxCounter("open-connections")
```
pjan: This should probably get a more apt name, since it's not really connections it is tracking, but outstanding requests.

pjan:
Any chance anyone can take a look at this? I have another week or so with some extra bandwidth to spend some time on this, but am unsure about availabilities after that. Would be good if we could have most of it reviewed by then.

ivantopo:
Oh, I'm doing it right now :)

kamon-akka-http-server/src/main/scala/kamon/akka/http/server/instrumentation/ServerInstrumentation.scala
```diff
@@ -0,0 +1,158 @@
...
+  def afterHttpRequestCreation(request: HttpRequest with TraceContextAware): Unit = {
+    val akkaHttpServerExtension = Kamon(AkkaHttpServer)
+    val incomingContext = Tracer.currentContext
+
+    if (incomingContext.isEmpty) {
+      val defaultTraceName = akkaHttpServerExtension.generateTraceName(request)
+      val token = request.headers.find(_.name == akkaHttpServerExtension.settings.traceTokenHeaderName).map(_.value)
+
+      val newContext = Kamon.tracer.newContext(defaultTraceName, token)
+      Tracer.setCurrentContext(newContext)
```
ivantopo: I'm not sure if this would be a viable approach for creating a TraceContext, given that a HttpRequest might be created both on the server side and in the client side as well. Typically, when a HttpRequest arrives on the server side we should start a new trace but when a new HttpRequest is created on the client side we should rather create a segment, and only if there is a TraceContext available.

Also, we should always try to match a `Tracer.setCurrentContext(...)` with a `Tracer.clearCurrentContext(...)` or not use these variants at all. It is best to use the `Tracer.withContext(...)` form because it will always put whatever context was before the code blocks executes. It might seem unnecessary at some points, but we have experienced weird behaviors where the TraceContext for different requests get mixed due to leaving dirty threads, so, always clean the context! :D

kamon-akka-http-server/src/main/scala/kamon/akka/http/server/instrumentation/ServerInstrumentation.scala
```diff
@@ -0,0 +1,158 @@
...
+  def requestContextCreation(context: RequestContext with TraceContextAware): Unit = {}
+
+  @Around("requestContextCreation(context)")
+  def afterCreation(pjp: ProceedingJoinPoint, context: RequestContext with TraceContextAware): Unit = {
+    val incomingContext = Tracer.currentContext
+
+    Tracer.withContext(incomingContext) {
+      pjp.proceed()
+    }
+  }
```
ivantopo: This little part here is not necessary at all. If you run it, it is basically taking the current context, replacing it with itself, running the `pjp.proceed()` and then set the current context to what it was before, which is the same context, again.

kamon-akka-http-server/src/main/scala/kamon/akka/http/server/instrumentation/ServerInstrumentation.scala
```diff
@@ -0,0 +1,158 @@
...
+class HttpResponseInstrumentation extends ServerInstrumentationUtils {
+
+  @DeclareMixin("akka.http.scaladsl.model.HttpResponse")
+  def mixinTraceContextAwareToResponse: TraceContextAware = TraceContextAware.default
+
+  @Pointcut("call(akka.http.scaladsl.model.HttpResponse.new(..)) && !cflow(execution(* *..attachTraceTokenHeader(..)))")
+  def httpResponseCreation(): Unit = {}
+
+  @Around("httpResponseCreation()")
+  def afterHttpResponseCreation(pjp: ProceedingJoinPoint) = {
```
ivantopo: Here, we have a similar situation to the HttpRequest counterpart.. probably this is not the best place to do this because it might get mixed with code called when the http client side receives a response for a given request.

kamon-akka-http-server/src/main/scala/kamon/akka/http/server/AkkaHttpServerMetrics.scala
ivantopo: Yeap, it probably can be just `openRequests`. I like the idea of having this metric! Also, this can be a val, to avoid looking it up in the entity map every time it is needed.

One more thing is that we (Diego and I) thought to do some changes to the HttpServerMetrics, not just for Spray but for Play and akka-http as well. Our main concern at the moment is that the current implementation is just based on conventions about how to name certain metrics and you end up with counters named like [this](https://github.com/kamon-io/Kamon/blob/master/kamon-spray/src/test/scala/kamon/spray/SprayServerMetricsSpec.scala#L68), which seem rather unfortunate. After we introduced support for tags, it feels like the correct way to represent HTTP server metrics would be a group of smaller entities with well defined properties and identifying tags rather than the current single, bloated, convention based entity. Also, by doing this we could add a tag for the binding port, meaning that we could get separate metrics for different bindings made by the same application, which we don't do with the current Spray and Play modules.

project/Projects.scala
```diff
@@ -79,6 +79,16 @@ object Projects extends Build {
         provided(aspectJ) ++
         test(scalatest, akkaTestKit, akkaSlf4j, slf4Jul, slf4Log4j, logback))
 
+  lazy val kamonAkkaHttp = Project("kamon-akka-http-server", file("kamon-akka-http-server"))
```
ivantopo: The project can be just `kamon-akka-http`, as both server and client instrumentation and goodies will end up in the very same module.

kamon-akka-http-server/src/main/scala/kamon/akka/http/server/instrumentation/ServerInstrumentation.scala
pjan: I'm aware of this issue, as I started creating some tests (for which I was using akka-http-client), which brought to light exactly this behaviour. There however seems not to be a clearcut `Pointcut` in akka-http for HttpRequest creation that is specific to either the client or the server. The creation of the `RequestContext` might be an alternative, but I would need to do some more testing on that. Also, this comes later in the chain of actions, so you loose precision. The `RequestContext` is also only used in the highlevel api, so everything that is connection-level api wouldn't be tracked.

Thanks for explaining about the behaviour of `Tracer.withContext(...)` and it's alternatives.

kamon-akka-http-server/src/main/scala/kamon/akka/http/server/instrumentation/ServerInstrumentation.scala
pjan: :+1:

pjan:
@ivantopo thanks for the initial review. Some questions

1) Is there a way to easily inspect potential pointcuts (a debugger tool of sorts) so I can further inspect/analyse what would be better alternatives to `HttpRequest` and `HttpResponse` creation. I spent a couple of hours (and some more) on this already, but given the very general approach of streams, it is much more of a challenge to track down a specific pointcut as was possible for Spray. 
2) What's kamon's behaviour when a context is never closed? Does it keep hold of resources, and might eventually blow up the program because of memory issues?

gamsd:
@dpsoft @ivantopo @pjan what's the status on this? Is there any way I can help you move this forward? I'm monitoring some Akka HTTP applications right now and having a module for it would be awesome.

ivantopo:
hello guys! To address @pjan's questions first:

1) not that we are aware of :(.. all instrumentation we have is a product of reading the libraries' code, analyzing it and debugging it until we clearly understand what it does and how we can plug Kamon into it. Certainly taking decisions in Spray's codebase was a bit simpler than in this case :/.

2) Kamon does not keep any reference to the created contexts, so, if the instrumentation and/or the user code looses track of a given context then it will be gone and garbage collected.


Now, @gamsd: thanks for showing interest on moving this PR forward but, sadly, the approach taken here is not fully viable at the moment... a lot more research into akka-http's internals need to happen in order to really understand how to correctly integrate Kamon with it. So, I would recommend taking a look at this PRs and the conversations on the initial review and then start doing some research and testing with akka-http.. and, of course, share your findings and whatever results you produce :D.

I would like to dig into this so that I could at least provide some guidance with regards to how to approach the integration, but at the moment there are some other outstanding issues that need attention and my bandwidth is limited. In case that no other contributor moves this forward before I get enough time then I will use this PR as a starting point, stay tuned just in case :).

gamsd:
@ivantopo Thanks! I'll try to take a deeper look at this and see what I can do.

jypma:
I think a viable approach might be to put the server aspect around the [IncomingConnection](https://github.com/akka/akka/blob/master/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala#L682) case class' constructor, and wrap its `flow` field with a stage to keep track of incoming and outgoing requests across that connection. That flow completing would then indicate the connection itself being done. 

The above is then completely separeted from any client-side implementation. What do you think?

ncanibano:
Hi Guys! How you doing? Do you have any news about akka-http integration? We are also interested in this kind of support. Let me know if we can contribute on anything you want. Thanks so much. Keep the good work!

dpsoft:
Hi @ncanibano we think that this integration module could be really useful, but currently there are some other outstanding issues and [Kamino] that we need our attention and our bandwidth is really limited. In case that no other contributor ;) moves this forward before... We must say that in the short term we can't take this integration :(.

[Kamino]: https://twitter.com/kamonteam/status/715919867061395456

jypma:
I don't have time ATM to create a proper pull request, but I can share how we are currently monitoring our akka http server, by wrapping its `Flow` in Kamon metrics. The code is in Java, but I think could easily be turned into Scala and put into an Aspect at the right spot. 

http://pastebin.com/DHVb54iK

timcharper:
Hey! So this would be a pretty awesome change! I'd like to be able to monitor our akka http server with Kamon, too.

It'd be cool to get per-path stats, also; although, to prevent too high of cardinality perhaps list a set of matches in the configuration, and then send the matching part as part of the metric, rather than the full path.

DarrenBishop:
@jypma looking at your suggestion, seems you are stacking request start times, so what happens with interleaved requests? My understanding is limited, but seems like the metrics won't be worth much?

dpsoft:
@pjan @ncanibano @gamsd @jypma @timcharper we are just relased the first version or `kamon-akka-http`.

dpsoft:
close with https://github.com/kamon-io/kamon-akka-http

Issue #165, opened 2015-03-15
Exceptions logged on shutdown


magro:
After upgrading to the latest kamon snapshot 0.3.6-125bee567e0b18ea5ec10f3a1fe76409673011d9, on shutdown there are some errors logged like this:

```
[ERROR] [03/16/2015 00:04:05.846] [kamon-akka.actor.default-dispatcher-12] [ActorSystem(kamon)] exception while executing timer task
java.lang.RuntimeException: Kamon has not been started yet.
        at scala.sys.package$.error(package.scala:27)
        at kamon.Kamon$.ifStarted(Kamon.scala:83)
        at kamon.Kamon$.metrics(Kamon.scala:66)
        at akka.kamon.instrumentation.DispatcherInstrumentation.afterLazyExecutorShutdown(DispatcherInstrumentation.scala:135)
        at akka.dispatch.Dispatcher$LazyExecutorServiceDelegate.shutdown(Dispatcher.scala:43)
        at akka.dispatch.Dispatcher.shutdown(Dispatcher.scala:106)
        at akka.dispatch.MessageDispatcher$$anon$2.run(AbstractDispatcher.scala:212)
        at akka.actor.LightArrayRevolverScheduler$TaskHolder.run(Scheduler.scala:476)
        at akka.actor.LightArrayRevolverScheduler$$anonfun$close$1.apply(Scheduler.scala:282)
        at akka.actor.LightArrayRevolverScheduler$$anonfun$close$1.apply(Scheduler.scala:281)
        at scala.collection.Iterator$class.foreach(Iterator.scala:750)
        at scala.collection.AbstractIterator.foreach(Iterator.scala:1202)
        at scala.collection.IterableLike$class.foreach(IterableLike.scala:72)
        at scala.collection.AbstractIterable.foreach(Iterable.scala:54)
        at akka.actor.LightArrayRevolverScheduler.close(Scheduler.scala:280)
        at akka.actor.ActorSystemImpl.stopScheduler(ActorSystem.scala:688)
        at akka.actor.ActorSystemImpl$$anonfun$liftedTree2$1$1.apply$mcV$sp(ActorSystem.scala:617)
        at akka.actor.ActorSystemImpl$$anonfun$liftedTree2$1$1.apply(ActorSystem.scala:617)
        at akka.actor.ActorSystemImpl$$anonfun$liftedTree2$1$1.apply(ActorSystem.scala:617)
[...]

PR #220, opened 2015-06-17
+ akka: avoid runtime exceptions logged on ActorCell shutdown #165

enix:
Fixing the second stage bug mentioned in #165.


dpsoft:
Thanks for this contribution, @enix !!!

Issue #227, opened 2015-07-08
Memory leak when removing entities with MinMaxCounter


nilsga:
Any metric that is using a `MinMaxCounter` will not be properly garbage collected when removed from the tracked entities map with `removeEntity`. This is due to the `MinMaxCounter` starting a [refresh schedule](https://github.com/kamon-io/Kamon/blob/v0.4.0/kamon-core/src/main/scala/kamon/metric/instrument/MinMaxCounter.scala#L42-L44) that apparently is never cancelled, since [`removeEntity`](https://github.com/kamon-io/Kamon/blob/v0.4.0/kamon-core/src/main/scala/kamon/metric/MetricsModule.scala#L316-L317) does not invoke the `cleanup` method on the instrument.

PR #229, opened 2015-07-08
! core: cleanup entity on removal

nilsga:
Fixes https://github.com/kamon-io/Kamon/issues/227

### DIFF
--- a/kamon-core/src/main/scala/kamon/metric/MetricsModule.scala
+++ b/kamon-core/src/main/scala/kamon/metric/MetricsModule.scala
@@ -313,8 +313,11 @@ private[kamon] class MetricsModuleImpl(config: Config) extends MetricsModule {
     }, _.cleanup).asInstanceOf[T]
   }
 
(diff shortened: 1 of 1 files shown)


nilsga:
The tests fails locally without any modifications to master, so I can't say wether this fix introduces new test errors or not.

dpsoft:
@nilsga  thanks for this PR! Just need to ask you please squash your commits and amend the message to something like `"! core: cleanup entity on removal"` and will merge right away, just to match our commit message conventions and please sign our [CLA] and we are good to go!

[CLA]: https://docs.google.com/forms/d/1G_IDrBTFzOMwHvhxfKRBwNtpRelSa_MZ6jecH8lpTlc/viewform

nilsga:
Will do. Note that this fix only fixes the leak within `removeEntity`. There might be other `remove*`methods that needs the same fix.

ivantopo:
Thanks a lot @nilsga! We will carefully check for other places where we might need to cleanup as well, regards!

PR #236, opened 2015-07-24
+ core: Allow creation of counters with units

sattvik:
Currently, the GenericEntityRecorder only allows two ways to create a
new counter:

1. Using a simple name, resulting in a counter with an Unknown unit of
   measure
2. Using a CounterKey, but that is not really helpful since that class
   is private to Kamon

This trivial patch adds a new counter method that accepts a unit of
measure.


ivantopo:
hey @sattvik, thanks for proposing this PR and for taking all considerations (commit message, CLA) in advance :), this will be out soon!

PR #244, opened 2015-08-20
+ Add memory pool metrics

tferi:
Record the amount of used, max, and committed memory in bytes for each memory pool.

In Java 8, the following metrics will be reported:
code-cache, metaspace, compressed-class-space, ps-eden-space, ps-survivor-space, ps-old-gen
Each with their own `-used`, `-committed` and `-max` values.

This is mainly meant to provide a replacement for the metrics in [NewRelic's JVM dashboards](https://docs.newrelic.com/docs/apm/applications-menu/monitoring/jvm-metrics-dashboard).

PR #245, opened 2015-08-21
= Generic mem metrics

tferi:
Hey guys,

This is based on the pending PR https://github.com/kamon-io/Kamon/pull/244 so you need to see https://github.com/tferi/Kamon/compare/mempool-metrics...tferi:generic-mem-metrics for the relative changeset.

As the Memory metrics work the same way, I took the liberty to merge them into one class + companion object. I'll squash the changes if you approve this.

I'm late to the party so I'm missing a lot of context here, and something occurred to me. I noticed that most (all?) metrics are collected as side effects in class constructors while they could be done in regular functions. I'm curious why the design took this direction - doesn't this approach generate more workload for the GC?


ivantopo:
Hello @tferi! I like the approach, it simplifies things here so go ahead and polish it so we can merge it :). 

Also, I don't get exactly what you mean by "I noticed that most (all?) metrics are collected as side effects in class constructors while they could be done in regular functions.", in all the entity recorders' constructors we are not actually recording anything, but rather creating a little function that is later called periodically to get the metric values and record them in their proper instruments. It seemed to be the easiest and shorter way to define those entity recorders but if you see something better please let us know! We are always open to improvements :D

dpsoft:
@tferi  thanks for this PR! Just need to ask you please squash your commits and amend the message to something like `"= system-metrics: Generic mem metrics "` and will merge right away  and we are good to go!

tferi:
@ivantopo I guess I misunderstood the code then, sorry about that! I've squashed those commits, is the PR okay now?

ivantopo:
Thanks for this contribution :), hope to see more coming!

PR #246, opened 2015-08-23
! core: generalize ThreadPoolExecutors metrics

dpsoft:


kamon-core/src/main/scala/kamon/util/executors/ExecutorMetricRecorder.scala
```diff
@@ -0,0 +1,104 @@
...
+  implicit object JavaForkJoin extends ForkJoinMetrics[JavaForkJoinPool] {
+    def getParallelism(fjp: JavaForkJoinPool) = fjp.getParallelism
+    def getPoolSize(fjp: JavaForkJoinPool) = fjp.getPoolSize.toLong
+    def getRunningThreadCount(fjp: JavaForkJoinPool) = fjp.getActiveThreadCount.toLong
+    def getActiveThreadCount(fjp: JavaForkJoinPool) = fjp.getRunningThreadCount.toLong
+    def getQueuedTaskCount(fjp: JavaForkJoinPool) = fjp.getQueuedTaskCount
+
+  }
+
+  implicit object AkkaForkJoin extends ForkJoinMetrics[AkkaForkJoinPool] {
```
ivantopo: Is this really necessary? The AkkaForkJoinPool is a ScalaForkJoinPool, shouldn't it be enough with just that?

kamon-core/src/main/scala/kamon/util/executors/ExecutorServiceMetrics.scala
```diff
@@ -0,0 +1,140 @@
...
+
+import kamon.Kamon
+import kamon.metric.Entity
+import java.util.concurrent.{ ForkJoinPool ⇒ JavaForkJoinPool }
+import java.util.concurrent.{ ExecutorService, ThreadPoolExecutor }
+
+import scala.concurrent.forkjoin.ForkJoinPool
+import scala.util.control.NoStackTrace
+
+object ExecutorServiceMetrics {
```
ivantopo: This is cool, but I'm not entirely sure about the fact that all the `register` methods are returning `Unit`. What about returning the entity recorder for the thread pool? that will make it consistent with the API exposed in Kamon.metrics.

kamon-core/src/main/scala/kamon/util/executors/ExecutorMetricRecorder.scala
dpsoft: you are right, indeed i've replaced in DispatcherInstrumentation the type of the Dispatcher  to ForkJoinPool instead of AkkaForkJoinPool

kamon-core/src/main/scala/kamon/util/executors/ExecutorServiceMetrics.scala
dpsoft: I don't see the use case where the user can do something with the recorder's. I think that for now we should return `Unit`.

dpsoft:
Merged!!!

PR #256, opened 2015-09-14
newrelic: de-tokenise the app name, fixes #255

### LINKED ISSUE #255: Support the use of multiple names for a New Relic app
New Relic supports up to 3 names for an app, as documented here:

https://docs.newrelic.com/docs/apm/new-relic-apm/installation-configuration/use-multiple-names-app

It would be nice if kamon-newrelic also supported this, so that aggregation is possible.

cheekychops:
To see the benefit of this change you need at least two applications running and reporting to new relic. Configure their app_name properties like this:

- "app1;parent"
- "app2;parent"

This will create the "app1", "app2" and "parent" applications in new relic, where "parent" is the aggregate of "app1" and "app2".


dpsoft:
@cheekychops LGTM. @ivantopo what do you think?

dpsoft:
@cheekychops thanks for this PR, we will make a snapshot release with this.

PR #260, opened 2015-09-26
Fix heap metrics update

alno:
For now `MemoryUsageMetrics` stores a direct reference to `MemoryUsage` object, got from MemoryMXBean on initialization. However, `MemoryUsage` is immutable object representing some snapshot and not being updated by `MemoryMXBean` (see http://docs.oracle.com/javase/7/docs/api/java/lang/management/MemoryUsage.html) - so heap usage metrics always report the same value.

Instead it's required to get new `MemoryUsage` instance each time we want to record heap usage.

In my patch I replaced reference to `MemoryUsage` with reference to function extracting it.


dpsoft:
@alno  thanks for this PR! I need to ask you please squash your commits and amend the message to something like `"! system-metrics: Fix heap metrics update"` and will merge right away, just to match our commit message conventions and please sign our [CLA] and we are good to go!

[CLA]: https://docs.google.com/forms/d/1G_IDrBTFzOMwHvhxfKRBwNtpRelSa_MZ6jecH8lpTlc/viewform

alno:
Done

dpsoft:
Merged!!!

PR #262, opened 2015-09-30
= core: fix typo in kamon auto-start error message

jozic:


jozic:
looks like some timing issue in the tests

```
[info] ActorMetricsSpec:
[info] the Kamon actor metrics
[info] - should respect the configured include and exclude filters
[info] - should not pick up the root supervisor
[info] - should reset all recording instruments after taking a snapshot *** FAILED ***
[info]   101 was not equal to 102 (ActorMetricsSpec.scala:74)
```

dpsoft:
@jozic thanks for this fix!!!

PR #263, opened 2015-09-30
= core: don't throw an NPE during shutdown if Kamon hasn't been started

jozic:
if you don't start Kamon, but try to call Shutdown, it throws an NPE, while it can just do nothing


dpsoft:
@jozic  thanks for this PR, we will make a snapshot release with this. BTW: currently we are rewriting all to related to the lifecycle of kamon-core

PR #269, opened 2015-10-14
= newrelic: Associate logged errors with correct transaction

cheekychops:
Previously, errors were associated with transactions named
"OtherTransaction/trace-name", with this change they are associated with
"WebTransaction/Uri/trace-name" which is the name expected by NewRelic.

The additional complexity of wrapping messages in LoggedError instances
ensures that the logged message is not repeated in the New Relic UI.

### DIFF
--- a/kamon-newrelic/src/main/scala/kamon/newrelic/NewRelicErrorLogger.scala
+++ b/kamon-newrelic/src/main/scala/kamon/newrelic/NewRelicErrorLogger.scala
@@ -17,11 +17,12 @@
 package kamon.newrelic
 
 import java.util
-
 import akka.actor.{ Actor, ActorLogging }
 import akka.event.Logging.{ Error, InitializeLogger, LoggerInitialized }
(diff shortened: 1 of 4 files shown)


ivantopo:
hey @cheekychops, I like what you did there! I'm not particularly happy with the dependency on the NR agent to get that `ThrowableError` and I don't even want to ask how you realized that using it would solve the issue, but it seems to be fine!

The only thing that I would like to understand before merging is if changing the name pattern for the web transactions from `WebTransaction/Custom/name` to `WebTransaction/Uri/name` will somehow change the behavior currently experienced by users. Can you please confirm if the dashboards currently working are still working correctly with this patch?

cheekychops:
hi @ivantopo I can confirm that the dashboards are working fine with this patch, and indeed this will fix the broken navigation to the transaction from the Error Details page.
The only other difference I observed in the UI is when you hover the mouse over a particular transaction in the list, the tooltip window shows "Uri" under the transaction name instead of "Custom"

Issue #111, opened 2014-11-10
allow kamon-statsd and kamon-datadog users to define a scale value when reporting data.


ivantopo:
Currently all measurements are always reported in nanoseconds because that's what we store in our histograms but in many cases it will be more useful to the users to report the data in milliseconds or microseconds to avoid having to tweak the dashboards to display nanoseconds as millis/micros and the data usually is in those ranges anyway.

PR #270, opened 2015-10-14
! statsd: allow custom statsd senders + add simple statsd sender which doesn't batch stats

jozic:
this is not backward compatible so should go in 0.6.x
although it can be made backward compatible, not sure if this is desired here


kamon-statsd/src/test/scala/kamon/statsd/BaseStatsDMetricSenderSpec.scala
```diff
@@ -0,0 +1,65 @@
+package kamon.statsd
```
dpsoft: @jozic please include the copyright header.

kamon-statsd/src/test/scala/kamon/statsd/SimpleStatsDMetricsSenderSpec.scala
```diff
@@ -0,0 +1,62 @@
+package kamon.statsd
```
dpsoft: @jozic  please include the copyright header.

dpsoft:
@jozic i've left some comments, but looks good to me, @ivantopo WDYT?

kamon-statsd/src/test/scala/kamon/statsd/BaseStatsDMetricSenderSpec.scala
jozic: done

kamon-statsd/src/test/scala/kamon/statsd/SimpleStatsDMetricsSenderSpec.scala
jozic: done

jozic:
failure from travis

```
[error] Failed tests:
[error] 	kamon.spm.SPMMetricsSenderSpec

....
[error] Failed tests:
[error] 	kamon.akka.DispatcherMetricsSpec
[error] (kamon-spm/test:test) sbt.TestsFailedException: Tests unsuccessful
[error] (kamon-akka/test:test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 592 s, completed Oct 23, 2015 6:13:23 PM
```
looks like those are non-stable tests. 
could someone with write permissions rerun the build on travis?

kamon-statsd/src/main/resources/reference.conf
```diff
@@ -5,18 +5,10 @@
 kamon {
   statsd {
 
-    # Hostname and port in which your StatsD is running. Remember that StatsD packets are sent using UDP and
-    # setting unreachable hosts and/or not open ports wont be warned by the Kamon, your data wont go anywhere.
-    hostname = "127.0.0.1"
-    port = 8125
```
ivantopo: I think we should keep the host/port configuration in a single place. After all, regardless of the concrete sender and key generator there will always be a single host/port to send info to. Also, it will be simpler for users!

ivantopo:
hey @jozic! I really like what you did here, please check the comment about the config and I think after that we are pretty much good to go on my side. I find it exciting that in theory with this changes, adding a TCP implementation would be quite easy. Thanks!

kamon-statsd/src/main/resources/reference.conf
jozic: hey @ivantopo, although here both of the senders use single host/port pair, custom senders (like we do have) can use some kind of sharding technic to send to statsd to different hosts. So i believe it make sense to have host/port in sender specific settings.
for example my custom sender has this key instead of host and port.

```
    cluster-nodes = [
      "localhost:8125"
    ]
```
we can keep those for backward compatibility, but again i thinks this is up to sender
wdyt?

ivantopo: I see what you mean, but custom senders will be a very special situation. Out of the box we should support UDP/TCP and Batched/Unbatched senders to a single StatsD instance and for most users that will be enough. So, to keep it simple for users I would still prefer to use the old `kamon.statsd.[hostname|port]` keys that will be read by the senders provided out-of-the-box, and then custom senders (like yours) have the choice of using that or reading that info from somewhere else, as you already do.

@dpsoft what are your thoughts on this too?

dpsoft: @ivantopo I agreed with you, I think  that is a good idea keep it simple for most users and i prefer use the old `kamon.statsd.[hostname|port]`

jozic: do you guys want them like
```
statsd {
  hostname = "localhost"
  port = 8125
  batch-metric-sender {
    hostname = ${kamon.statsd.hostname}
    port = ${kamon.statsd.port}
    max-packet-size = 1024 bytes
  }
}
```
or just keep them on top level only? what about  `max-packet-size`

ivantopo: top level only, please :). `max-packet-size` is specific to the batch metric sender, so it can remain where it is now.. thanks!

jozic:
@ivantopo @dpsoft 
- rebased on latest master
- __hostname/port config settings moved to top level (removed simple-statics-sender section as a result)__
- udpSender is kept in base class, passing flush function to subclasses
- socketAddress is converted to lazy val instead of def

please review

jozic:
sent another PR to update docs 
https://github.com/kamon-io/kamon.io/pull/11

jozic:
hey @ivantopo @dpsoft 
how do you feel about it?
I'd like to give #111 a shot, but would like to finish this thing first

ivantopo:
hey @jozic, thanks again for being such a great contributor! :D hope to keep seeing cool stuff from you!

jozic:
thanks @ivantopo!

PR #275, opened 2015-10-28
= build: upgrade to sbt 0.13.9

cheekychops:
The previous version of sbt has a bug which prevents publishM2 from overwriting artifacts. This fixes that.


cheekychops:
dang wrong branch, sorry!

PR #288, opened 2015-11-21
+ statsd: scale statsd time and memory values to values from config

jozic:
this is an attempt to fix part of #111 

@ivantopo @dpsoft please review
please note changes to Memory UnitOfMeasurement

if this looks okay, i can send another PR for datadog module

### DIFF
--- a/kamon-core/src/main/scala/kamon/metric/EntityRecorder.scala
+++ b/kamon-core/src/main/scala/kamon/metric/EntityRecorder.scala
@@ -89,7 +89,7 @@ case class GaugeRecorder(key: MetricKey, instrument: Gauge) extends SingleInstru
 
 /**
  *  Base class with plenty of utility methods to facilitate the creation of [[EntityRecorder]] implementations.
- *  It is not required to use this base class for defining custom a custom [[EntityRecorder]], but it is certainly
+ *  It is not required to use this base class for defining a custom [[EntityRecorder]], but it is certainly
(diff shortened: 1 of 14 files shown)


dpsoft:
@jozic LGTM. @ivantopo WDYT?

ivantopo:
hey @jozic, this looks good to me! The only thing that I kind of not like is the name of the settings and feature.. the word "rescale" sounds like if things were already scaled and need to be scaled again. I would go for a simpler config name like "memory-units" and "time-units" that default to Nanoseconds and Bytes, respectively. WDYT?

Other than that, I'm happy with the approach, thanks! :D

jozic:
sure, no problem, will rename things

jozic:
hey @dpsoft @ivantopo 
i did renaming according to your suggestion...

But i also tried another approach in a separate branch. 
The idea is to have an intermediate actor which scales TickMetricSnapshot (via InstrumentationSnapshots) before forwarding to a sender.
It adds one more actor in chain, but decouples scaling from any sender, and can be easily used for any backend. Please take a look here https://github.com/jozic/Kamon/commit/761ae0e7db07244ddade3402bcf8e0dbc85a8d37 and let me know what you think. i will add more tests and docs if you think we should move this way

jozic:
@ivantopo @dpsoft ping, 
i know you guys are busy, but would like to finish this before next release

jozic:
closed in favor of #294

PR #292, opened 2015-11-29
Kamon autoweave

dpsoft:
This new module allow attach the AspectJ loadtime weaving agent to a JVM after it has started (you don't need to use -javaagent). This offers extra flexibility but obviously any classes loaded before attachment will not be woven.


dpsoft:
closed in favor of #296

PR #294, opened 2015-12-01
Scale

### LINKED ISSUE #111: allow kamon-statsd and kamon-datadog users to define a scale value when reporting data.

jozic:
supersedes #288 and fixes #111  
@ivantopo @dpsoft please review
if this looks okay, i will send another pr to kamon-io/kamon.io to update docs for statsd and datadog

### DIFF
--- a/kamon-core/src/main/scala/kamon/metric/MetricScaleDecorator.scala
+++ b/kamon-core/src/main/scala/kamon/metric/MetricScaleDecorator.scala
@@ -1,0 +1,55 @@
+/*
+ * =========================================================================================
+ * Copyright © 2013-2015 the kamon project <http://kamon.io/>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ * =========================================================================================
+ */
+
+package kamon.metric
+
+import akka.actor.{ Actor, ActorRef, Props }
+import kamon.metric.SubscriptionsDispatcher.TickMetricSnapshot
+import kamon.metric.instrument._
+
+/**
+ * Can be used as a decorator to scale TickMetricSnapshot messages to given `timeUnits` and/or `memoryUnits`
+ * before forwarding to original receiver
+ * @param timeUnits Optional time units to scale time metrics to
+ * @param memoryUnits Optional memory units to scale memory metrics to
+ * @param receiver Receiver of scaled metrics snapshot, usually a backend sender
+ */
+class MetricScaleDecorator(timeUnits: Option[Time], memoryUnits: Option[Memory], receiver: ActorRef) extends Actor {
+  require(timeUnits.isDefined || memoryUnits.isDefined,
+    "Use MetricScaleDecorator only when any of units is defined")
+
+  override def receive: Receive = {
+    case tick: TickMetricSnapshot ⇒
+      val scaled = tick.copy(metrics = tick.metrics.mapValues { entitySnapshot ⇒
+        new DefaultEntitySnapshot(entitySnapshot.metrics.map {
+          case (metricKey, metricSnapshot) ⇒
+            val scaledSnapshot = (metricKey.unitOfMeasurement, timeUnits, memoryUnits) match {
+              case (time: Time, Some(to), _)     ⇒ metricSnapshot.scale(time, to)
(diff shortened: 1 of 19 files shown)


dpsoft:
@jozic looks good to me! @ivantopo WDYT?

kamon-core/src/main/scala/kamon/metric/instrument/Counter.scala
```diff
@@ -33,6 +33,7 @@ object Counter {
   trait Snapshot extends InstrumentSnapshot {
     def count: Long
     def merge(that: InstrumentSnapshot, context: CollectionContext): Counter.Snapshot
+    def scale(from: UnitOfMeasurement, to: UnitOfMeasurement): Counter.Snapshot
```
ivantopo: Not sure if this should be part of InstrumentSnapshot... I clearly see histograms and gauges being scaled depending on their units, but min-max-counters and regular counters shouldn't be scaled.. wdyt?

kamon-core/src/main/scala/kamon/metric/instrument/Histogram.scala
```diff
@@ -85,6 +85,39 @@ object Histogram {
     def recordsIterator: Iterator[Record]
     def merge(that: InstrumentSnapshot, context: CollectionContext): InstrumentSnapshot
     def merge(that: Histogram.Snapshot, context: CollectionContext): Histogram.Snapshot
+
+    override def scale(from: UnitOfMeasurement, to: UnitOfMeasurement): Histogram.Snapshot =
+      new ScaledSnapshot(from, to, this)
+  }
+
+  class ScaledSnapshot(from: UnitOfMeasurement, to: UnitOfMeasurement, snapshot: Snapshot) extends Snapshot {
```
ivantopo: This I like, cool! :D

kamon-core/src/main/scala/kamon/metric/MetricScaleDecorator.scala
```diff
@@ -0,0 +1,55 @@
...
+import kamon.metric.instrument._
+
+/**
+ * Can be used as a decorator to scale TickMetricSnapshot messages to given `timeUnits` and/or `memoryUnits`
+ * before forwarding to original receiver
+ * @param timeUnits Optional time units to scale time metrics to
+ * @param memoryUnits Optional memory units to scale memory metrics to
+ * @param receiver Receiver of scaled metrics snapshot, usually a backend sender
+ */
+class MetricScaleDecorator(timeUnits: Option[Time], memoryUnits: Option[Memory], receiver: ActorRef) extends Actor {
```
ivantopo: What about merging this functionality with that of `TickMetricSnapshotBuffer`? so that applying the scales would be just another config to the snapshot buffer? Just thinking of it, not a requirement.. WDYT?

kamon-datadog/src/main/resources/reference.conf
```diff
@@ -37,6 +37,16 @@ kamon {
     # this pattern:
     #    application.entity-name.metric-name
     application-name = "kamon"
+
+    # All time values are collected in nanoseconds,
+    # to scale before sending to datadog set "time-units" to "s" or "ms" or "µs".
+    # Value "ns" is equivalent to omitting the setting
+    # time-units = "ns"
```
ivantopo: Why are these values and the ones in the datadog reference file commented out? Shouldn't they be always there?

ivantopo:
hey @jozic! Again, thanks for such a great job! There are just a few things to settle but you got the right approach :).. I'm sorry that I have been a bit away (well, more than a bit to be honest), but let's get this done!

kamon-core/src/main/scala/kamon/metric/instrument/Counter.scala
jozic: can't one have, say, regular counters with Memory or Time units?

kamon-core/src/main/scala/kamon/metric/MetricScaleDecorator.scala
jozic: what if one want to scale, but not buffer?

kamon-datadog/src/main/resources/reference.conf
jozic: if scaling is not needed (values are commented), then there's one actor less in the chain and scaling from 'ns' to 'ns' is omitted. I see this as a nice feature, but it does add more executable code, so i thought it's good if people decide on their own if they want it or not. If they want - they set proper values, not defaults that do nothing.
does it make sense?

jozic:
no problem @ivantopo :)

jozic:
@ivantopo ping

dpsoft:
hey @jozi!!! Again, thanks for such a great job! and we discussed with @ivantopo  and we think that the PR is really good,  then we will merge... 

BTW: waiting for docs.

PR #298, opened 2015-12-24
= statsd, datadog: fix time unit naming in reference.conf

jozic:
while updating docs for kamon.io I noticed that I have used `ns` instead of `n` for nano seconds in #294 . This patch fixes the issue


dpsoft:
@jozic thanks!!

PR #304, opened 2016-01-19
= Upgrade jrudolph/sbt-dependency-graph to autoplugin version

lustefaniak:
Latest version of jrudolph/sbt-dependency-graph allows installing globally as it is now AutoPlugin.


dpsoft:
@lustefaniak thanks!!!

PR #305, opened 2016-01-20
= system-metrics: apply metric filters

jcazevedo:
The metric filters regarding the `system-metric` entity category weren't being properly applied. This PR enables that by adding calls to the `shouldTrack` method before registering the recorders.


dpsoft:
@jcazevedo thanks!!!

Issue #101, opened 2014-10-27
Provide actor-group metrics


ivantopo:
Currently we have RouterMetrics which consolidate processing-time and time-in-mailbox metrics for all the routees of a given routers and it seems like a good idea to provide something similar for use with the work-queue/workers solution in which there is a number of workers that typically only process one message and then ask for the next piece of work and having a unified view of all actors is better than having individual metrics. There is a short conversation in the mailing list that started this idea: https://groups.google.com/forum/#!topic/kamon-user/7woEIS0j_KQ

cheekychops:
The problem with actor name patterns (as currently implemented) is that they will not catch any new per-request actors that may be added to the system under different paths. This is a major concern for us as I don't expect every developer who adds an actor to the system to understand the details of what might break new relic! I would also like our Kamon config to be fairly static, and not have to keep changing as new actors are defined.

What I feel we need is a way to configure a regex, or multiple, which when matched against an actor name tells us that the actor is a per-request actor. For example, if you use akka's default naming for your actors, it will contain at least one $ in its name, so that could be the default config. In my project we don't use that naming scheme, so would need to override this default. 

Also, once we have detected the per-request actor we will need to do extra normalisation of its name to get something that we can report on. That again will depend on the naming scheme implemented.

frossi85:
@dpsoft @ivantopo Could you provide key files where the logic reside to help with a PR in my free time?

dpsoft:
hey @frossi85 you need dig in the [Instrumentation] and [ActorInstrumentation] as starting point!!
@ivantopo WDYT?


[Instrumentation]: https://github.com/kamon-io/Kamon/tree/master/kamon-akka/src/main/scala/kamon/akka/instrumentation 

[ActorInstrumentation]: https://github.com/kamon-io/Kamon/blob/master/kamon-akka/src/main/scala/kamon/akka/instrumentation/ActorInstrumentation.scala

ivantopo:
I think there is a question to make before going to details: should we allow actors to be tracked as both regular actor/router AND in actor groups? the current instrumentation ensures that any actor is only tracked as one thing: either actor or router.. what happens if we put actor groups, should it then be one of the three options or do we make groups an independent concept?

sebigavril:
@ivantopo i think `one of the three options` should be `one of the tow options`, right?

[...]

PR #321, opened 2016-03-07
+ kamon-akka: support for custom Entity factory for actors

hsun-cnnxty:
This is similar to #101, but we are looking for a more flexible way to customize how the entity is created for Actors.  The proposed implementation allows user to define a custom entity factory and default to existing behavior if none is provided. 

-thanks

### DIFF
--- a/kamon-akka/src/main/scala/kamon/akka/AkkaExtension.scala
+++ b/kamon-akka/src/main/scala/kamon/akka/AkkaExtension.scala
@@ -16,11 +16,40 @@
 
 package kamon.akka
 
+import _root_.akka.actor._
 import com.typesafe.config.Config
-import kamon.Kamon
+import kamon._
+import kamon.metric.Entity
+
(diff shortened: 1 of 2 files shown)


ivantopo:
hello there! May I ask why would you like to provide your own entity factory?... besides that, there is some work in a [branch](https://github.com/kamon-io/Kamon/tree/issue%23271/fix-balancing-pool-metrics) that includes a lot of changes to the Akka module, mostly related to the instrumentation... I should be merging that in the following days and I would like to pull any other changes after that is done, hope you don't mind.

hsun-cnnxty:
@ivantopo I have two use cases for it

1) our app has a few "actor groups". One example is an accounting group consists of many actor instances named as "account-actor-<store-id>". We would like to group the metrics of all the actors in one group into a single data point.

2) our monitoring app prefers to have JMX MBeans following our internal naming standards instead of those defaults from Kamon

Still in the early stage of development, so we can wait for your next release.

-thanks

PR #323, opened 2016-03-14
introduce selective instrumentation for akka actors.

ivantopo:
@dpsoft please take a look at this branch :).


dpsoft:
@ivantopo LGTM!!!

PR #333, opened 2016-04-01
+ kamon-core: introduce finishWithError(Throwable) for Traces and Segments

dpsoft:
@ivantopo please take a look.

### DIFF
--- a/kamon-core/src/main/scala/kamon/metric/TraceMetrics.scala
+++ b/kamon-core/src/main/scala/kamon/metric/TraceMetrics.scala
@@ -24,7 +24,7 @@ class TraceMetrics(instrumentFactory: InstrumentFactory) extends GenericEntityRe
    *  Records blah blah
    */
   val elapsedTime = histogram("elapsed-time", unitOfMeasurement = Time.Nanoseconds)
-  val traceErrors = counter("trace-errors")
+  val errors = counter("errors")
 }
 
 object TraceMetrics extends EntityRecorderFactory[TraceMetrics] {
@@ -41,7 +41,7 @@ class SegmentMetrics(instrumentFactory: InstrumentFactory) extends GenericEntity
    *  Records blah blah
    */
   val elapsedTime = histogram("elapsed-time", unitOfMeasurement = Time.Nanoseconds)
-  val segmentErrors = counter("segment-errors")
+  val errors = counter("errors")
 }
 
 object SegmentMetrics extends EntityRecorderFactory[SegmentMetrics] {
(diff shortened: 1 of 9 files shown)


kamon-core/src/main/scala/kamon/metric/TraceMetrics.scala
```diff
@@ -24,6 +24,7 @@ class TraceMetrics(instrumentFactory: InstrumentFactory) extends GenericEntityRe
    *  Records blah blah
    */
   val elapsedTime = histogram("elapsed-time", unitOfMeasurement = Time.Nanoseconds)
+  val traceErrors = counter("trace-errors")
```
ivantopo: the name should be just `errors`, as it is already in a trace entity, same bellow for the segments.

kamon-core/src/main/scala/kamon/trace/MetricsOnlyContext.scala
```diff
@@ -52,11 +54,21 @@ private[kamon] class MetricsOnlyContext(traceName: String, val token: String, iz
     val traceElapsedTime = NanoInterval.since(startTimestamp)
     _elapsedTime = traceElapsedTime
 
-    if (Kamon.metrics.shouldTrack(name, TraceMetrics.category))
-      Kamon.metrics.entity(TraceMetrics, name).elapsedTime.record(traceElapsedTime.nanos)
+    if (Kamon.metrics.shouldTrack(name, TraceMetrics.category)) {
+      val traceEntity = Kamon.metrics.entity(TraceMetrics, name)
```
ivantopo: we should get the tags PR merged first, then rebase this branch and ensure that this uses tags as well.

kamon-core/src/main/scala/kamon/trace/MetricsOnlyContext.scala
```diff
@@ -31,6 +31,7 @@ private[kamon] class MetricsOnlyContext(traceName: String, val token: String, iz
 
   @volatile private var _name = traceName
   @volatile private var _isOpen = izOpen
+  @volatile private var _isFinishedWithError = false
```
ivantopo: don't add this variable, this extra bit of state is not necessary since we could just add a parameter to finish that says "this finished as error" and have the previous parameter-less version of finish() to call the new assuming the trace was successful.

ivantopo:
`isFinishedWithError` should be out of the TraceContext interface and we should also take out the `elapsedTime` and `startTimestamp`.. I think they are legacy from before we had the traces subscription protocol. And, the classes used in the simple trace subscriptions should also include the error indicator. 

I even thought that instead of having the isOpen and isFinishedWithError we could have a single variable called status where we have Open/FinishedWithError/FinishedSuccessfully, wdyt?

PR #334, opened 2016-04-01
Add tags for traces and closes #327

### LINKED ISSUE #327: Add tags for traces
Currently kamon.trace.withNewContext does not take any tags map, like kamon.metric.Counter does.

dpsoft:
@ivantopo take a look


ivantopo:
the API looks fine but we need to do more changes there: we should be able to add/remove tags at any point while the trace is still open, otherwise we wont be able to add tags from Spray directives as the trace is already generated on the spray-can side before we get to the routes. Take a look at what is there with the metadata, we will need to do something pretty similar.

ivantopo:
and also, segments should also have tags!

PR #335, opened 2016-04-01
core: catch any exception being thrown when recording values on histograms

### LINKED ISSUE #232: Histogram recorded value cannot be negative.
I noticed the following error shortly after starting up one of my scala services running in a docker container on OSX.

I do not know how to reproduce it but if there are any steps I can take to help troubleshoot the issue please let me know.

java.lang.ArrayIndexOutOfBoundsException: Histogram recorded value cannot be negative.
dispatcher_1       | 	at org.HdrHistogram.AbstractHistogram.countsArrayIndex(AbstractHistogram.java:1944)
dispatcher_1       | 	at org.HdrHistogram.AbstractHistogram.recordSingleValue(AbstractHistogram.java:414)
dispatcher_1       | 	at org.HdrHistogram.AbstractHistogram.recordValue(AbstractHistogram.java:333)
dispatcher_1       | 	at kamon.metric.instrument.HdrHistogram.record(Histogram.scala:115)
dispatcher_1       | 	at akka.kamon.instrumentation.ActorCellInstrumentation.aroundBehaviourInvoke(ActorCellInstrumentation.scala:68)
dispatcher_1       | 	at akka.actor.ActorCell.invoke(ActorCell.scala:483)
dispatcher_1       | 	at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:254)
dispatcher_1       | 	at akka.dispatch.Mailbox.run(Mailbox.scala:221)
dispatcher_1       | 	at akka.dispatch.Mailbox.exec(Mailbox.scala:231)
dispatcher_1       | 	at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
dispatcher_1       | 	at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
dispatcher_1       | 	at scala.concurrent.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
dispatcher_1       | 	at scala.concurrent.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107)

### LINKED ISSUE #332: ArrayIndexOutOfBoundsException: value outside of histogram covered range
I recieved the following exception:

    java.lang.ArrayIndexOutOfBoundsException: value outside of histogram covered range. Caused by: java.lang.IndexOutOfBoundsException: index 4614
        at org.HdrHistogram.AbstractHistogram.handleRecordException(AbstractHistogram.java:441)
        at org.HdrHistogram.AbstractHistogram.recordSingleValue(AbstractHistogram.java:433)
        at org.HdrHistogram.AbstractHistogram.recordValue(AbstractHistogram.java:346)
        at kamon.metric.instrument.HdrHistogram.record(Histogram.scala:146)
        at kamon.trace.MetricsOnlyContext.finish(MetricsOnlyContext.scala:56)

Dependencies:

    object Versions {
      val akka = "2.4.2"
      val kamon = "0.6.0"
      val aspectjweaver = "1.8.6" // Should this be 1.8.8 ?
    }
    ...
    "io.kamon" %% "kamon-core" % Versions.kamon,
    "io.kamon" %% "kamon-scala" % Versions.kamon,
    "io.kamon" %% "kamon-akka" % Versions.kamon,
    "io.kamon" %% "kamon-newrelic" % Versions.kamon, 

This application is a long-running ETL that only runs for ~ 70 minutes before crashing due to this error.  We've configured Kamon to time our blocking IO as individual transactions reported to newrelic to help us understand inter-system performance characteristics.

I actually see this as being two bugs.  First is the IndexOutOfBoundsExeption.  Second, and more importantly, I think Kamon should re-evaluate its approach to error handling.  I've been shocked twice by the way Kamon caused my application to simply not work because of its own needs and behaviors (this error and the need to start Kamon at startup, even with dependencies and when testing)  My opinion is Kamon should never cause an application to degrade its service levels, logging its own failures (without spamming) while allowing the application to continue doing what it needs to do.  By its nature, Kamon is an observer of the system, but should go to great lengths to stay out of the applications way even if things go wrong.

ivantopo:
fixes #232, fixes #332

### DIFF
--- a/kamon-core/src/main/scala/kamon/metric/instrument/Histogram.scala
+++ b/kamon-core/src/main/scala/kamon/metric/instrument/Histogram.scala
@@ -20,6 +20,7 @@ import java.nio.LongBuffer
 
 import kamon.metric.instrument.Histogram.{ DynamicRange, Snapshot }
 import org.HdrHistogram.ModifiedAtomicHistogram
+import org.slf4j.LoggerFactory
 
 trait Histogram extends Instrument {
   type SnapshotType = Histogram.Snapshot
@@ -135,6 +136,10 @@ object Histogram {
   }
 }
 
+object HdrHistogram {
+  private val log = LoggerFactory.getLogger(classOf[HdrHistogram])
+}
+
 /**
  *  This implementation is meant to be used for real time data collection where data snapshots are taken often over time.
  *  The collect(..) operation extracts all the recorded values from the histogram and resets the counts, but still
@@ -142,10 +147,20 @@ object Histogram {
  */
 class HdrHistogram(dynamicRange: DynamicRange) extends ModifiedAtomicHistogram(dynamicRange.lowestDiscernibleValue,
   dynamicRange.highestTrackableValue, dynamicRange.precision) with Histogram {
+  import HdrHistogram.log
 
-  def record(value: Long): Unit = recordValue(value)
+  def record(value: Long): Unit = tryRecord(value, 1L)
 
-  def record(value: Long, count: Long): Unit = recordValueWithCount(value, count)
+  def record(value: Long, count: Long): Unit = tryRecord(value, count)
+
+  private def tryRecord(value: Long, count: Long): Unit = {
+    try {
+      recordValueWithCount(value, count)
+    } catch {
+      case anyException: Throwable ⇒
+        log.warn("Failed to store value {} in HdrHistogram, please review your range configuration.")
+    }
+  }
 
   def collect(context: CollectionContext): Histogram.Snapshot = {
     import context.buffer
--- a/kamon-core/src/test/scala/kamon/metric/instrument/HistogramSpec.scala
+++ b/kamon-core/src/test/scala/kamon/metric/instrument/HistogramSpec.scala
@@ -32,10 +32,8 @@ class HistogramSpec extends WordSpec with Matchers {
       histogram.record(10000)
     }
 
-    "fail when recording values higher than the highest trackable value" in new HistogramFixture {
-      intercept[IndexOutOfBoundsException] {
-        histogram.record(1000000)
-      }
+    "not fail when recording values higher than the highest trackable value" in new HistogramFixture {
+      histogram.record(Long.MaxValue)
     }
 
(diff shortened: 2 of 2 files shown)


ivantopo:
maybe we can record the max allowed value in a histogram when we hit the out of range exception? I think it might be better to have a huge value than not to have the value at all. What do @dpsoft and @mladens say?

kamon-core/src/main/scala/kamon/metric/instrument/Histogram.scala
```diff
@@ -135,6 +136,10 @@ object Histogram {
   }
 }
 
+object HdrHistogram {
+  private val log = LoggerFactory.getLogger(classOf[HdrHistogram])
```
dpsoft: I think that we can use our LazyLogger instead of LoggerFactory ;):

 ```private val log = kamon.util.logger.LazyLogger(classOf[HdrHistogram])```

dpsoft:
Maybe could be a good idea if before record the max allowed value,  we can show a message in the log with this behavior.

kamon-core/src/main/scala/kamon/metric/instrument/Histogram.scala
```diff
@@ -135,17 +136,31 @@ object Histogram {
...
-  def record(value: Long): Unit = recordValue(value)
+  def record(value: Long, count: Long): Unit = tryRecord(value, count)
 
-  def record(value: Long, count: Long): Unit = recordValueWithCount(value, count)
+  private def tryRecord(value: Long, count: Long): Unit = {
+    try {
+      recordValueWithCount(value, count)
+    } catch {
+      case anyException: Throwable ⇒
+        log.warn("Failed to store value {} in HdrHistogram, please review your range configuration.")
```
eiennohito: value is not passed to logger

dpsoft:
@eiennohito thanks for reporting this. I've fixed this with https://github.com/kamon-io/Kamon/commit/243658cffdd10f331334962f83b56e7647a42c64

PR #338, opened 2016-04-04
= kamon-autoweave: Remove usage of `breakable`

markus1189:
Replaces the usage of `breakable` with a better version using `.iterator`.  Just had to do this ;)


dpsoft:
@markus1189 thanks for this PR!!!

PR #339, opened 2016-04-08
+ Kamon module for Akka Sharding metrics

nilsga:
Kamon module that collects metrics for Akka Sharding.

### DIFF
--- a/kamon-akka-sharding/src/main/scala/kamon/akka/sharding/instrumentation/AkkaShardingInstrumentation.scala
+++ b/kamon-akka-sharding/src/main/scala/kamon/akka/sharding/instrumentation/AkkaShardingInstrumentation.scala
@@ -1,0 +1,77 @@
+package akka.kamon.instrumentation
+
+import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.{ Executors, ThreadFactory, TimeUnit }
+
+import akka.actor.{ ActorRef, ExtendedActorSystem, Props }
+import akka.cluster.sharding.ShardCoordinator.ShardAllocationStrategy
+import akka.cluster.sharding.ShardRegion.{ CurrentShardRegionState, ExtractEntityId, ExtractShardId, GetShardRegionState }
+import akka.cluster.sharding.{ ClusterSharding, ClusterShardingSettings }
+import akka.pattern._
+import akka.util.Timeout
+import com.typesafe.config.Config
+import kamon.Kamon
+import kamon.akka.sharding.metrics.AkkaShardMetrics
+import kamon.metric.Entity
+import org.aspectj.lang.annotation._
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
+
+object AkkaShardingInstrumentation {
+  implicit lazy val ec: ExecutionContext = {
+    val threadFactory = new ThreadFactory {
+
+      val counter: AtomicInteger = new AtomicInteger()
+      val group: ThreadGroup = new ThreadGroup("kamon-akka-sharding")
+
+      override def newThread(r: Runnable): Thread =
+        new Thread(group, r, s"metrics-refresher-${counter.getAndIncrement()}")
+    }
+    ExecutionContext.fromExecutor(Executors.newFixedThreadPool(5, threadFactory))
+  }
+  implicit val sender: ActorRef = ActorRef.noSender
+  implicit val timeout: Timeout = 10 seconds
+
+  val config: Config = Kamon.config.getConfig("kamon.akka-sharding")
+
+  implicit def toFiniteDuration(d: java.time.Duration): FiniteDuration = FiniteDuration(d.toNanos, TimeUnit.NANOSECONDS)
+
+}
+
+@Aspect
+class AkkaShardingInstrumentation {
+
+  @Pointcut("execution(* akka.cluster.sharding.ClusterSharding.start(..)) && this(clusterSharding) && args(typeName, entityProps, settings, extractEntityId, extractShardId, allocationStrategy, handOffStopMessage)")
+  def constructShardRegion(clusterSharding: ClusterSharding, typeName: String, entityProps: Props, settings: ClusterShardingSettings, extractEntityId: ExtractEntityId, extractShardId: ExtractShardId, allocationStrategy: ShardAllocationStrategy, handOffStopMessage: Any): Unit = {}
+
+  @AfterReturning(pointcut = "constructShardRegion(clusterSharding, typeName, entityProps, settings, extractEntityId, extractShardId, allocationStrategy, handOffStopMessage)", returning = "shardRegion")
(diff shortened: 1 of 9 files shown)


kamon-akka-sharding/src/main/scala/kamon/akka/sharding/instrumentation/AkkaShardingInstrumentation.scala
```diff
@@ -0,0 +1,77 @@
...
+import kamon.Kamon
+import kamon.akka.sharding.metrics.ShardMetrics
+import kamon.metric.Entity
+import org.aspectj.lang.annotation._
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
+
+object AkkaShardingInstrumentation {
+  implicit lazy val ec: ExecutionContext = {
```
nilsga: I had to create a dedicated `ExecutionContext`, since I could not access the default dispatcher from the Kamon actor system.

nilsga:
I had to use `kamon-java8` as base since the metrics for akka cluster sharding is only implemented in akka 2.4.

kamon-akka-sharding/src/main/scala/kamon/akka/sharding/instrumentation/AkkaShardingInstrumentation.scala
```diff
@@ -0,0 +1,77 @@
...
+  @AfterReturning(pointcut = "constructShardRegion(clusterSharding, typeName, entityProps, settings, extractEntityId, extractShardId, allocationStrategy, handOffStopMessage)", returning = "shardRegion")
+  def afterShardRegionConstructed(clusterSharding: ClusterSharding, typeName: String, entityProps: Props, settings: ClusterShardingSettings, extractEntityId: ExtractEntityId, extractShardId: ExtractShardId, allocationStrategy: ShardAllocationStrategy, handOffStopMessage: Any, shardRegion: ActorRef): Unit = {
+    val trackedSystem = clusterSharding.asInstanceOf[ActorSystemAware].actorSystem
+    val entity = Entity(s"${trackedSystem.name}/$typeName", AkkaShardMetrics.ShardingCategoryName)
+    if (Kamon.metrics.shouldTrack(entity)) {
+      val shardEntity = Kamon.metrics.entity(AkkaShardMetrics, entity)
+      import AkkaShardingInstrumentation._
+      val scheduler = Kamon.metrics.settings.refreshScheduler
+      val interval: FiniteDuration = config.getDuration("refresh-interval")
+      scheduler.schedule(interval, () ⇒ {
```
nilsga: I was thinking of collecting the `Cancellable`s created by the scheduler, but I'm not sure if that's necessary, given that the configuration is static and it's not possible to change the Kamon configuration in runtime.

kamon-akka-sharding/src/main/scala/kamon/akka/sharding/instrumentation/AkkaShardingInstrumentation.scala
```diff
@@ -0,0 +1,77 @@
...
+  @Pointcut("initialization(akka.cluster.sharding.ClusterSharding.new(..)) && target(clusterSharding) && args(system)")
+  def clusterShardingCreation(clusterSharding: ClusterSharding, system: ExtendedActorSystem): Unit = {}
+
+  @After("clusterShardingCreation(clusterSharding, system)")
+  def afterClusterShardingCreation(clusterSharding: ClusterSharding, system: ExtendedActorSystem): Unit = {
+    clusterSharding.asInstanceOf[ActorSystemAware].actorSystem = system
+  }
+
+  @DeclareMixin("akka.cluster.sharding.ClusterSharding")
+  def clusterShardingActorSystem: ActorSystemAware = ActorSystemAware()
```
nilsga: Following the pattern in `DispatcherInstrumentation` in `kamon-akka`.

kamon-akka-sharding/src/main/scala/kamon/akka/sharding/instrumentation/AkkaShardingInstrumentation.scala
```diff
@@ -0,0 +1,77 @@
+package akka.kamon.instrumentation
```
nilsga: I did note that the package and the directory is not the same, but I just followed the same pattern as the instrumentation in the other modules.

kamon-akka-sharding/src/main/scala/kamon/akka/sharding/instrumentation/AkkaShardingInstrumentation.scala
```diff
@@ -0,0 +1,77 @@
...
+    val entity = Entity(s"${trackedSystem.name}/$typeName", AkkaShardMetrics.ShardingCategoryName)
+    if (Kamon.metrics.shouldTrack(entity)) {
+      val shardEntity = Kamon.metrics.entity(AkkaShardMetrics, entity)
+      import AkkaShardingInstrumentation._
+      val scheduler = Kamon.metrics.settings.refreshScheduler
+      val interval: FiniteDuration = config.getDuration("refresh-interval")
+      scheduler.schedule(interval, () ⇒ {
+        val shardStateFuture = shardRegion ? GetShardRegionState
+        shardStateFuture.mapTo[CurrentShardRegionState].foreach(state ⇒ {
+          shardEntity.record(state.shards)
```
nilsga: After some more testing, this doesn't seem to work. It does not update the current value of the `CurrentValueCollector`. So when the value is refreshed by the internal gauge refresh schedule, it will record the initial value for all subsequent invocations. I guess I either need to create a `Gauge` that is never refreshed (only recorded), or create a custom `CurrentValueCollector` that can be updated from within this method.

kamon-akka-sharding/src/main/scala/kamon/akka/sharding/metrics/AkkaShardMetrics.scala
```diff
@@ -0,0 +1,24 @@
...
+
+class AkkaShardMetrics(instrumentFactory: InstrumentFactory) extends GenericEntityRecorder(instrumentFactory) {
+
+  val numShardsGauge = gauge("number-of-shards", 0L)
+
+  def record(shardStats: Set[ShardState]) = {
+    numShardsGauge.record(shardStats.size)
+    shardStats.foreach(state ⇒ {
+      val shardGauge = gauge(s"number-of-entities-shard-${state.shardId}", 0L)
+      shardGauge.record(state.entityIds.size)
```
nilsga: After some more testing, this doesn't seem to work. It does not update the current value of the `CurrentValueCollector`. So when the value is refreshed by the internal gauge refresh schedule, it will record the initial value for all subsequent invocations. I guess I either need to create a `Gauge` that is never refreshed (only recorded), or create a custom `CurrentValueCollector` that can be updated from within this method.

nilsga: Also, doesn't take into consideration rebalancing of shards, so gauges will never be removed.

PR #353, opened 2016-05-12
Replace AspectJ Weaver error message

adlawson:
I know it's bad if AspectJ Weaver isn't running when it should be, but I don't think it's cool to draw such big attention to it in the error logs.

```
  _______ _                 _        
 |__   __| |               | |       
    | |  | |__   __ _ _ __ | | _____ 
    | |  | '_ \ / _` | '_ \| |/ / __|
    | |  | | | | (_| | | | |   <\__ \
    |_|  |_| |_|\__,_|_| |_|_|\_\___/

```

### DIFF
--- a/kamon-core/src/main/scala/kamon/ModuleLoader.scala
+++ b/kamon-core/src/main/scala/kamon/ModuleLoader.scala
@@ -58,14 +58,7 @@ private[kamon] class ModuleLoaderExtension(system: ExtendedActorSystem) extends 
     val weaverMissingMessage =
       """
         |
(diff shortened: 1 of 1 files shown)


ivantopo:
hey @adlawson, sorry for the long long wait. I personally also find it annoying, but that was made big on purpose, probably the most common issue we have with our users is not having the AspectJ weaver setup correctly and this definitely makes it quite difficult to have that error go unnoticed, and still, the warning can be disabled if wanted. 

That made me think of something, though.. we should always show the warning if the weaver is missing, but let the user select between big and small notice, having the big one by default.

Thanks for your help!

PR #363, opened 2016-06-18
= kamon-core: add context name and token to MDC

dspasojevic:
This change pushes the trace name and token into the MDC. This means that they can be accessed in async appenders using the normal mechanism - without using the LogbackTraceTokenConverter.

### DIFF
--- a/kamon-core/src/main/scala/kamon/trace/logging/MdcKeysSupport.scala
+++ b/kamon-core/src/main/scala/kamon/trace/logging/MdcKeysSupport.scala
@@ -17,12 +17,17 @@
 package kamon.trace.logging
 
 import kamon.trace.TraceLocal.AvailableToMdc
-import kamon.trace.{ Tracer, EmptyTraceContext, MetricsOnlyContext, TraceContext }
+import kamon.trace.{ EmptyTraceContext, MetricsOnlyContext, TraceContext, Tracer }
 import kamon.util.Supplier
-
 import org.slf4j.MDC
 
 trait MdcKeysSupport {
+
+  val traceTokenKey = "traceToken"
+
+  val traceNameKey = "traceName"
+
+  private val defaultKeys = Seq(traceTokenKey, traceNameKey)
 
   def withMdc[A](thunk: ⇒ A): A = {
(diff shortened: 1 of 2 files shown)


kamon-core/src/main/scala/kamon/trace/logging/MdcKeysSupport.scala
```diff
@@ -34,7 +39,12 @@ trait MdcKeysSupport {
 
   private[this] def copyToMdc(traceContext: TraceContext): Iterable[String] = traceContext match {
     case ctx: MetricsOnlyContext ⇒
-      ctx.traceLocalStorage.underlyingStorage.collect {
+
+      // Add the default key value pairs for the trace token and trace name.
```
dpsoft: @dspasojevic @ivantopo guys I think that instead of using:
```scala

val traceTokenKey = "traceToken"
val traceNameKey = "traceName"

private val defaultKeys = Seq(traceTokenKey, traceNameKey)

MDC.put(traceTokenKey, ctx.token)
MDC.put(traceNameKey, ctx.name)
 ```
we can use the kamon api like this:

```scala
val TraceTokenKey = AvailableToMdc.fromKey("traceToken")
val TraceNameKey = AvailableToMdc.fromKey("traceName")

TraceLocal.store(TraceTokenKey)(ctx.token)
TraceLocal.store(TraceNameKey)(ctx.name)
```
this is only a idea, please guys WDYT?

dspasojevic: That is fine with me.

It does mean that calling `withMdc` has the side-effect of modifying TraceLocal's storage though. At the moment, it should return the `MDC` and TraceLocal storage to the initial state.

dpsoft: I see your point, and I think that is a good point :). @ivantopo?

dspasojevic:
@dpsoft are there any changes that you would like to this PR?

dpsoft:
@dspasojevic thanks for proposing this PR!!!! Done.

PR #364, opened 2016-06-27
Add Slf4J logging to use Marker for tagged logs.

osbornk:
We have a need to import tagged metrics into LogStash over SLf4J. This is basically the existing logreporter, except that is uses SLf4J Markers to create the tag.

It is now controlled by two flags: kamon.logreporter.formatted-slf4j (default false) and kamon.logreporter.prettyprint-log (default true). These control whether logreporter will use the existing Akka logging (more asynchronous, but missing features) or a manually built logger.

Of course, the user will now need to do some more manual work to get SLf4J working properly with Akka, but that is up to the user (http://doc.akka.io/docs/akka/2.4.4/scala/logging.html#Using_the_SLF4J_API_directly).


osbornk:
Looks like the tests are failing, but I don't think it is due to any code that I touched.

osbornk:
Could I get some feedback on this pull request. Thanks.

dpsoft:
@osbornk sorry for the mess, please give me a couple of days and I will be back with some feedback.

osbornk:
May I ask why this was closed? The ability to send Kamon stats over SLF4J is definitely something we need.

dpsoft:
@osbornk sorry man, we deleted the master repository and it was replaced with other branch and I think that was a side effect. :(. Please you can open again?

osbornk:
No problem. I will move this PR over to the new branch after the holidays.

PR #370, opened 2016-07-05
Add support for twitter-util Futures

yanana:
I searched a conversation about twitter-util Futures in the kamon-user mailing list, but I couldn't find it. So I implemented the support for them.


dpsoft:
@yanana thanks for proposing this PR!!!

Issue #379, opened 2016-07-28
Reading current tags in TraceContext


elm-:
Hello,

is there a reason why it is not possible to read the current tags / metadata in the context? If so, would there be something against adding it via a PR?

Best,
Elmar

dpsoft:
@elm- I think that doesn't exist a reason to not read the current tags. Please feel free to make a PR.

dpsoft:
related to https://github.com/kamon-io/Kamon/pull/386.

PR #386, opened 2016-08-12
+ core: allow accessing tags on TraceContext and Segment

bfil:
This change would allow accessing the tags defined on a TraceContext or a Segment.


dpsoft:
@bfil thanks proposing this PR. done!. closes #379

PR #388, opened 2016-08-22
+ core: support for regex path filters

danbim:
In my application I had a rather complicated use case where I had to exclude anonymous actors (created for Akka asks) from generating Kamon metrics. I found this very cumbersome to do based on the GlobPathFilter so I created this PR.

Putting it up here for discussion (e.g. to see if you're interested). Obviously some things are left to be done (e.g. end-user documentation).

### DIFF
--- a/kamon-core/src/main/scala/kamon/util/PathFilter.scala
+++ b/kamon-core/src/main/scala/kamon/util/PathFilter.scala
@@ -1,0 +1,5 @@
+package kamon.util
+
+trait PathFilter {
+  def accept(path: String): Boolean
+}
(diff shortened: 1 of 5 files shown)


kamon-core/src/main/scala/kamon/util/PathFilter.scala
```diff
@@ -0,0 +1,5 @@
+package kamon.util
```
dpsoft: @danbim please add the kamon copyright header.

dpsoft:
@danbim  the PR looks good for me, i've left a comment. I would like to merge it BUT there is one missing piece, not in this repository but in our [documentation website]! yes, I know, it is a bit boring to write docs but it is a necessary thing... so if you could change the [filtering-entities] documentation in oder to reflect this change it would be amazing.

[filtering-entities]:http://kamon.io/core/metrics/core-concepts/#filtering-entities

danbim:
I'll update the docs soon!

dpsoft:
@danbim thanks for proposing this PR!!!

Issue #393, opened 2016-09-20
Counter.collect is not thread safe


isaacl:
It looks like the Kamon Counter uses an unsafe method to reset and calculate sums. `collect()` uses `LongAdder.sumThenReset()`, and this method not safe during contention.  In particular, these lines:
```scala
sum += a.value;
a.value = 0L;
```
can lose data if the Cell is incremented concurrently.  Looks like LongAdder needs a CAS based method?

dpsoft:
related to https://github.com/kamon-io/Kamon/pull/399

dpsoft:
@isaacl we just published the Kamon [0.6.3] version with this fix. Thanks!!

[0.6.3]:http://search.maven.org/#artifactdetails%7Cio.kamon%7Ckamon-core_2.11%7C0.6.3%7C

PR #394, opened 2016-09-20
Use CAS for safer counter collection

isaacl:
Note: this patch only fixes the scala counter. Other Kamon metrics have similar semantics which could benefit from CAS.

For #393

### DIFF
--- a/kamon-core/src/main/java/kamon/jsr166/Striped64.java
+++ b/kamon-core/src/main/java/kamon/jsr166/Striped64.java
@@ -105,6 +105,10 @@ abstract class Striped64 extends Number {
             return UNSAFE.compareAndSwapLong(this, valueOffset, cmp, val);
         }
 
+        final long getAndSet(long val) {
+            return UNSAFE.getAndSetLong(this, valueOffset, val);
+        }
+
         // Unsafe mechanics
         private static final sun.misc.Unsafe UNSAFE;
         private static final long valueOffset;
@@ -179,6 +183,13 @@ abstract class Striped64 extends Number {
      */
     final boolean casBase(long cmp, long val) {
         return UNSAFE.compareAndSwapLong(this, baseOffset, cmp, val);
+    }
+
+    /**
+     * CASes the base field.
+     */
+    final long getAndSetBase(long val) {
+        return UNSAFE.getAndSetLong(this, baseOffset, val);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


kamon-core/src/main/java/kamon/jsr166/Striped64.java
```diff
@@ -182,6 +186,13 @@ final boolean casBase(long cmp, long val) {
     }
 
     /**
+     * CASes the base field.
```
dpsoft: This code are using Java 8 Unsafe intrinsics (in this case lock xchg) and isn't supported in prior java versions.
Maybe we could check if support getAndSetObject in this way Unsafe.class.getMethod("getAndSetObject", Object.class, Long.TYPE,Object.class) and use this method, otherwise fallback to something like:

```java
 final long getAndSetBase(long val) {
  long v;
     do {
          v = UNSAFE.getLongVolatile(this, baseOffset);
      } while (!UNSAFE.compareAndSwapLong(this, baseOffset, v, val));
  return v;
 }
```
or simply use the code above (CAS version, I do not know if the code compiles ;)) and the same for the `final long getAndSet(long val)`

isaacl:
mostly sent this pull-req as a request for comment, as I was surprised to see this type of issue in Kamon. But if you're willing to review I will invest time to clean up this pull req etc.

dpsoft:
@isaacl thanks for find this issue, no problem, we can follow the issue in our side (It was a surprise for us also, this is embarrassingly) if you want?

kamon-core/src/main/java/kamon/jsr166/Striped64.java
isaacl: If getAndSetLong was added too recently I'd suggest adding the function as a static helper, as it's needed for both the base get and the cell get.

The code you wrote is identical to the [actual impl](http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/sun/misc/Unsafe.java#l1085).

dpsoft: Yes, is a good idea!

ivantopo:
hey @isaacl, it would be awesome if you can put that extra time to clean up and get this merged! :).. I'm sorry that we didn't get back to you right away, we certainly did take a look at this as soon as you proposed it because at this point, we shouldn't have this kind of bugs. I would like to move this forward as soon as possible and publish a new release with it. Thanks in advance for your effort! And, if time doesn't favor you let us know and we will take it from here, regards!

isaacl:
I should have some time in the next couple days, will follow up when the patch is cleaned up.

dpsoft:
related to https://github.com/kamon-io/Kamon/pull/399

isaacl:
Thanks for handling. Cheers guys.

PR #399, opened 2016-10-02
Cas collect compatibility with earlier jvms

ivantopo:
This PR is based on what @dpsoft and @isaacl discussed here: https://github.com/kamon-io/Kamon/pull/394, I'm just putting it in to start the discussion and getting it merged as soon as possible. We should publish a new artifact as soon as this is merged.

### DIFF
--- a/kamon-core/src/main/java/kamon/jsr166/LongAdder.java
+++ b/kamon-core/src/main/java/kamon/jsr166/LongAdder.java
@@ -151,6 +151,21 @@ public class LongAdder extends Striped64 implements Serializable {
         return sum;
     }
 
+    public long sumAndReset() {
+        long sum = getAndSetBase(0L);
+        Cell[] as = cells;
+        if (as != null) {
+            int n = as.length;
+            for (int i = 0; i < n; ++i) {
+                Cell a = as[i];
+                if (a != null) {
+                    sum += a.getAndSet(0L);
+                }
+            }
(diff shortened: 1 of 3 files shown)


dpsoft:
@ivantopo @isaacl LGTM, I think that for now is ok, maybe in the future is a good idea have two specialized versions of [LongAdder, LongMaxUpdater,Striped64] in order to not punish the users that using java versions > 8  and the new Unsafe intrinsics (lock addq and lock xchg) that improve the performance.

BTW: I think that in [LongMaxUpdater]::maxThenReset() we are experimenting the same.

[LongMaxUpdater]: https://github.com/kamon-io/Kamon/blob/master/kamon-core/src/main/java/kamon/jsr166/LongMaxUpdater.java#L117

dpsoft:
@ivantopo @isaacl LGTM, I think that for now is ok, maybe in the future is a good idea have two specialized versions of [LongAdder, LongMaxUpdater,Striped64] in order to not punish the users that using java versions > 8  and the new Unsafe intrinsics (lock addq and lock xchg) that improve the performance.

BTW: I think that in [LongMaxUpdater]::maxThenReset() we are experimenting the same.

[LongMaxUpdater]: https://github.com/kamon-io/Kamon/blob/master/kamon-core/src/main/java/kamon/jsr166/LongMaxUpdater.java#L117

PR #401, opened 2016-10-05
Generic jmx importing

hunterpayne-ck:
This is new functionality for Kamon.  It allows importing of custom JMX metrics.  The JMX metrics to import are specified in a config file and pumped into kamon just like any other metric.  Then those metrics can be sent to any Kamon backend.  Great for places that use Kamon and also older systems (like Kafka) that export their metrics via JMX.  Also, many other parts of Kamon could be refactored to use this module instead of using custom code to access specific JMX metrics.


kamon-jmx/src/main/scala/kamon/jmx/extension/ExportedMBeanQuery.scala
```diff
@@ -0,0 +1,362 @@
+
```
dpsoft: @hunterpayne-ck please add the kamon copyright in each file

kamon-jmx/src/main/scala/kamon/jmx/extension/MetricDefinition.scala
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * This call represents one kamon metric which can be a counter, gauge,
+ * histogram or min-max counter.  It allows a metric to be represented
+ * in a config file.
+ */
+case class MetricDefinition(
+    val metricType: MetricTypeEnum.MetricType, val name: String,
+    unitOfMeasure: UnitOfMeasurement, // = UnitOfMeasurement.Unknown,
+    range: DynamicRange = null,
+    refreshInterval: FiniteDuration = null,
```
dpsoft: why you use null values instead of Options values?

hunterpayne-ck:
Great, thank you.

Use of null is probably an oversight.  I often switch between Java and
Scala so sometimes I forget that one.
I will add the copywrite notices on my next commit.  Probably on Monday.

Also, the Kamon unit tests seem to fail intermittently on tests in modules
I haven't touched.  Is this normal?  If not, what likely am I do wrong?

Hunter

dpsoft:
@hunterpayne-ck we are some issues with the tests and travis..

hunterpayne-ck:
The requested changes have been made

hunterpayne-ck:
Ok, finally found time to make the requested changes for this PR.  Nulls
are removed (except for sanity checks from outside input) and license
headers have been added.

Hunter

dpsoft:
@hunterpayne-ck the changes are ok, but maybe is a good idea make a test. WDYT?

hunterpayne-ck:
Ok, some tests...will be a while until I can find the time to write them.  We test them at CK using an integration test against a staging monitoring stack so we haven't needed them so far.  But Kamon can't do that so we probably need unit tests here.

hunterpayne-ck:
So I've been writing a test for this PR but have run into a significant
roadblock.  The BaseKamonSpec's config isn't used to initialize Kamon so I
can't activate the new JMX Exporter module that this PR introduces from
inside of the Spec class.  I have another PR to allow Kamon to use a
provided config (PR 403).

https://github.com/kamon-io/Kamon/pull/403

Perhaps we should try to get that one merged first as it makes testing much
easier?  What does everyone think?

Hunter

hunterpayne-ck:
So I got an unit test working...but its hard to have a good test suite without being about to run kamon with different configs.  So this could use more testing but at least it has a basic positive unit test.  Until PR #403 is merged in, doing lots of negative unit tests will be difficult.   Will revisit after that happens.

hunterpayne-ck:
Hey, we got a good build...can you merge this PR now?

dpsoft:
@hunterpayne-ck  the PR looks good for me. I would like to merge it BUT there is one missing piece, not in this repository but in our [documentation website]! yes, I know, it is a bit boring to write docs but it is a necessary thing... so if you could change the [jmx-documentation] documentation in oder to reflect this change it would be amazing.

[documentation website]: https://github.com/kamon-io/kamon.io
[jmx-documentation]:http://kamon.io/backends/jmx/

hunterpayne-ck:
Can I do that in a separate PR on this branch?  We have a good build and I
don't want to waste that.

Hunter

dpsoft:
@hunterpayne-ck please make a new PR here: https://github.com/kamon-io/kamon.io

hunterpayne-ck:
Done...https://github.com/kamon-io/kamon.io/pull/27

dpsoft:
@hunterpayne-ck thanks for this PR. Done!!!

Issue #395, opened 2016-09-24
Solve the startup and configuration hell.


ivantopo:
There are a few inconveniences related to how Kamon's lifetime is handled (or poorly handled, to be more accurate):
- First, Kamon has to rely on some global state (currently inside the Kamon companion object) so that all instrumentation can easily use the APIs without having to make a big deal trying to find the right Kamon instance.
- Since there is a chance of instrumentation using some of Kamon's APIs before Kamon is effectively started, we broke the initialization in 2 steps: first is starting and configuring the tracing and metric modules, but not start the reporting modules until `Kamon.start()` is called. This is good in the sense that whenever you access any of Kamon's APIs you will be guaranteed that Kamon is started, but it took away the possibility of calling `Kamon.start(config)` which many users seem to be missing now. As a side note, we introduced config providers so that you can still kick in and provide a config object, but it might not be the most comfortable API.
- When Kamon is shutdown, it isn't shutdown very cleanly. The only thing it does is shutdown Kamon's ActorSystem, which is enough for stop reporting to the backends, but then Kamon cannot be started again. This is specially noticeable if you are running a Play! application with Kamon on Development, after the first reload Kamon will stop working. This is a very unacceptable user experience in my opinion.
- It shouldn't be required to have Kamon started for running tests. We used to force users to start Kamon to make sure that they do it only once and in the right place in the app, but when running tests it might get really tricky and in most cases it is actually unnecessary to have Kamon started.
### Proposals (WIP):

Kamon should get a proper and deterministic startup and shutdown behavior. It should be very clear what to expect if you don't start Kamon, if you start it, or if you start it a hundred times, same goes with shutdown.

The shutdown process is the simplest to approach. When Kamon is shutdown, besides shutting down the internal ActorSystem we should also make sure that Kamon's internal state is put back to as if it was never started. We might consider extending the shutdown process to signal Kamon's shutdown to all reporting modules so that they can close any related resources. (Note that this could also be achieved by hooking in the reporting module's actor lifecycle). If we shutdown and start Kamon again it should work exactly as if it was started for the first time, except that all entities that were created with the first Kamon instance will not be linked to the new instance. 

We should allow users to pass a custom config object. This is also required for Play! applications where we should always be using the config object form the context on the application loader. So, this leaves a few questions open:
[...]

PR #403, opened 2016-10-05
more-flexible-kamon-config: adding back a method to allow passing of …

hunterpayne-ck:
adding back a method to allow passing of a configuration into kamon but still allows for kamon's internal config to control the actor system responsible for metrics

### DIFF
--- a/kamon-core/src/main/scala/kamon/Kamon.scala
+++ b/kamon-core/src/main/scala/kamon/Kamon.scala
@@ -42,14 +42,16 @@ object Kamon {
 
   trait Extension extends actor.Extension
 
+  def defaultConfig = ConfigFactory.load(this.getClass.getClassLoader, ConfigParseOptions.defaults(), ConfigResolveOptions.defaults().setAllowUnresolved(true))
+
   class KamonDefaultConfigProvider extends ConfigProvider {
     def config = resolveConfiguration
 
     private def resolveConfiguration: Config = {
-      val defaultConfig = ConfigFactory.load(this.getClass.getClassLoader, ConfigParseOptions.defaults(), ConfigResolveOptions.defaults().setAllowUnresolved(true))
+      val defaultConf = defaultConfig
 
-      defaultConfig.getString("kamon.config-provider") match {
-        case "default" ⇒ defaultConfig
+      defaultConf.getString("kamon.config-provider") match {
+        case "default" ⇒ defaultConf
         case fqcn ⇒
           val dynamic = new ReflectiveDynamicAccess(getClass.getClassLoader)
           dynamic.createInstanceFor[ConfigProvider](fqcn, Nil).get.config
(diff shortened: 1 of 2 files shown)


TomDeMille:
why did they remove this in the first place?    Makes our configuration painfully awkward without.

dpsoft:
@TomDeMille take a look https://github.com/kamon-io/Kamon/issues/395

hunterpayne-ck:
I think it has to do with Kamon needing an Akka ActorSystem.  In 0.5.2 and
before Kamon reused an existing ActorSystem (I think) configured in the
standard way (the akka config object).  Kamon now configures its own actor
system and configures it from a nested configuration object in a Kamon
specific place in the config.  This is to prevent users from accidentally
mis-configuring Kamon's ActorSystem.

This PR is about restoring the ability to pass a config to Kamon without
re-introducing the issues that caused the removal in the first place.  In
this case, Kamon still configures its own actor system from the same place
in the config but can use a provided Config instead of one from
ConfigFactory.load() if the user wishes this behavior.

Hunter

hunterpayne-ck:
So this PR is ready to go now.  Any thing else I need to be before merging?

hunterpayne-ck:
This PR is ready to be merged.  Diego, can you review and let me know if
there is anything left to do on it or if you have any questions/concerns
about it?

Hunter

dpsoft:
@hunterpayne-ck please give me one or two days in order to review this.

hunterpayne-ck:
Any progress?

hunterpayne-ck:
Diego, any progress on reviewing this PR?

Hunter

dpsoft:
@hunterpayne-ck sorry for the late response, but I've been focusing on scala 2.12 support #410  :(.
In relation to the PR, I think is good for users the possibility to passing the configuration in the `Kamon::start` method. I will merge this PR and then we need focus in #395.

hunterpayne-ck:
What is still wrong w.r.t. #395?  I has hoping that #403 kept the
configuration hell from happening as the Kamon actor system config wasn't
mixed into the main actor system config even after #403.  Is there
something still wrong?

Hunter

ivantopo:
hey @hunterpayne-ck, I think that there is a bit of a misunderstanding on why the `Kamon.start(config)` was removed in the first place.. it didn't have anything to do with messing up the user's ActorSystem config with Kamon's internal one, it had to do with the fact that in certain situations (specially, when running tests) you might need to use Kamon without it being started yet because some instrumentation or some code that you can't control is calling Kamon APIs and people would just get a bunch of "Kamon not started yet" errors... it was a pain, and we changed one pain for another, by allowing Kamon to automatically start when needed for the first time and introducing the ConfigProvider so that if you needed to provide some custom config you still could.. but that definitely wasn't the best solution. 

For many people this is not a problem, at all, but if you are using Play (and a lot of people do), the experience is terrible.. Kamon stops working after the first application reload (see #395 and #421).. so, even though this PR is bringing back a feature that users want and need (don't get me wrong, we are quite happy that you proposed this! :D), we still need to put a lot of effort to make Kamon behave decently.

I will keep dumping stuff related to this on #395 and #421, as I will be primarily focusing on solving those issues ASAP, so, please stop by and bring as much feedback as you can.

And, once again, thanks for this PR! :D

PR #406, opened 2016-10-18
+ kamon-play-25: AspectJ weaving for Play's Java logging API

Skyr:
The Kamon tracing token didn't work for logging in Play projects using the Java API because the Java API uses a different logging wrapper class.
This PR adds aspectj weaving for this class from the Java API.


dpsoft:
@Skyr thanks for proposing this PR!!

PR #409, opened 2016-11-02
Update initial config code ignore unresolved configs

briantopping:
Should avoid issues where application configs require special loading behavior. Requires that the config class name is not interpolated.


dpsoft:
@briantopping thanks for proposing this PR!!! Done.

Issue #410, opened 2016-11-10
Scala 2.12 compatible release


pjan:
As per the title? Anything on the horizon? Anything blocking this?

henders:
Looks like its blocked on 2.12 releases for:
```
com.twitter#util-core
io.spray#spray-can
io.spray#spray-client
io.spray#spray-routing
net.virtual-void#json-lenses
org.fluentd#fluent-logger-scala
```

lustefaniak:
Spray 2.12.0 build is not planned: https://github.com/spray/spray/issues/1132

ivantopo:
oh boy, we need to separate our projects! We internally talked about it a lot, having each sub-project on their own repo allows them to evolve independently, we could make sure that Spray gets support for the cool features included in upcoming Kamon versions without having to have a Spray for Scala 2.12, and kamon-core could evolve without problems. The other side of that is that we will have to have some sort of compatibility rules, so that users can know what versions of sub projects work with what versions of Kamon.. it has been postponed (procrastinated, if you will) but its time to do this.

Thanks for bringing this topic up, we will do something about it!

pjan:
Maybe a great time to move toa build.sbt based project definition as well, since the project-based one will get deprecated.

Anything that the community could do with helping in this effort/these efforts?

ivantopo:
one thing that would be awesome is to be able to move things to another repositories but keeping the history of the files that will remain there.. that's something I don't know how to achieve and will need to research a bit on how to do it, do you have any ideas?

henders:
@ivantopo , You can use `git filter-branch`. Examples here: http://blogs.atlassian.com/2014/04/tear-apart-repository-git-way/

henders:
Sorry, better official example: https://help.github.com/articles/splitting-a-subfolder-out-into-a-new-repository/

dpsoft:
Tasks that we need complete in order to achieve a Scala 2.12 compatible release:
- [ ] Separate our projects in their own repositories. 
     * Separated in repos:
          * Play
          * Backends
          * Spray
          * Akka
          * Core, Annotation and Scala in Master repo?
- [ ] Move to a ```build.sbt``` based project?
- [ ] Remove Scala 2.10 support?
- [ ] ????

thougths?

pjan:
I believe only the first tick is required for 2.12 compatible releases

what would help to get this started, is if these repo's exist for starters

leachbj:
[...]

PR #416, opened 2016-12-07
Add support for building with scala 2.12 (#410)

leachbj:


project/Projects.scala
```diff
@@ -24,10 +24,10 @@ object Projects extends Build {
...
   lazy val kamon = Project("kamon", file("."))
     .enablePlugins(CrossPerProjectPlugin)
-    .aggregate(kamonCore, kamonScala, kamonAkka, kamonSpray, kamonNewrelic, kamonPlayground, kamonTestkit,
-      kamonStatsD, kamonRiemann, kamonDatadog, kamonSPM, kamonSystemMetrics, kamonLogReporter, kamonAkkaRemote, kamonJdbc, kamonElasticsearch,
-      kamonAnnotation, kamonPlay23, kamonPlay24, kamonPlay25, kamonJMXReporter, kamonFluentd, kamonKhronus,
-      kamonAutoweave, kamonInfluxDB)
+    .aggregate(kamonCore, kamonScala, kamonAkka, /*kamonSpray, kamonNewrelic,*/ kamonPlayground, kamonTestkit,
+      kamonStatsD, kamonRiemann, kamonDatadog, /*kamonSPM,*/ kamonSystemMetrics, kamonLogReporter, /*kamonAkkaRemote,*/ kamonJdbc, kamonElasticsearch,
+      kamonAnnotation, /*kamonPlay23, kamonPlay24, kamonPlay25,*/ kamonJMXReporter, /*kamonFluentd,*/ kamonKhronus,
+      kamonAutoweave/*, kamonInfluxDB*/)
```
leachbj: This just removes the projects that have missing 2.12 deps (or in the case of kamonAkkaRemote won't compile).

Haven't worked out a proper way to make this list depend on the current scala version.

project/Settings.scala
```diff
@@ -49,7 +49,7 @@ object Settings {
       "-target:jvm-1.6",
       "-language:postfixOps",
       "-language:implicitConversions",
-      "-Yinline-warnings",
+      //"-Yinline-warnings",
```
leachbj: 2.12 doesn't have this option, can we just remove it wholesale?

project/Dependencies.scala
```diff
@@ -24,17 +24,20 @@ object Dependencies {
...
   val play23Version     = "2.3.10"
   val play24Version     = "2.4.6"
   val play25Version     = "2.5.4"
   val elasticsearchVersion = "2.1.0"
 
-  val sprayJson         = "io.spray"                  %%  "spray-json"            % "1.3.1"
-  val sprayJsonLenses   = "net.virtual-void"          %%  "json-lenses"           % "0.6.0"
-  val scalatest         = "org.scalatest"             %%  "scalatest"             % "2.2.4"
+  val sprayJson         = "io.spray"                  %%  "spray-json"            % "1.3.2"
+  val sprayJsonLenses   = "net.virtual-void"          %%  "json-lenses"           % "0.6.2-SNAPSHOT"
```
leachbj: this isn't actually needed until kamonSPM or kamonNewRelic is added.

kamon-core/src/main/scala/kamon/util/executors/ExecutorMetricRecorder.scala
```diff
@@ -54,6 +44,17 @@ object ForkJoinPools {
   }
 }
 
+trait ForkJoinPoolsImplicitsLowPrio {
+  implicit object ScalaForkJoin extends ForkJoinMetrics[ForkJoinPool] {
```
leachbj: `ForkJoinPool` is deprecated and now justa a `JavaForkJoinPool` so this causes implicit ambiguity.  Moving this to a lower priority implicit should work in both cases.

Issue #421, opened 2016-12-22
make sure that Kamon can be started/stopped in the same process


ivantopo:
Currently, once you stop Kamon it gets into a weird state that doesn't allow you to start it again, and it is entirely necessary to be able to start it again if you want it to run in Play! applications in development mode. 

As a side note, I want a full solution to #395, but the scope of that issue is kind of big, the idea here is to provide a solution that works for with the current way Kamon works.

ivantopo:
This one wont fix on 0.6 and doesn't apply to 1.0, although we will need to fix this one instead: https://github.com/kamon-io/Kamon/issues/502

PR #422, opened 2016-12-24
allow restarting Kamon in the same process.

ivantopo:
taking some ideas from #395, this commit removes ConfigProviders and allows Kamon to be used in "permissive" mode until it gets started.

Related to #421.

### DIFF
--- a/kamon-core/src/main/resources/reference.conf
+++ b/kamon-core/src/main/resources/reference.conf
@@ -141,9 +141,6 @@ kamon {
     }
   }
 
-  # FQCN for a implementation of
-  config-provider = default
-
   # All settings included under the internal-config key will be used to repleace the akka.* and spray.* settings. By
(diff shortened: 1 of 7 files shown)


ivantopo:
@hunterpayne-ck it would be nice if you could take a look at how things look like after this PR, it is removing a few things you worked on, hopefully for a greater good. I would like to know how this would affect the common use cases.

dpsoft:
@ivantopo looks good to me!

hunterpayne-ck:
Seems ok to me.  I really just wanted to be able to provide a Config to Kamon.  I was only trying to follow the pattern in the existing code as I didn't completely understand the reasons for it so I didn't feel comfortable rewriting it.  As this PR preserves the start(Config) method, I don't really care that it changes the access pattern of the Kamon class/object.  If the rest of the team is OK with this PR, then so am I.

PR #442, opened 2017-02-27
Changes to make kamon-core OSGI ready

anatoliy-balakirev:
To make kamon-core OSGI ready I've added new dependency (to create OsgiActorSystem instead of ActorSystem), for that we need to have BundleContext, so I've created two new methods in the Kamon:
startWithinOsgi(context: BundleContext)
startWithinOsgi(conf: Config, context: BundleContext)

To make them working I also had to adjust Kamon class a bit.

Another required change - to generate proper OSGI manifest I've added SbtOsgi plugin + proper configuration.

This is done only for kamon-core module so far.

P.S. I've never used neither scala, nor sbt before, so code might be not the best. Fell free to adjust it if needed.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -16,16 +16,29 @@
 parallelExecution in Test in Global := false
 
 lazy val kamon = (project in file("."))
+  .enablePlugins(SbtOsgi)
   .settings(moduleName := "root")
   .settings(noPublishing: _*)
   .aggregate(core, autoweave, testkit, bench)
 
 
 lazy val core = (project in file("kamon-core"))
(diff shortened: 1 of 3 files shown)


anatoliy-balakirev:
Nice that you have this check here :) Looks like I need to fix some tests...

ivantopo:
Hey @anatoliy-balakirev, thanks for this PR! I'm sorry for not reacting on time but :disappointed:.. I'm going to close the PR since the changes are really really outdated, hope that is fine with you. See you around!

Issue #443, opened 2017-02-28
Akka 2.4.17 does not have system.shutdown() anymore


michelroolvink:
After upgrading to akka 2.4.17 the Kamon.shutdown() gives an exception:

NoSuchMethodError on _system.shutdown() which is no longer available. _system.terminate() shoudl be used to terminate an actorsystem.

dpsoft:
https://github.com/kamon-io/Kamon/pull/446

PR #446, opened 2017-03-14
[issue 443] actor system terminate

pjfanning:
https://github.com/kamon-io/Kamon/issues/443

### DIFF
--- a/kamon-core/src/main/scala/kamon/ActorSystemTools.scala
+++ b/kamon-core/src/main/scala/kamon/ActorSystemTools.scala
@@ -1,0 +1,30 @@
+/* =========================================================================================
+ * Copyright © 2013-2016 the kamon project <http://kamon.io/>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


dpsoft:
@pjfanning looks good to me! @ivantopo WDYT?

pjfanning:
https://github.com/akka/akka/blob/release-2.4/akka-actor/src/main/scala/akka/actor/ActorSystem.scala indicates that akka 2.4 still has the shutdown method but akka 2.5 removes it so it might still be useful to proceed. Another way to approach this would be to use reflection. If the build was done with akka 2.4 or 2.5 then the catch block could use reflection to call the shutdown method.

ivantopo:
we will soon be removing Akka from kamon-core (or at least, we want to)... so this won't be a long term problem. I'm fine with getting the proposed fix by @pjfanning, but there wont be need to get into reflection in the future!

ivantopo:
hey @pjfanning, thanks a lot for everything! I see you helping with PRs, answering questions to other people, thanks a lot for all your efforts! :smile:

pjfanning:
Thanks @ivantopo, @dpsoft for all the hard work you've put in to making Kamon such a valuable monitoring tool.

Issue #210, opened 2015-05-29
Make it configurable to include identification tags in the datadog module.


ivantopo:
For some users have the automatically generated `entity-category=entity-name` tag in datadog is nice, but for some others it can be pain, so, we should make it configurable and let the users decide what they want! Take a look at the comments in #188. 

Also, it might be of help to allow users to provide their own metric name and tags generator, the very same way we do in the StatsD module.

toddkazakov:
@ivantopo is there any update on this issue? We're running a kubernetes cluster and using kamon for traces and system metrics. 

Traces work really nice, because we can tag them with the pod name (implemented in 0.6.1). On the other hand, though, system metrics are pretty much unusable, because we cannot apply any tags. I'm happy to follow a similar approach with the statsd module - adding a new `KeyMetricGenerator`.

ivantopo:
I'm closing this issue because it is either too old to be relevant or stale. Stop be our [Github Discussions](https://github.com/kamon-io/Kamon/discussions) section and drop a message if you need any additional help with the topic.

Have a nice day!

PR #449, opened 2017-04-05
Add default metrics loaded from config

jasonmartens:
We are using Kubernetes for our apps, and there is currently no way to add tags to the jmx stats. This means that all of the JMX metrics have no tags for the container (aka pod in k8) they are running on, and thus you can only see aggregate java metrics for groups of containers. This is at least *a* solution to the problem mentioned in #210. 

I'm not 100% sure if this will do what I want, but thought it would be a good starting place and perhaps you would have a better idea about how to solve this problem.

### DIFF
--- a/kamon-core/src/main/resources/reference.conf
+++ b/kamon-core/src/main/resources/reference.conf
@@ -167,4 +167,15 @@ kamon {
     # Just a place holder to ensure that the key is always available. Non-core Kamon modules should provide their
     # settings in a module-info section.
   }
+
+  # Add tags to all reported metrics. Can be useful to identify the source of metrics from a particluar JVM instance.
+  # Example:
+  #
+  # default-tags {
+  #   host: ${?HOSTNAME}
+  #   container-name: ${?CONTAINER_NAME}
+  # }
+  default-tags {
+
+  }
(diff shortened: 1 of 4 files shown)


jasonmartens:
I have tested this in my dev environment, and is doing what I want. Would love some feedback.

jasonmartens:
After feedback in gitter, refactored this so the default tags were not injected in the Entity object, but in the MetricsModule. The original suggestion was to do so in the TickMetricSnapshot, but I thought this would add additional overhead since the default tags would need to be merged with custom tags every time a subscription was sent. This approach requires adding them just once when they are registered.

This also has tests, and I have tested locally with a snapshot version.

ivantopo:
this new implementation looks nicer, liking it :smile:, any comments @dpsoft?

dpsoft:
@jasonmartens @ivantopo I really like the idea and the implementation looks good to me ;)

jasonmartens:
Awesome, feel free to merge when you are ready.

jasonmartens:
@dpsoft @ivantopo Reminder that I don't have merge permissions, so need someone else to merge this. Thanks!

PR #450, opened 2017-04-06
Add cross-compile to travis build

jasonmartens:
This should prevent build failures from alternate Scala versions from being merged into master.


dpsoft:
@jasonmartens thanks for proposing this PR!! sorry for the late response :(

PR #452, opened 2017-04-18
disable akka-remote

rcavalcanti:
Akka remote gets configured if dependency is on the class path.
This can be annoying because it binds to the default port, often the same used by application actor system.

Remote can be disabled by setting provider = "local".
This change makes sure that kamon's akka system won't start akka-remote inadvertently.


dpsoft:
@rcavalcanti thanks!!!!

PR #454, opened 2017-04-21
Rename class vars so they do not shadow the def from their parent

### LINKED ISSUE #453: Tags added to Tracing types are dropped
Tags which are added to TracingContext or TracingSegment do not carry through to the corresponding "Info" case class.

ryandbair:
Fixes #453


dpsoft:
@ryandbair thanks again for proposing this PR!! ;)

PR #460, opened 2017-05-22
Call correct method for shutting down the ActorSystem

### LINKED ISSUE #459: NoSuchMethodError ActorSystem.shutdown()
Kamon.shutdown() uses the deprecated method ActorSystem.shutdown() that is no more available in Akka 2.5.

jan-pieter:
This solution uses reflection to determine if the `shutdown` method is available. Otherwise the `terminate` method is used.
Fixes #459.


kamon-core/src/main/scala/kamon/ActorSystemTools.scala
```diff
@@ -14,12 +14,12 @@
...
-import scala.util.control.NonFatal
-
 import akka.actor.ActorSystem
 
 object ActorSystemTools {
   private[kamon] def terminateActorSystem(system: ActorSystem): Unit = {
-    system.shutdown()
+    // Use system.shutdown() if it exists (akka 2.4 and lower), system.terminate() otherwise (akka 2.5+)
+    system.getClass.getMethods.find(_.getName == "shutdown").map(_.invoke(system))
+      .getOrElse(system.getClass.getMethod("terminate").invoke(system))
```
pjfanning: If you are getting rid of the scala-2.10/2.11/2.12 differences, could you move this back to src/main/scala/kamon ? I was just about to raise this as an issue myself.

dpsoft:
@jan-pieter  thanks for proposing this PR, please sign our [CLA](https://docs.google.com/forms/d/1G_IDrBTFzOMwHvhxfKRBwNtpRelSa_MZ6jecH8lpTlc/viewform) and  we are good to go!

kamon-core/src/main/scala/kamon/ActorSystemTools.scala
jan-pieter: I did ;) It however shows as if the scala-2.12 file has been moved...

jan-pieter:
@dpsoft Done!

dpsoft:
@jan-pieter done!!

jan-pieter:
Thanks! Glad I could do something in return for this great piece of software :smile:

ivantopo:
thanks @jan-pieter!

dpsoft:
@jan-pieter  we just released a new snapshot version in our bintray repository 0.6.7-f50a47658f5dfe4c2718cd88ed934d138f5fdc31 ;)

Issue #461, opened 2017-05-23
upgrading kamon-akka dependency on kamon-core to 0.6.7 breaks kamon-akka-2.3 unit tests


pjfanning:
Seems that this change is akka 2.4 specific
https://github.com/kamon-io/Kamon/commit/dc7f240540b468ab68b99011d7fa587207a98609

The provider="local" causes the kamon-akka-2.3 unit tests to fail.

One example:
                     <testcase classname="kamon.instrumentation.akka.ActorSystemMessageInstrumentationSpec" name="(It is not a test)" time="0.003">
                                                 <error message="java.lang.ClassNotFoundException: local" type="sbt.ForkMain$ForkError">sbt.ForkMain$ForkError: java.lang.ClassNotFoundException: local
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:348)
	at akka.actor.ReflectiveDynamicAccess$$anonfun$getClassFor$1.apply(DynamicAccess.scala:67)
	at akka.actor.ReflectiveDynamicAccess$$anonfun$getClassFor$1.apply(DynamicAccess.scala:66)
	at scala.util.Try$.apply(Try.scala:192)
	at akka.actor.ReflectiveDynamicAccess.getClassFor(DynamicAccess.scala:66)
	at akka.actor.ReflectiveDynamicAccess.createInstanceFor(DynamicAccess.scala:84)
	at akka.actor.ActorSystemImpl.liftedTree1$1(ActorSystem.scala:585)
	at akka.actor.ActorSystemImpl.&lt;init&gt;(ActorSystem.scala:578)
	at akka.actor.ActorSystem$.apply(ActorSystem.scala:142)
	at akka.actor.ActorSystem$.apply(ActorSystem.scala:119)
	at kamon.Kamon$Instance._start$lzycompute(Kamon.scala:61)
	at kamon.Kamon$Instance._start(Kamon.scala:58)
	at kamon.Kamon$Instance.start(Kamon.scala:69)
	at kamon.Kamon$.start(Kamon.scala:37)
[...]

PR #462, opened 2017-05-24
[#461] issue with kamon reference.conf when using akka-2.3

pjfanning:
https://github.com/kamon-io/Kamon/issues/461


dpsoft:
@pjfanning Done!!!

PR #475, opened 2017-07-05
Log warn when Entity with name=null is created / if dispatching entities fails

magro:
An application that creates a metric while passing `null` as metrics name
will not receive the expected `TickMetricSnapshot` (if it subscribes for
metrics).

As log level _warn_ (instead of error) is chosen, because from an application
perspective the app's functionality is not affected, i.e. there's no _immediate_
action required, but it's ok if this gets solved eventually.


dpsoft:
@magro thanks!!!

PR #480, opened 2017-07-17
Wip/moving ot support to a separate project

ivantopo:
This PR starts the process of moving OT support to a separate project (kamon-opentracing) and tries to keep Kamon's Tracer compatible with what we have seen happening around Census, Brave, Jaeger, Stackdriver and X-Ray.. The main changes are focused on:
  - Having a pluggable mechanism for generating and moving Trace/Span IDs around.. now instead of having just a long we have Identifier instances which contain both a String and a Array[Byte] representations. The default generator works with 8 byte identifiers but we could easily plug different implementations to interoperate with other systems.
  - Completely remove OT from kamon-core. We keep many of the concepts and ideas but removing the most concerning pain point: ref counting on ActiveSpans. Also, instead of having "Logs" we have "Annotations", more aligned with Zipkin/Census in that sense.
  - Since now we are talking with Kamon types, we have first class support for adding metric tags to Spans.

This is definitely not ready to be merged, we need a lot of tests coverage, actually create the OT wrapper and review the soundness of the proposed abstractions, looking forward to feedback!


kamon-core/src/main/scala/kamon/trace/Span.scala
```diff
@@ -174,14 +197,14 @@ object Span {
   val BooleanTagTrueValue = "1"
```
tferega: This is used only in one place a few lines up, so I'm not sure what's the point of having it here? Also the `"1"` literal is used in `SpanContextCodec`, so we should probably put it somewhere publicly accessible, and use it consistently.

kamon-core/src/main/scala/kamon/trace/Span.scala
```diff
@@ -16,155 +16,178 @@
...
+    * @param initialOperationName
+    * @param initialTags
+    * @param startTimestampMicros
+    * @param reporterRegistry
+    */
+  final class Real(spanContext: SpanContext, initialOperationName: String, initialTags: Map[String, String],
+    startTimestampMicros: Long, reporterRegistry: ReporterRegistryImpl, tracer: Tracer) extends Span {
+
+    private var collectMetrics: Boolean = true
+    private var isOpen: Boolean = true
```
dpsoft: maybe in the case of `collectMetrics` we use the `should` prefix : `shouldCollectMetrics`

cspinetta:
I really like the idea to be independent of any specification keeping the compatibility via a separated module.
About the code, it looks very good and it's more Scala idiomatic.
I'm available to help with anything, either testing a version, coding coverage tests, or whatever you need, so just tell me.

kamon-core/src/main/scala/kamon/Kamon.scala
```diff
@@ -155,7 +148,7 @@ object Kamon extends MetricLookup with ReporterRegistry with io.opentracing.Trac
     * was no active Span then the provided fallback value
     */
   def fromActiveSpan[T](code: ActiveSpan => T): Option[T] =
```
ivantopo: Sure thing! There is nothing that we can actually read from the Span but rather the SpanContext.. before it would be nice to get the TraceID/SpanID for logging, or get some baggage key but now that will be handled differently.. cool.

kamon-core/src/main/scala/kamon/trace/ActiveSpan.scala
```diff
@@ -0,0 +1,70 @@
...
+  * Thread, take it's place as the currently active Span and put the original one once this ActiveSpan gets deactivated.
+  *
+  */
+trait ActiveSpan extends Span {
+
+  /**
+    * Sets the currently active Span to whatever Span was active when this Span was activated.
+    *
+    */
+  def deactivate(): Unit
```
ivantopo: Nice catch. I think it was just like that in OT and we kept it, but it seems more useful to return the wrapped Span. :+1:

kamon-core/src/main/scala/kamon/trace/Sampler.scala
```diff
@@ -44,8 +47,10 @@ object Sampler {
     val upperBoundary = Long.MaxValue * chance
     val lowerBoundary = -upperBoundary
 
-    override def decide(spanID: Long): Boolean =
-      spanID >= lowerBoundary && spanID <= upperBoundary
+    override def decide(operationName: String, builderTags: Map[String, String]): SamplingDecision = {
```
ivantopo: To be honest, it's just this internal thing telling me that comparing/generating Longs is cheaper than doing it with Doubles :/.. no better justification than that.

kamon-core/src/main/scala/kamon/trace/Span.scala
```diff
@@ -16,155 +16,178 @@
...
-    spanContext
+  /**
+    *
+    * @param spanContext
+    * @param initialOperationName
+    * @param initialTags
+    * @param startTimestampMicros
+    * @param reporterRegistry
+    */
+  final class Real(spanContext: SpanContext, initialOperationName: String, initialTags: Map[String, String],
```
ivantopo: why would that be slightly better? Once again, the idea I have is that the JVM can optimize synchronized on this better and there is no contention expected for Spans, we just need safety :D

kamon-core/src/main/scala/kamon/trace/Span.scala
ivantopo: removed the "is", kinda looks nicer like that.

kamon-core/src/main/scala/kamon/trace/Span.scala
```diff
@@ -16,155 +16,178 @@
...
+    override def setOperationName(operationName: String): Span = synchronized {
+      if(isOpen)
+        this.operationName = operationName
+      this
+    }
 
-  private def extractMetricTag(tag: String, value: String): Unit =
-    if(tag.startsWith(Span.MetricTagPrefix))
-      additionalMetricTags = additionalMetricTags ++ Map(tag.substring(Span.MetricTagPrefix.length) -> value)
+    override def finish(): Unit =
```
ivantopo: No, the FinishedSpan is not meant to be exposed to the users but to the reporters.

kamon-core/src/main/scala/kamon/trace/Span.scala
ivantopo: This was related to tag values, which might be either numbers, booleans or strings. The new implementation takes care of it.

kamon-core/src/main/scala/kamon/trace/Span.scala
```diff
@@ -16,155 +16,178 @@
...
+    * @param reporterRegistry
+    */
+  final class Real(spanContext: SpanContext, initialOperationName: String, initialTags: Map[String, String],
+    startTimestampMicros: Long, reporterRegistry: ReporterRegistryImpl, tracer: Tracer) extends Span {
+
+    private var collectMetrics: Boolean = true
+    private var isOpen: Boolean = true
+    private val sampled: Boolean = spanContext.samplingDecision == SamplingDecision.Sample
+    private var operationName: String = initialOperationName
+    private var endTimestampMicros: Long = 0
```
ivantopo: did it.

ivantopo:
btw, this didn't really move it to another project yet, just removed it from core. Either @tferega or me will be working on kamon-opentracing proper soon.

PR #486, opened 2017-10-08
Log exceptions of reporters, instead of silent failure

markvandertol:
I've played with the new Kamon 1.0.0 RC and I tested creating my own reporter. There I found that exceptions created were silently discarded. This is how Futures work, when you don't have an onFailure afterwards or Try inside. To make it easier to debug when something goes wrong, I propose to always log unhandled exceptions from loggers.

Without this commit, there is only a log message for "failed to process a metrics tick". I've added the same kind of log message to the other invocations of Reporters.

### DIFF
--- a/kamon-core/src/main/scala/kamon/ReporterRegistry.scala
+++ b/kamon-core/src/main/scala/kamon/ReporterRegistry.scala
@@ -122,7 +122,13 @@ object ReporterRegistry {
         executionContext = ExecutionContext.fromExecutorService(executor)
       )
 
-      Future(reporterEntry.reporter.start())(reporterEntry.executionContext)
+      Future {
+        Try {
+          reporterEntry.reporter.start()
+        }.failed.foreach { error =>
+          logger.error(s"Metric reporter [$name] failed to start.", error)
+        }
+      }(reporterEntry.executionContext)
 
       if(metricReporters.isEmpty)
         reStartMetricTicker()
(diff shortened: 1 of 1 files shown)


ivantopo:
hey @markus1189, awesome! I guess it won't break any test but I'll wait for Travis to say it's all fine and merge it, thanks! :smile:

markvandertol:
@ivantopo Great that you had a look at this so fast! 

Note that I'm not markus1189.

ivantopo:
oh, sorry! btw.. any other piece of feedback regarding Kamon 1.0?

markvandertol:
It is easy to use 👍. I was able to write my own metric and trace reporter for Stackdriver fairly quick. I will share the code of that soon on Github. I haven't tried it out in production, though.

A bit more Scaladoc comments would have been nice for the Kamon class, as that is the entry point of the code and not everything is straight away obvious. I had to look in the source code to see how it works. Most methods you first have to register something and then you can use it. A short description could make this clearer.

A thing that took me more than one look to find that out was about "unrefined" and refined metrics. Metrics without tags don't start reporting straight away (only after the first value is registered), but Metrics which are refined with tags do.

ivantopo:
Thanks, Mark!

I see the issue you mentioned about refined/unrefined metrics. I guess we didn't get into this before because we never use metrics without tags... I'll take a look at this a bit later, maybe there is a way in which we could improve a bit this behavior to be more natural (besides improving docs and scaladoc, of course!)

Looking forward to that Stackdriver reporter, we are planning to release one as well so maybe we could join forces on this. I'll ping you when we start working on it to see if there is any chance of sharing work!

PR #490, opened 2017-11-17
Add span marks

ivantopo:
Marks in a Span are ways of adding additional information about things that happened during the processing of a Span (think something like events inside a Span). Marks have a simple model at the moment: Timestamp and Key. One of the most obvious use cases are actor messages, we want a Span representing the time spent since the message was sent to an actor until the message processing has finished, but there is a part of that time which was spent just waiting in a mailbox. A mark can be added at that very moment and then Span reporting backens can (if possible) report that information.

Additionally there are some minor changes to the Span API (addTag -> tag and so on) which I think are the last breaking changes we will have in the tracer API, getting ready for 1.0.0 GA :fireworks:


dpsoft:
@ivantopo LGTM!

PR #491, opened 2017-11-22
Serializable Context

mladens:
Make Context implement Serializable, use Codecs instead of default serialization, only BroadcastKeys get serialized

### DIFF
--- a/kamon-core-tests/src/test/scala/kamon/context/ContextSerializationSpec.scala
+++ b/kamon-core-tests/src/test/scala/kamon/context/ContextSerializationSpec.scala
@@ -1,0 +1,50 @@
+/* =========================================================================================
+ * Copyright © 2013-2017 the kamon project <http://kamon.io/>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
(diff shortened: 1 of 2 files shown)


dpsoft:
@mladens done!!

kamon-core/src/main/scala/kamon/context/Context.scala
```diff
@@ -15,15 +15,48 @@
...
+import kamon.Kamon
+
+class Context private (private[context] val entries: Map[Key[_], Any]) extends scala.Serializable {
   def get[T](key: Key[T]): T =
     entries.getOrElse(key, key.emptyValue).asInstanceOf[T]
 
   def withKey[T](key: Key[T], value: T): Context =
     new Context(entries.updated(key, value))
+
+  var _deserializedEntries: Map[Key[_], Any] = Map.empty
```
ivantopo: this should be private

kamon-core-tests/src/test/scala/kamon/context/ContextSerializationSpec.scala
```diff
@@ -0,0 +1,50 @@
...
+
+import java.io.{ByteArrayInputStream, ByteArrayOutputStream, ObjectInputStream, ObjectOutputStream}
+
+import kamon.Kamon
+import kamon.testkit.ContextTesting
+import org.scalatest.{Matchers, OptionValues, WordSpec}
+
+class ContextSerializationSpec extends WordSpec with Matchers with ContextTesting with OptionValues {
+  "the Context is Serializable" should {
+    "empty " in {
```
ivantopo: please add a more descriptive name for the tests :)

PR #497, opened 2017-12-13
Use instant for spans

ivantopo:
Stop using plain Long values and use java.util.Instant instead. Even though Java 8 doesn't support nanosecond precision in the default clocks, the data structure supports it and we extended the trick we had in our clock to work with relative nanos so that it works now with Instant, then changed the timestamps in the tracer to use Instant.

Using plain calls to System.nanoTime() is about 23 ns/ops and this new Instant creation is about 30 ns/ops. Need to formalize the JMH project and push it up. It's a bit more expensive but nothing really noticeable when compared to actual work done in a given Span.


dpsoft:
@ivantopo LGTM!!

PR #503, opened 2018-01-24
Allow accessing tags on Span.Local

nadavwr:
Used to be possible in 0.6.x (see #386)

### DIFF
--- a/kamon-core/src/main/scala/kamon/trace/Span.scala
+++ b/kamon-core/src/main/scala/kamon/trace/Span.scala
@@ -95,36 +95,38 @@ object Span {
     private var hasError: Boolean = false
     private var operationName: String = initialOperationName
 
-    private var spanTags: Map[String, Span.TagValue] = initialSpanTags
+    private var _spanTags: Map[String, Span.TagValue] = initialSpanTags
     private var marks: List[Mark] = Nil
     private var customMetricTags = initialMetricTags
 
     override def isEmpty(): Boolean = false
     override def isLocal(): Boolean = true
 
+    def spanTags: Map[String, Span.TagValue] = _spanTags
+
     override def tag(key: String, value: String): Span = synchronized {
       if(sampled && open)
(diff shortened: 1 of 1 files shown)


nadavwr:
Locally, `+test` seems to pass just fine. What could be failing the build?

nadavwr:
Sporadic test failures occur also with vanilla 1.0 for the same tests

ivantopo:
Hello Nadav!

There are big differences between the old TraceContext/Segments and the new Context/Span infrastructure. Before users would add information to the TraceContext and then be able to retrieve it at any point where the same TraceContext was available because there was always the very same TraceContext instance available for a single request. Now that we have Spans instead, a single request might be generating several Spans and putting data in one Span doesn't mean that you can retrieve it down the road because there might be a different current span down there.

We introduced the `Context` abstraction for these specific cases and all our instrumentation is passing the Context all over the place, regardless of tracing being available or not. The migration path that see from here is to move user data from the old TraceContext/Segment to the new Context.

Could you please share a bit more of what motivated this PR and whether the explanation above covers the use case?

ivantopo:
Hey @nadavwr, looking forward to your comments!

nadavwr:
Hi Ivan,
Our original intent had been to reflect span tags in MDC. Being able to peek into span context seemed like an absent feature.
We have since abandoned our initial focus on using span—`TraceContext` is adequate for our needs. This still seems "nice to have" though :)

ivantopo:
@nadavwr did you mean `Context` instead? the `TraceContext` is from 0.6! 

I'm going to close this PR then, but there is something I would like to understand: Do you actually need certain things to be on the MDC or you just need them to be logged? Also, there is a conversation [on the kamon-logback module](https://github.com/kamon-io/kamon-logback/pull/4) that might be interesting to you if on the need to put stuff on the MDC.

nadavwr:
We skip the MDC and have them logged from context

ivantopo:
thanks for the feedback :)

PR #504, opened 2018-01-24
Add warning when no reporters are loaded

briantopping:
While adding a warning about nothing to collect, I think I noticed a bug with the comparison about `traceTickInterval`.

The warning is good when people don't even know if Kamon is loading. They'll at least see something in their logs this way.


briantopping:
Caution: I would close this PR for potential lack of understanding, but wondering if the change on line 203 is correct.

ivantopo:
Hey @briantopping, thanks for the PR! the change in line 203 definitely has to come in, but the warning is not necessary. Reconfiguring Kamon doesn't necessarily mean that reporters should be present or not and there is already a warning log when loading reporters from configuration and nothing is in the `kamon.reporters` config key.

Could you please re-purpose the PR (commit message, PR title and so on) towards only fixing the issue on line 203? Thanks!

ivantopo:
thanks @briantopping :+1:

PR #508, opened 2018-01-30
turn all Kamon threads into daemon threads, except for reporters

### LINKED ISSUE #502: No way to fully shut down Kamon 1.0
There should be a way to completely shut down Kamon programatically, making the JVM terminate normally without recurring to explicit `System.exit` calls. With Kamon 1.0, unfortunately, I can't seem to do this:

```scala
object Main extends App {
  Kamon.loadReportersFromConfig() // this may even be an empty list of reporters
  Kamon.stopAllReporters()
  // the JVM should terminate after this
}
```

It seems that due to the `registryExecutionContext` inside `ReporterRegistry`, the JVM keeps active threads that prevent the JVM shutdown. We should be able to shut down the `registryExecutionContext` either by a `Kamon.stop()` method like before or by `stopAllReporters()`.

ivantopo:
Turn all Threads spawned by Kamon into daemon threads, except for those assigned to reporters. The only thing that will keep a JVM running from Kamon's side is any reported that might still be running. 

Fixes #502

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -50,7 +50,10 @@ lazy val testkit = (project in file("kamon-testkit"))
 
 
 lazy val coreTests = (project in file("kamon-core-tests"))
-  .settings(moduleName := "kamon-core-tests", resolvers += Resolver.mavenLocal)
+  .settings(
+    moduleName := "kamon-core-tests",
+    resolvers += Resolver.mavenLocal,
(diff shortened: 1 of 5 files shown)


dpsoft:
@ivantopo LGTM!!!

dpsoft:
@ivantopo Done!!

PR #509, opened 2018-02-02
allow arbitrary header names for automatic broadcast string keys, fixes #505

### LINKED ISSUE #505: Allow users to specify arbitrary header names for string keys
This would make it dead simple to pass arbitrary information around, specially in cases where these headers are defined prior to introducing Kamon in the infrastructure.

ivantopo:
This PR is changing the way `kamon.context.codecs.string-keys` is read from a regular string list to a object. The object allows to have a key name and the correspondent HTTP header name for it. This should make it extremely simple to integrate with things like LinkerD or Envoy, plus the typical cases of correlation fields and external request ids.

Fixes #505

### DIFF
--- a/kamon-core-tests/src/test/resources/reference.conf
+++ b/kamon-core-tests/src/test/resources/reference.conf
@@ -1,0 +1,5 @@
+kamon {
+  context.codecs.string-keys {
+    request-id = "X-Request-ID"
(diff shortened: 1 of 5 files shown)


dpsoft:
@ivantopo LGTM!!

PR #520, opened 2018-03-15
add a identity provider that supports UUIDs

### LINKED ISSUE #498: add a identity provider that supports UUIDs
Conversation [on Gitter](https://gitter.im/kamon-io/Kamon?at=5a4dfb8dba39a53f1ad524e8)

I was trying to hack a quick SpanCodec that could use [Heroku's request ids](https://devcenter.heroku.com/articles/http-request-id) as the TraceID but then would also need a IdentityProvider that can unpack those UUIDs..

jtjeferreira:
fixes #498

### DIFF
--- a/kamon-core/src/main/scala/kamon/ContextStorage.scala
+++ b/kamon-core/src/main/scala/kamon/ContextStorage.scala
@@ -140,5 +140,10 @@ object ContextStorage {
     * store and remove Context instances for small periods of time as events flow through the system and the
     * instrumentation follows them around.
     */
-  private val _contextStorage = Storage.ThreadLocal()
+  private val _contextStorage: Storage = {
+    if("true".equals(System.getProperty("kamon.context.debug")))
+      Storage.Debug()
+    else
+      Storage.ThreadLocal()
+  }
 }
--- a/kamon-core/src/main/scala/kamon/context/Storage.scala
+++ b/kamon-core/src/main/scala/kamon/context/Storage.scala
@@ -17,6 +17,9 @@
 package kamon
 package context
 
+import java.util.concurrent.ConcurrentLinkedQueue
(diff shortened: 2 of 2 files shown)


jtjeferreira:
Hi everybody.

The current implementation creates Identifiers where the string part is just the hex view of the bytes, but accepts the canonical UUID representation separated by `-` (like `abc6c98d-2404-41cd-968c-e21b5e3cd9d1` vs `abc6c98d240441cd968ce21b5e3cd9d1`

I would like that the string representation would be the version with separated by `-`, but `kamon-zipkin` does not like that because it is using the string representation when creating spans, and the string is greater than 32 chars.

What are yours thoughts about this? Could you elaborate why the identifiers have a string and a byte array representation?

ivantopo:
The reason for having both a string and bytes representation of identifiers is that these are the 2 kinds of encoding we support for Kamon's Context, which in turn has to carry around Spans with the identifiers. At the time, it seemed like some reporters were going to need the binary version of the identifiers but the traceID is something that is included in logs all the time so also not having to transform the identifier to a string all the time was a plus, but now I see how this is a fragile and more SpanReporter-dependent approach than expected :(

> I would like that the string representation would be the version with separated by -

How important is the above? I think that if you keep it without the dashes then Zipkin will accept it without problems.

jtjeferreira:
> How important is the above? I think that if you keep it without the dashes then Zipkin will accept it without problems.

I would like to have logs with the exact version that comes in from the header (e.g `abc6c98d-2404-41cd-968c-e21b5e3cd9d1`). If we changed `kamon-zipkin` to use the binary version, the `UUIDTraceID` could carry the version with the `-` in the string part and the `byte[]` in the binary part and everything would work (at least for my use case :laughing: )

PR #521, opened 2018-03-18
* Introduce some scalac params in order to improve the emited bytecode

dpsoft:
`l:method`: Enable intra-method optimizations: unreachable-code,simplify-jumps,compact-locals,copy-propagation,redundant-casts,box-unbox,nullness-tracking,closure-invocations.


ivantopo:
Awesome! We have to JMH the shit out of core ASAP, I would bet that this improves performance but we need to have the sweet sweet numbers to be sure :)

dispalt:
Just fyi, https://github.com/scala/bug/issues/10717

PR #532, opened 2018-05-08
Allow using negative values in RangeSampler#increment and RangeSampler#decrement

jcazevedo:
This pull request modifies the `increment` and `decrement` methods of `RangeSampler` so that they correctly track the minimum and maximum values if they're supplied with negative values.

### DIFF
--- a/kamon-core-tests/src/test/scala/kamon/metric/RangeSamplerSpec.scala
+++ b/kamon-core-tests/src/test/scala/kamon/metric/RangeSamplerSpec.scala
@@ -83,6 +83,34 @@ class RangeSamplerSpec extends WordSpec with Matchers {
       snapshot.distribution.min should be(0)
       snapshot.distribution.max should be(0)
     }
+
+    "report correct min and max values if increment or decrement are used with negative values" in {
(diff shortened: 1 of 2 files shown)


ivantopo:
Hey @jcazevedo! Thanks for the PR :)... Just wanted to ask, what was the use case where allowing negative numbers was necessary?

jcazevedo:
It's mostly an API convenience. I have a situation where I'm keeping track of deltas with a min max counter and I'm currently doing:

```scala
if (delta >= 0) metric.increment(delta)
else metric.decrement(-1 * delta)
```

If the `increment` and `decrement` methods would support negative values, I could simply do `metric.increment(delta)`.

ivantopo:
Seems like a nice thing to have, thanks for sharing! Although, I think it would be better to add a new method to the RangeSampler API, lets say `def delta(value: Long)` (or another name that makes sense) that does the right thing for this case.

If we go that way we keep the increment/decrement with the same cost as before (only updating two places) and only incur in the additional expense when it is actually needed. What do you think about that? Also, what does @dpsoft think about this change?

PR #535, opened 2018-06-01
Use spanIdGenerator for decoding spans in SpanCodec.Colfer

lustefaniak:
Right now `traceIdGenerator` is used for all decoding, which causes problems when `traceId` and `spanId` are in different format.
This works fine in `SpanCodec.B3` so is probably a mistake.


ivantopo:
Nice catch, @lustefaniak, hope it wasn't that much of a nightmare to figure it out!

PR #536, opened 2018-06-06
Make span metrics names configurable

fazy:
When using multiple languages and libraries it can be helpful to ensure all are outputting consistent metrics. This PR helps to address that by allowing the metric spans to be renamed.

Example:

```
kamon {
  trace {
    span-metrics {
      processing-time-name = "request-latency"
      error-count-name = "request-error"
    }
  }
```

This affects the results, e.g. with Prometheus:

```
$curl http://localhost:90
request_latency_seconds_bucket{...}
...
```

Alternative ways to achieve this: I tried to look at whether this can be done in kamon-prometheus but I think by then it is too late to change the name.

It seems this problem is fairly specific to the names of spans - we can add other metrics, and we can manipulate tags.

In kamon-core, there is the `Metrics` singleton object. This could be made into a class and the config injected in, but I'm not sure if that complexity is needed - in particular, is there any performance implication in making it no longer a singleton.

### DIFF
--- a/kamon-core/src/main/resources/reference.conf
+++ b/kamon-core/src/main/resources/reference.conf
@@ -147,6 +147,14 @@ kamon {
       # When this option is enabled the metrics collected for Spans will automatically add a tag named "parentOperation"
       # with the name of the operation on the parent Span, if any.
       scope-spans-to-parent = yes
+
+      # Name to use for metric spans. This should not normally be changed, except when needed to maintain consistency
+      # with other languages/libraries.
+      processing-time-name = "span.processing-time"
+
+      # Name to use for error count spans. This should not normally be changed, except when needed to maintain
+      # consistency with other languages/libraries.
+      error-count-name = "span.error-count"
     }
   }
 
--- a/kamon-core/src/main/scala/kamon/trace/Span.scala
+++ b/kamon-core/src/main/scala/kamon/trace/Span.scala
@@ -18,6 +18,7 @@ package trace
 
 import java.time.Instant
 
+import com.typesafe.config.ConfigFactory
 import kamon.ReporterRegistry.SpanSink
 import kamon.context.Key
 import kamon.metric.MeasurementUnit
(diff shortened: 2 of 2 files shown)


ivantopo:
Hey @fazy! 

> Alternative ways to achieve this: I tried to look at whether this can be done in kamon-prometheus but I think by then it is too late to change the name.

I actually think that moving this to kamon-prometheus is the way to go. So far this would only be important for people using Prometheus because it is the only platform with some sort of well-known convention regarding metric names for HTTP request and such convention differs from what we do in Kamon. Additionally to that, not all Spans would need this kind of transformation, only Spans generated on the server side of HTTP/RPC instrumentation, all others like HTTP client calls, JDBC calls, Akka messages and so on shouldn't get this transformation since it would be breaking the conventions on a different way (having metrics called request-latency that actually don't mean the same).

Given that at the moment this would only be applied to the `span.processing-time` metric when the tag `span.kind=server` is set, I would let's just add a setting in the kamon-prometheus module and apply the transformation right before passing the [snapshot to the snapshot accumulator](https://github.com/kamon-io/kamon-prometheus/blob/master/src/main/scala/kamon/prometheus/PrometheusReporter.scala#L60).

If this becomes a recurring situation and we see the need to apply more rules to other metrics we could do something similar to the wrapper in https://github.com/kamon-io/kamon-datadog/pull/22.

What do you think about the above?

fazy:
@ivantopo @dpsoft do you have any comments on https://github.com/kamon-io/kamon-prometheus/pull/19? - if it's likely to be merged I'll be happy to close this one. :)

ivantopo:
Hey @fazy! Sorry for leaving this hanging for so long :sob: ... that PR was just merged and released as [v1.1.2](https://github.com/kamon-io/kamon-prometheus/releases/tag/v1.1.2), thanks a lot for that! It is actually a super awesome addition to the module :tada: 

I'll proceed to close this PR as well since there is now a solution on the Prometheus reporter. But also, just thinking about it, we should have an easy way to do that in core. I just opened https://github.com/kamon-io/Kamon/issues/589 to keep track of that. See you around!

PR #538, opened 2018-06-11
Ensure that histograms snapshots are immutable

ivantopo:
Previously if any reporter was able to access the `zigZagCounts` ByteBuffer then it would be possible to change it's state (most notably, position) and could break operations for other reporters. This ensures no single reporter can mess with the counts.


dpsoft:
@ivantopo LGTM!

PR #539, opened 2018-06-13
Allow access to local span operation name

Falmarri:


ivantopo:
This looks good to me. Will just wait a bit for @dpsoft to take a look at it and make sure all is fine from his side.

PR #540, opened 2018-06-14
= core: exclude com.lihaoyi.fansi dependency

dpsoft:


build.sbt
```diff
@@ -44,13 +49,9 @@ lazy val core = (project in file("kamon-core"))
...
-    libraryDependencies ++= Seq(
-      "com.typesafe"     %  "config"          % "1.3.1",
-      "org.slf4j"        %  "slf4j-api"       % "1.7.25",
-      "org.hdrhistogram" %  "HdrHistogram"    % "2.1.9",
-      "com.lihaoyi"      %% "fansi"           % "0.2.4"
-    )
-  )
+    libraryDependencies ++=
+      compileScope(typesafeConfig, sl4jApi, hdr) ++
+      testScope(fansi))
```
ivantopo: Don't even leave it as test, we are not using it at all.

PR #553, opened 2018-09-12
Improvements in ThreadLocalStorage

dpsoft:
* Wrapper that implements optimized `ThreadLocal` access pattern ideal for heavily used ThreadLocals.

* It is faster to use a mutable holder object and always perform `ThreadLocal.get()` and never use  `ThreadLocal.set()`, because the value is more likely to be found in the `ThreadLocalMap` direct hash
slot and avoid the slow path `ThreadLocalMap.getEntryAfterMiss()`.

* With a twist proposed by @adriancole where we don't use an `Kamon-defined` holder object as that would prevent class unloading.

JMH Results:

`taskset -c 0-2  sbt 'coreBench/jmh:run -i 10 -wi 5 -t2 .*Benchmark.*'`
```
[info] Benchmark                                       Mode  Cnt   Score   Error  Units
[info] ThreadLocalStorageBenchmark.currentThreadLocal  avgt  100  17.579 ± 0.261  ns/op
[info] ThreadLocalStorageBenchmark.fastThreadLocal     avgt  100  13.769 ± 0.275  ns/op
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -80,7 +80,6 @@ lazy val coreBench = (project in file("kamon-core-bench"))
   .enablePlugins(JmhPlugin)
   .settings(
     moduleName := "kamon-core-bench",
-    resolvers += Resolver.mavenLocal,
     fork in Test := true)
   .settings(noPublishing: _*)
   .settings(commonSettings: _*)
--- a/build.sbt
+++ b/build.sbt
@@ -17,7 +17,7 @@
 lazy val kamon = (project in file("."))
   .settings(moduleName := "kamon")
   .settings(noPublishing: _*)
-  .aggregate(core, testkit, coreTests)
+  .aggregate(core, testkit, coreTests, coreBench)
 
 val commonSettings = Seq(
   scalaVersion := "2.12.4",
@@ -74,3 +74,14 @@ lazy val coreTests = (project in file("kamon-core-tests"))
       "ch.qos.logback" % "logback-classic" % "1.2.2" % "test"
     )
   ).dependsOn(testkit)
+
+
+lazy val coreBench = (project in file("kamon-core-bench"))
+  .enablePlugins(JmhPlugin)
+  .settings(
+    moduleName := "kamon-core-bench",
+    resolvers += Resolver.mavenLocal,
+    fork in Test := true)
+  .settings(noPublishing: _*)
+  .settings(commonSettings: _*)
+  .dependsOn(core)
--- a/kamon-core-bench/src/main/scala/kamon/bench/ThreadLocalStorageBenchmark.scala
+++ b/kamon-core-bench/src/main/scala/kamon/bench/ThreadLocalStorageBenchmark.scala
@@ -22,7 +22,7 @@ import kamon.context.Storage.Scope
 import kamon.context.{Context, Key, Storage}
 import org.openjdk.jmh.annotations._
 
-@State(Scope.Benchmark	)
+@State(Scope.Benchmark)
 class ThreadLocalStorageBenchmark {
 
   val TestKey: Key[Int] = Key.local("test-key", 0)
--- a/kamon-core-bench/src/main/scala/kamon/bench/ThreadLocalStorageBenchmark.scala
+++ b/kamon-core-bench/src/main/scala/kamon/bench/ThreadLocalStorageBenchmark.scala
@@ -26,7 +26,7 @@ import org.openjdk.jmh.annotations._
 class ThreadLocalStorageBenchmark {
 
   val TestKey: Key[Int] = Key.local("test-key", 0)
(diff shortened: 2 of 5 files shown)


dpsoft:
All credits to [FastThreadLocal] by @trask 

[FastThreadLocal]: https://github.com/glowroot/glowroot/blob/master/agent/plugin-api/src/main/java/org/glowroot/agent/plugin/api/util/FastThreadLocal.java#L18-L28

trask:
👍

kamon-core/src/main/scala/kamon/context/Storage.scala
```diff
@@ -27,28 +29,45 @@ object Storage {
...
+    * ThreadLocal.set(), because the value is more likely to be found in the ThreadLocalMap direct hash
+    * slot and avoid the slow path ThreadLocalMap.getEntryAfterMiss().
+    *
+    * Important: this thread local will live in ThreadLocalMap forever, so use with care.
+    */
   class ThreadLocal extends Storage {
-    private val tls = new java.lang.ThreadLocal[Context]() {
-      override def initialValue(): Context = Context.Empty
+    private val tls = new java.lang.ThreadLocal[ContextHolder]() {
+      override def initialValue() = new ContextHolder(Context.Empty)
```
adriancole: I like this but won't it risk a classloader leak? ContextHolder will prevent the classloader from being unloadable unless it is explicitly cleaned. maybe this doesn't mean anything if it is loaded via an agent or the classloader is never expected to be unloaded...

kamon-core/src/main/scala/kamon/context/Storage.scala
```diff
@@ -27,28 +29,45 @@ object Storage {
...
 
     override def current(): Context =
-      tls.get()
+      get()
 
     override def store(context: Context): Scope = {
       val newContext = context
-      val previousContext = tls.get()
-      tls.set(newContext)
+      val previousContext = get()
```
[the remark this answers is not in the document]
adriancole: ex using a holder object like Object[] would allow the same but not prevent classloader unloading (of kamon types)

kamon-core/src/main/scala/kamon/context/Storage.scala
dpsoft: @adriancole thanks for pointing that, yes you are right, here we can have a classloader leak if the context is not correctly closed, what do you think about something like this:     

``` scala
final class ContextHolder(var value:Context) extends WeakReference(value)

```

adriancole:
the classloader thing happens even if you clear every time.. it is a weird
one! we have a kindof horrible framework to show this problem

Ex this test will fail even if you clear the ref.. the test below will fail
until it is a weak ref, but weak ref isn't cheap either..
https://github.com/openzipkin/brave/blob/master/brave-tests/src/main/java/brave/test/propagation/CurrentTraceContextTest.java#L265
https://github.com/openzipkin/brave/blob/master/brave-tests/src/main/java/brave/test/util/ClassLoaders.java

adriancole:
cheap being relative to a plain ref or array access I mean.. not saying
weakref isn't cheap enough

kamon-core/src/main/scala/kamon/context/Storage.scala
dpsoft: @adriancole I've run the benchmarks with yours proposed changes:

image

looks really good!!

adriancole: 👍

kamon-core-bench/src/main/scala/kamon/bench/ThreadLocalStorageBenchmark.scala
```diff
@@ -0,0 +1,81 @@
...
+
+import kamon.context.Storage.Scope
+import kamon.context.{Context, Key, Storage}
+import org.openjdk.jmh.annotations._
+
+@State(Scope.Benchmark)
+class ThreadLocalStorageBenchmark {
+
+  val TestKey: Key[Int] = Key.local("test-key", 0)
+  val ScopeWithKey: Context = Context.create().withKey(TestKey, 43)
```
ivantopo: This is actually a ContextWithKey, not a scope ;) #nitpicking

build.sbt
```diff
@@ -74,3 +74,14 @@ lazy val coreTests = (project in file("kamon-core-tests"))
       "ch.qos.logback" % "logback-classic" % "1.2.2" % "test"
     )
   ).dependsOn(testkit)
+
+
+lazy val coreBench = (project in file("kamon-core-bench"))
+  .enablePlugins(JmhPlugin)
+  .settings(
+    moduleName := "kamon-core-bench",
+    resolvers += Resolver.mavenLocal,
```
ivantopo: why is this resolver necessary?

PR #554, opened 2018-09-16
During reconfigure, tickers are restarted using old config

fchiron:
I noticed that when Kamon is reconfigured with a new tick-interval, the new interval is ignored (the old one is used).
It turns out that `reStartMetricTicker` and `reStartTraceTicker` read the configuration from `registryConfiguration`, but the `registryConfiguration` has not been updated yet when these methods are called.

I just moved the update of `registryConfiguration` to happen before the tickers are restarted, but an alternative would be to pass a `Configuration` parameter to `reStartMetricTicker` and `reStartTraceTicker`.
What do you think ?

### DIFF
--- a/kamon-core/src/main/scala/kamon/ReporterRegistry.scala
+++ b/kamon-core/src/main/scala/kamon/ReporterRegistry.scala
@@ -195,12 +195,13 @@ object ReporterRegistry {
     }
 
     private[kamon] def reconfigure(config: Config): Unit = synchronized {
-      val newConfig = readRegistryConfiguration(config)
+      val oldConfig = registryConfiguration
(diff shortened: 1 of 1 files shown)


ivantopo:
Hey @fchiron, thanks a lot for proposing this PR and my apologies for taking so long to come and look at it. The PR is fine and I'll merge it right away, but the code that it is touching is about to be removed by #559.

It seems like given the way reconfigure [is handled in that PR](https://github.com/kamon-io/Kamon/pull/559/files#diff-739444bf6d4092f2e9c6963976c467aaR139) it shouldn't be a problem anymore. If you could take a look there just to make sure we are not making the same mistake again it would be awesome. Thanks!

PR #559, opened 2018-10-16
Discoverable modules and the new module registry

ivantopo:
This PR brings back something we had back in Kamon 0.6.x: the ability to discover modules by simply having them in the classpath. From this point on, calling `Kamon.loadModules()` will pick everything under the `kamon.modules` configuration and try to start them.

We didn't have the actual "modules" concept in 1.x, only reporters were available but with this PR the MetricReporter and SpanReporter traits are just a special type of module. Soon we should be moving things like the executor metrics collectors and JVM/host metrics to modules so that we can globally control their lifecycle and allow a "drop a jar" installation method :).

Also deprecated the "addReporter" variants but kept them compatible.


dpsoft:
@ivantopo awesome man!!! 
We have only a compilation error:
image

dpsoft:
@ivantopo LGTM!

PR #560, opened 2018-10-30
include trace identifiers in HTTP responses, fixes #558

### LINKED ISSUE #558: Copy certain context keys into the HTTP responses
There is a common request for adding the trace and span ids on the HTTP responses from an instrumented service (see https://github.com/kamon-io/kamon-play/issues/16 for example). This is a common enough request that we should add it, but we need to figure out what's the cleanest way to do it.

Since #552 we will have an abstraction for instrumenting HTTP servers so, my proposal would be to add some special configuration there so that we can say "write the key `span` into the response, using the XXXX entry writer". This needs some more thought, looking forward for comments.

ivantopo:
Allows users to specify whether trace identifiers (trace and span) should be written to instrumented HTTP responses. By default we are including the trace ID using the `trace-id` header. The header names are configurable.

### DIFF
--- a/kamon-core-tests/src/test/resources/reference.conf
+++ b/kamon-core-tests/src/test/resources/reference.conf
@@ -21,8 +21,14 @@ kamon {
   instrumentation {
     http-server {
       default {
-        tracing.preferred-trace-id-tag = "correlation-id"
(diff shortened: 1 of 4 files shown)


dpsoft:
@ivantopo done!

PR #561, opened 2018-11-14
add a simple metric group utility for instrumentations

ivantopo:
This MetricGroup abstraction helps to achieve what we had back in 0.6 with the EntityRecorder abstraction. It is basically an utility to tag several metrics with the same common tags and remember them, so when the group is cleaned up all related metrics will be cleaned up as well.


dpsoft:
@ivantopo LGTM.

PR #567, opened 2019-01-25
Updating NoIdentifier to match bytes value

sjoerdmulder:
To match the string with the bytes value for the NoIdentifier it should have value `0` also this nicely fixes the issue reported here: https://github.com/kamon-io/kamon-zipkin/issues/9

### DIFF
--- a/kamon-core/src/main/scala/kamon/trace/IdentityProvider.scala
+++ b/kamon-core/src/main/scala/kamon/trace/IdentityProvider.scala
@@ -37,7 +37,7 @@ object IdentityProvider {
     }
   }
 
-  val NoIdentifier = Identifier("", new Array[Byte](0))
+  val NoIdentifier = Identifier("0", new Array[Byte](0))
 
   trait Generator {
     def generate(): Identifier


kamon-core/src/main/scala/kamon/trace/IdentityProvider.scala
```diff
@@ -37,7 +37,7 @@ object IdentityProvider {
     }
   }
 
-  val NoIdentifier = Identifier("", new Array[Byte](0))
+  val NoIdentifier = Identifier("0", new Array[Byte](0))
```
ivantopo: Hey @sjoerdmulder! Thanks for bringing this up :)

I'm not sure if this is the right solution to the problem, if you look a bit upper, this is the implementation of equals for an identifier:

```scala
    override def equals(obj: Any): Boolean = {
      if(obj != null && obj.isInstanceOf[Identifier])
        obj.asInstanceOf[Identifier].string == string
      else false
    }
```

If only the string representation is being compared then it might happen that comparing a "0" identifier with NoIdentifier will yield equality, which I recognize is a very very rare case but still, there is "some" identifier. Also this breaks the assumption of the String identifier always being a representation of the binary identifier and vice-versa, which is relied upon in other parts of the instrumentation. 

I think that instead of solving it here in core it would be better to do this check on the zipkin reporter and only add the parentId if it is not empty. What do you think?

sjoerdmulder: @ivantopo I kind of agree, ideally either the string or byte representation is removed and always derived from it's counterpart e.g. bytes value is kept and string representation is derived from those bytes.

B.t.w. it seems that the Zipkin Span has functionality to strip and pad zeros of the string identifier making `0000000000` equaling `0` therefor equal `null` in the parentId. 

Will update the PR on zipkin for now so we can move forward :smile:

sjoerdmulder:
@ivantopo Closed this PR in favor of other way to fix for kamon-io/kamon-zipkin#9

PR #569, opened 2019-02-08
Status page

ivantopo:
This PR brings a long-awaited tool for debugging application setup on Kamon :tada:. Long story short is that when Kamon starts it runs an embedded HTTP server with a simple status page that shows you the current internal status of Kamon with enough information to troubleshoot the most common options:
  - Is instrumentation enabled?
  - Do I have to many metrics?
  - Are modules started?
  - Are instrumentation modules working?

Here is an example of how it looks like in a dumb test application:

image

I'll be publishing snapshots of more modules with Kanela support and see if this breaks on any real world applications but so far seems working fine!


jtjeferreira:
any chance to make this a separate module from core? I don't mind if it is a core dependency, but that would make it easier to exclude if I dont want to use it...

ivantopo:
Hey @jtjeferreira may I ask what brings up the question? We are shading the couple extra (and super lightweight) dependencies inside the kamon-core jar and the resulting jar only depends on Typesafe Config and SLF4J. The main reason to have this as a built-in component in core is that it will greatly simplify supporting Kamon (as in, helping people set it up correctly) and it opens the door for a few things we would like to have in the future like changing the tracer sampler configuration in runtime or even showing some traces/metrics right there on the status page!

Of course, if you just don't want the status page exposed it can be disabled :smile:

jtjeferreira:
I understand the need to "helping people set it up correctly", but you don't give the option for "power" users to completely disable this new status page and not include the unnecessary classes, because they are included in the "core" jar. 

My suggestion of making it a core dependency, was that it would be included by default (so new users have it OOTB), but I could exclude it easily with `sbt`, however that would introduce a cyclic dependency, because this new module would need to use core classes...

dpsoft:
@ivantopo Awesome man!!. Looks really good to me!

ivantopo:
Hey @jtjeferreira, your last comment made me reflect on this issue a bit further and realize that the status page definitely doesn't need to be part of the kamon-core jar, so I split the original PR into two different things: the core Status API which is just a couple classes that allow users to get status from internal components for debugging purposes and the `kamon-status-page` project which is a separate dependency that has the embedded HTTP server and JsonMarshalling required to make it work. It will not even be a dependency on kamon-core, just a separate project within the same repository.

The only reason I had to include this in kamon-core was that it is easier for new users if this tool is there and starts automatically, but it really didn't feel like it should be there. Now it is not :smile:. There is something that has been on the table for long: the kamon-bundle, which is just a big jar that contains a bunch of instrumentation and the Kanela agent, which will make it a lot easier to setup and maintain Kamon. Instead of including the status page on core, we will included in the kamon-bundle. Then users will have the choice: they can add each dependency separately as they have done since the beginning of time and have fine control over what gets into the classpath or, they can use the bundle and let Kamon do the job. Everybody happy (hopefully :joy:)

PR #573, opened 2019-03-12
Add support for providing a custom Sampler

notNotDaniel:
This PR makes it possible to set `kamon.trace.sampler = custom`, after which `kamon.trace.custom-sampler` may be used to specify the name of a custom `Sampler` to be used. This is useful in our development environment where we generally don't want to sample traces, but we occasionally want to white-list specific top-level operations to be sampled for debugging.

### DIFF
--- a/kamon-core-tests/src/test/scala/kamon/trace/TracerSpec.scala
+++ b/kamon-core-tests/src/test/scala/kamon/trace/TracerSpec.scala
@@ -25,6 +25,22 @@ import kamon.trace.Span.TagValue
 import kamon.trace.SpanContext.SamplingDecision
 import org.scalatest.{Matchers, OptionValues, WordSpec}
 
+// A custom sampler for testing, samples all operations that start with a specific prefix
+object TestSampler {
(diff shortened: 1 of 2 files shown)


ivantopo:
Hey @notNotDaniel (Daniel? :smile:), nice PR! 

Is this something that you need as part of Kamon 1.1 or can hold it to Kamon 2.0?

notNotDaniel:
Thanks Ivan, we're just starting to play with Kamon (we have it fully integrated, but the overhead has been problematic, mostly due to aspectj I suspect), so we can wait for 2.0, though if it makes it into 1.1 we will use it there.

ivantopo:
Hey @notNotDaniel, this got implemented on Kamon 2.0 a few months ago so I'm closing it. For reference, here is the setting for that: https://github.com/kamon-io/Kamon/blob/master/kamon-core/src/main/resources/reference.conf#L169-L177

notNotDaniel:
Excellent, thanks for letting me know, we're looking forward to making use
of it.

Issue #533, opened 2018-05-08
Give reporters a final chance to flush data before stopping


ivantopo:
In the particular case of all reporters being shut down because of the JVM exiting we should have a means of reporting whatever data has been accumulated since the last tick before completely stopping the reporters. I think the flow should be:
  1. `Kamon.stopAllReporters()` is called
  2. Whatever state is available for metrics and span reporters is flushed via `reportPeriodSnapshots`/`reportSpans`
  3. The `stop` hook is called
  4. After completing all the above successfully or a timeout is reached Kamon finishes shutting down the reporters.

PR #588, opened 2019-05-27
remove the start method from the Module interface

ivantopo:
While creating and updating modules I noticed one small annoyance that is sort of forced by the `Module` interface: since users are supposed to implement a `start` method but obviously there are always a number of "things" that need to be instantiated for the module to work then we end up creating modules with a bunch of `vars` that are initialized with some sort of default value or defined as `Option[X]` and then create the actual values when `start` is called, but, in practice, there is no need to do that. All modules are instantiated right when they are going to be started, there is no chance to have an instantiated but not started module so I'm proposing to remove the `start` method from the `Module` interface.

Besides that change, I'm adding a `ModuleFactory` to be used when the modules are automatically instantiated by Kamon instead of provided by the user (which will be the default behavior).. this way, we force the parameter-less constructor only on the Factory and the Module itself is free to receive/grab any dependencies it might need on its constructor.

### DIFF
--- a/kamon-core/src/main/scala/kamon/module/ModuleRegistry.scala
+++ b/kamon-core/src/main/scala/kamon/module/ModuleRegistry.scala
@@ -5,7 +5,7 @@ import java.time.{Duration, Instant}
 import java.util.concurrent.{CountDownLatch, Executors, ScheduledFuture, TimeUnit}
 import java.util.concurrent.atomic.AtomicReference
 
-import com.typesafe.config.Config
+import com.typesafe.config.{Config, ConfigFactory}
 import kamon.module.Module.Registration
 import kamon.status.Status
 import kamon.metric.MetricRegistry
@@ -13,7 +13,6 @@ import kamon.trace.Tracer
 import kamon.util.Clock
 import org.slf4j.LoggerFactory
 
-import scala.collection.JavaConverters.asScalaBufferConverter
 import scala.concurrent.{ExecutionContext, ExecutionContextExecutorService, Future, Promise}
 import scala.util.Try
 import scala.util.control.NonFatal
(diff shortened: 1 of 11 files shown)


kamon-core/src/main/scala/kamon/module/ModuleRegistry.scala
```diff
@@ -269,63 +266,39 @@ class ModuleRegistry(classLoading: ClassLoading, configuration: Configuration, c
...
+    })
     moduleEntry.toOption
   }
 
-  private def createEntry(settings: Module.Settings, programmaticallyAdded: Boolean, instance: Module): Entry[Module] = {
+  private def createExecutor(settings: Module.Settings): ExecutionContextExecutorService = {
     val executor = Executors.newSingleThreadExecutor(threadFactory(settings.name))
-    Entry(settings.name, ExecutionContext.fromExecutorService(executor), programmaticallyAdded, settings, instance)
+
+    // Scheduling any task on the executor ensures that the underlying Thread is created and that the JVM will stay
```
Falmarri: Why is this desired? Couldn't this cause the JVM to hang on exit if there is an issue? I think that could be extremely confusing fo rusers

ivantopo: The idea is to force people to shutdown the modules so that they have a final chance to report any accumulated data, we even have [some tests](https://github.com/kamon-io/Kamon/blob/master/kamon-core-tests/src/test/scala/kamon/KamonLifecycleSpec.scala#L16-L32) covering this behavior and https://github.com/kamon-io/Kamon/issues/533 that are sort of related to this.

If an user sees that while testing their JVM doesn't shutdown then their are much more likely to take this into account and properly call Kamon.stopModules() before shutting down the app.

dpsoft: I like the idea of the users have a final chance to report theirs accumutated data but I not sure if return a `ExecutionContextExecutorService` is a good idea thinking in the java users, maybe a simple `java.util.concurrent.ExecutorService` is enough.

dpsoft:
@ivantopo LGTM!

kamon-core/src/main/scala/kamon/module/ModuleRegistry.scala
shobull: This should be probably documented somewhere. We've just came into this issue of hanging application because of Kamon modules. I'd expect somewhere in installation guide that user should handle stopping of modules.

PR #591, opened 2019-06-17
AtomicGetOrElseUpdateOnTrieMap::atomicGetOrElseUpdate -> TrieMap::getOrElseUpdate

dpsoft:
* Remove `AtomicGetOrElseUpdateOnTrieMap::atomicGetOrElseUpdate` method in favor of `TrieMap::getOrElseUpdate`

### DIFF
--- a/core/kamon-core/src/main/scala/kamon/Utilities.scala
+++ b/core/kamon-core/src/main/scala/kamon/Utilities.scala
@@ -46,7 +46,7 @@ trait Utilities { self: Configuration =>
     * empty.
     */
   def filter(configKey: String): Filter =
-    _filters.atomicGetOrElseUpdate(configKey, Filter.from(configKey))
+    _filters.getOrElseUpdate(configKey, Filter.from(configKey))
 
   /**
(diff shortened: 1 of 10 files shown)


ivantopo:
Yo @hughsimpson did you explicitly catch an issue from this branch? I see you pushed a commit and I was wondering about it :thinking: 

As you can see, this PR is quite old and I do remember having a reason not to merge it yet, but I cannot remember it anymore. Probably I discussed it privately with Diego and can't find the conversation anymore :shrug:

hughsimpson:
@ivantopo I actually just messed a couple of lines up during a rebase 😅 the second commit was just to fix that. The only behavioural changes I could spot here was an explicit throw in case the value was null, but it can't be for any of the call-sites. Seemed a reasonable change to me...

dpsoft:
Hello, I can't recall the reason why this was never merged—perhaps due to a specific Scala version, but I'm uncertain. The related issue can be found here: https://github.com/scala/bug/issues/7943. I believe it's time to remove the workaround and proceed with the merge. 👍

PR #594, opened 2019-07-21
change the store prefix to runWith on Context storage functions

ivantopo:
After going through all modules updating code and trying to help out people do little bits of manual context propagation it seemed like the new `.storeXXX` methods are not really explicit about what they do. Earlier they were called `.withXXX` (e.g. `Kamon.withSpan(...) {....}`) but they were changed because of inconsistency with other parts of the codebase where all methods starting with `.with` are always returning a new/modified version of the object in which they were called. After some consideration, it seems like the `.runWith` prefix makes it a lot easier to understand what is happening and more importantly, to give implicitly express that there is a scope for the action: only while the provided function runs.

This would be the very last change before release 2.0 and since it doesn't change any semantics but only naming, I would vote for merging and starting the release of all modules.


jalaziz:
This would be great! I found the `store` prefix very confusing myself when upgrading.

PR #596, opened 2019-07-21
reintroduce increment/decrement API on Gauge

jtjeferreira:


kamon-core/src/main/scala/kamon/metric/Gauge.scala
```diff
@@ -42,6 +62,20 @@ object Gauge {
...
+
+    override def increment(times: Double): Gauge = {
+      _currentValue += times
+      this
+    }
+
+    override def decrement(): Gauge = decrement(1)
+
+    override def decrement(times: Double): Gauge = {
+      _currentValue -= times
```
ivantopo: This should guard against negative values, same as `update` does.

kamon-core/src/main/scala/kamon/metric/Gauge.scala
```diff
@@ -42,6 +62,20 @@ object Gauge {
 
     @volatile private var _currentValue = 0D
 
+    override def increment(): Gauge = increment(1)
+
+    override def increment(times: Double): Gauge = {
```
ivantopo: This should guard against negative increments.

jtjeferreira:
> but we could probably think in the near future whether we should allow negative values in a gauge, although doesn't seem like a priority now.

want me to implement that in this PR?

ivantopo:
Not really. It is not that much about allowing it to happen at the Gauge API level but rather to do the research to ensure that if we allow this to happen here, we will not be breaking any reporters or at least modify them to handle the situation. 

It might be my tired self speaking here, but if we are really releasing tomorrow I would very much like to keep the scope of this change as limited as possible.

jtjeferreira:
> It might be my tired self speaking here, but if we are really releasing tomorrow I would very much like to keep the scope of this change as limited as possible.

Tomorrow? :running_man:

ivantopo:
Yes!

kamon-core/src/main/scala/kamon/metric/Gauge.scala
jtjeferreira: @ivantopo should I change `_currentValue` to an `AtomicDouble`?

ivantopo: Is there an AtomicDouble in the JDK?

jtjeferreira: More or less: https://stackoverflow.com/questions/5505460/java-is-there-no-atomicfloat-or-atomicdouble

jtjeferreira:
@ivantopo can you rerun the tests?

ivantopo:
Running again now! I'm not sure why the lifecycle test fails in Travis :(

ivantopo:
Hey @jtjeferreira, I just PRed some tests to your branch: https://github.com/jtjeferreira/Kamon/pull/1

The last 3 tests are failing, something is wrong with increment/decrement, could you please take a look? let me know if you need any help!

ivantopo:
And we are now all green!

PR #597, opened 2019-07-21
add a finishAfter(at: Duration) to Span API,

jtjeferreira:
The idea is that we can finish the span saying that it took 1s.
The use case is for an integration where I receive `CommandSucceededEvent`
and this has an elapsedTimeNanos...

If Span exposed the `_startedAt`, I could implement this myself but then
what would the implementation for `Empty` and `Remote` be?


jtjeferreira:
One unrelated test is failing... Can you re trigger tests?

PR #598, opened 2019-07-24
Update reference.conf

schrepfler:
Typo on B3Single span mapper.


ivantopo:
Oh, after merging noticed that we have the same typo on the incoming part of the config. :man_facepalming:

schrepfler:
😅

PR #600, opened 2019-07-28
Warning cleanup

melezov:
Run some static code analysis - seems to have found a buggy with config param.
Other stuff is just making linters happy.

Scream if you want me to exclude / squash commits.


ivantopo:
Thanks a lot @melezov! :tada:

PR #606, opened 2019-08-29
+ core: provide span propagation codec for "uber-trace-id", default jaeger format

slavaschmidt:
…as described here: https://www.jaegertracing.io/docs/1.7/client-libraries/#propagation-format

### DIFF
--- a/kamon-core/src/main/scala/kamon/trace/SpanPropagation.scala
+++ b/kamon-core/src/main/scala/kamon/trace/SpanPropagation.scala
@@ -25,11 +25,20 @@ import kamon.context.generated.binary.span.{Span => ColferSpan}
 import kamon.context.{Context, _}
 import kamon.trace.Trace.SamplingDecision
 
+import scala.util.Try
+
 
 /**
   * Propagation mechanisms for Kamon's Span data to and from HTTP and Binary mediums.
   */
 object SpanPropagation {
+
+  object Util {
+    def urlEncode(s: String): String = URLEncoder.encode(s, "UTF-8")
+    def urlDecode(s: String): String = URLDecoder.decode(s, "UTF-8")
+  }
+
+  import Util._
 
   /**
     * Reads and Writes a Span instance using the B3 propagation format. The specification and semantics of the B3
@@ -88,8 +97,6 @@ object SpanPropagation {
       case SamplingDecision.Unknown     => None
     }
 
-    private def urlEncode(s: String): String = URLEncoder.encode(s, "UTF-8")
-    private def urlDecode(s: String): String = URLDecoder.decode(s, "UTF-8")
   }
 
   object B3 {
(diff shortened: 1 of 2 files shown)


kamon-core/src/main/scala/kamon/trace/SpanPropagation.scala
```diff
@@ -194,6 +199,80 @@ object SpanPropagation {
...
+    private def decodeSamplingDecision(flags: String) =
+      if (flags.equalsIgnoreCase(DebugFlag)) SamplingDecision.Sample
+      else if (lowestBit(flags).contains(1)) SamplingDecision.Sample
+      else if (lowestBit(flags).contains(0)) SamplingDecision.DoNotSample
+      else SamplingDecision.Unknown
+
+    private def encodeSamplingDecision(samplingDecision: SamplingDecision): Byte = samplingDecision match {
+      case SamplingDecision.Sample      => 1
+      case SamplingDecision.DoNotSample => 0
+      case SamplingDecision.Unknown     => 1 // the sampling decision is mandatory in this format
```
ivantopo: Wouldn't it be safer to default to `DoNotSample` in this case?

kamon-core/src/main/scala/kamon/trace/SpanPropagation.scala
```diff
@@ -194,6 +199,80 @@ object SpanPropagation {
...
+      val span = context.get(Span.Key)
+
+      if (span != Span.Empty) {
+        val parentContext = if (span.parentId != Identifier.Empty) span.parentId.string else Default
+        val sampling = encodeSamplingDecision(span.trace.samplingDecision)
+        val debug: Byte = 0
+        val flags = (sampling + (debug << 1)).toHexString
+        val headerValue = Seq(span.trace.id.string, span.id.string, parentContext, flags).mkString(Separator)
+
+        writer.write(HeaderName, urlEncode(headerValue))
```
ivantopo: Why is the value being url-encoded? I couldn't find anything in the Jaeger docs saying that it should be like that :confused:

slavaschmidt: The `TextMapCodec` on jaeger side can be configured to both flavors (`urlEncoded` or not). The urlEncode reflects the default configuration. To formulate it differently, it didn't worked for us out-of-the box otherwise.

kamon-core/src/main/scala/kamon/trace/SpanPropagation.scala
slavaschmidt: yep, I wasn't sure about that. can you see the change?

ivantopo:
Hey @slavaschmidt, I just cherry-picked this PR and merged via 25cc4ffdce9e87194a9592d9c65f6803a277544b. Thanks a lot for the contribution!

PR #613, opened 2019-10-03
Take the enabled flag into account

DieBauer:
The `kamon.status-page.enabled = false` is always starting the statuspage server.

This PR will stop the server on reload, and on creation.


ivantopo:
Hey @DieBauer, I was looking into it and realized that the actual problem here is that the `kamon.status-page.enabled` setting should have never been there in the first place. All modules have an `enabled` flag that comes by default with Kamon and in the case of the Status Page module the right setting is `kamon.modules.status-page.enabled`, you can see it [here](https://github.com/kamon-io/Kamon/blob/master/kamon-status-page/src/main/resources/reference.conf#L20).

Would it be fine for you to change this PR to remove the `kamon.status-page.enabled` setting instead?

ivantopo:
Closing this one cause #621 fixed this very same thing.

PR #625, opened 2019-12-09
Span error.message from throwable

mladens:
Use throwable message for error.message in spans failed with a throwable

### DIFF
--- a/kamon-core/src/main/scala/kamon/trace/Span.scala
+++ b/kamon-core/src/main/scala/kamon/trace/Span.scala
@@ -517,8 +517,12 @@ object Span {
       if(_isOpen) {
         _hasError = true
 
-        if(isSampled && includeErrorStacktrace)
-          _spanTags.add(TagKeys.ErrorStacktrace, toStackTraceString(throwable))
+        if(isSampled) {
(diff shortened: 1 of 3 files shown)


kamon-core/src/main/scala/kamon/trace/Span.scala
```diff
@@ -517,8 +517,12 @@ object Span {
       if(_isOpen) {
         _hasError = true
 
-        if(isSampled && includeErrorStacktrace)
-          _spanTags.add(TagKeys.ErrorStacktrace, toStackTraceString(throwable))
+        if(isSampled) {
+          _spanTags.add(TagKeys.ErrorMessage, throwable.getMessage)
```
lustefaniak: getMessage might be `null`, from what I remember I had issues with reporters if tag had `null` value

ivantopo: Thanks for the comment @lustefaniak, let's quickly fix and merge this one.

ivantopo: Let me correct myself. I double checked this and both the `TagSet` and `TagSet.Builder` classes will make sure that no null keys nor values make it into a `TagSet` so we are safe here.

PR #637, opened 2020-03-04
Make `context.Storage.Scope` `AutoCloseable`

liff:
For a small convenience increase when using it with `scala.util.Using`, `cats.effect.Resource` or whatever resource management tools one might use.


ivantopo:
Thanks @liff!

PR #763, opened 2020-04-07
Cassandra driver instrumentation

mladens:
Merging cassandra driver instrumentation from kamon-io/kamon-cassandra/pull/6

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -120,6 +120,7 @@ lazy val instrumentation = (project in file("instrumentation"))
     `kamon-logback`,
     `kamon-jdbc`,
     `kamon-mongo`,
+    `kamon-cassandra`,
     `kamon-annotation`,
     `kamon-annotation-api`,
     `kamon-system-metrics`,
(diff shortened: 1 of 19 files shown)


eugenemiretsky:
Is there a plan to merge this?

dispalt:
Im interested too!

osbornk:
What is the status of this? We would love to use this feature.

ivantopo:
I didn't get the chance to look into the actual code behind the new driver yet, only high level review of the APIs and documentation.

The plan I have so far for this PR is to:
  - Merge the PR with support for 3.x driver only. Support for 4.x should come soon after.
  - Remove the "cluster" tag. It was removed from the 4.x driver and wasn't even accurate on 3.x.
  - Rename mentions of "host" to "node". The word node aligns much better with the language on the cassandra documentation.
  - Reduce the number of generated Spans. Out of the box, the instrumentation is generating two spans for every executed query: one for the query and one for the actual round trip to cassandra. It makes sense because there might be more than one round trip if a query needs to be retried or speculated on, but we are still getting a fuckton of useless Spans out of the box:

image

I see a few options for reducing the number of Spans:
  - Turn all the executions/retries into marks in the span. This would allow to keep track of all round trips, while still keeping a single span for the query.
  - Have a setting to enable/disable tracking of executions/retries. 
  - Both? 

I'll be working on this in the next few days, feedback is welcome!

ivantopo:
I just applied the changes I mentioned above. Went for a setting to disable tracking of executions (for reducing the number of Spans).

Changes were pushed to this branch in case anyone wants to take a look!

PR #764, opened 2020-04-08
Migrate kafka consumer/producer instrumentation

mladens:
Contributed by @boxsterman


dpsoft:
@mladens  @boxsterman awesome guys!!!!

eugenemiretsky:
Is there a plan to merge this? Looks pretty cool

ivantopo:
I think we need to make some changes to the way we are handling the consumer side of the instrumentation. In particular, the start time of the consumer Spans is set to the instant when `KafkaConsumer.poll` is called. In realtime scenarios, the poll could have started before the actual records were produced and leads to strange looking traces like this:

image

Also, the consumer is creating N + 1 spans for every poll:
  - One Span for every consumed record. Each one of these Spans is either a child or linked to the original trace on the producer side of the system. These Spans are finished as soon as the poll finishes, though.
  - One Span for the entire duration of the `KafkaConsumer.poll` call. This Span is not linked with anything, it is only meant to track how long it takes to poll records from Kafka. Since this Span is not linked with anything, we end up with single-span traces that aren't useful at all.

I'm wondering:

#### Should we actually have a Span for the `KafkaConsumer.poll`? 
Maybe a metric would be more appropriate. Also, it would be nice if there was a way to know whether the poll time was spent waiting for records (there was nothing new in the topic) or in actual processing/transfer of those records to the consumer.

I would remove the Span and create a metric instead.

#### Should we automatically finish the consumer Spans as we do today? 
When instrumenting a consumer service, I believe the most important goals are:
  - Keep the same context tags as on the producer side. This enables transparent propagation of data across all services connected through a topic.
  - Continue or link to the same trace on the producer side. So that we can figure out where the produced records came from.
  - Track how long it takes to process the consumed record.
  - Track how much lag is experienced by the consumer.

As the instrumentation stands today, we would get a Span that finishes as soon as the poll is done, leaving out any actual processing time on the consumer side. It is necessary for users to manually wrap their message processing logic with a Span if they want to track the processing time. That code would look similar to this:

```scala
    // records are the result of polling on the Kafka consumer
    records.iterator().forEachRemaining(record => {
      val incomingContext = KafkaInstrumentation.getContext(record)
      Kamon.runWithSpan(Kamon.spanBuilder("message-processing").context(incomingContext).start()) {
        // actual message processing code
      }
    })
```

The code above leads to a trace like the following:
image

What I'm thinking to do here is:
  - Drop the poll span.
  - Instead of creating and finishing a message processing Span, expose a friendly method for users to create the Span for their processing logic. Similar to `KafkaInstrumentation.getContext`. See more bellow.
  - Add a tag on the spans with the poll and lag times.


**Preparing the Consumer-side Spans**
The utility for creating the consumer-side spans might look like this:

```scala
val context = KafkaInstrumentation.getContext(consumerRecord)
val span = KafkaInstrumentation.consumerSpan("operationName", consumerRecord)

Kamon.runWithContext(context) {
  Kamon.runWithSpan(span) {
    // Actual processing logic goes here.
  }
}
```
The call to `KafkaInstrumentation.consumerSpan` would create the Span, associate it with the parent, add the lag and poll time tags and leave the Span ready for the user to start processing.

I find it particularly annoying to have the double nesting of `runWithContext` and `runWithSpan`. Maybe we can add a variant of `Kamon.runWithContext` that sets the Context as current AND finishes the Span within the context, if any. 

I'll get back to this later today or early next week. /cc @boxsterman @mladens @dpsoft

ivantopo:
Went ahead with the changes and the current work in progress produces a trace like this:

image

The producer side is completely automatic. The instrumentation creates a new Span for the call to `producer.send` and stores both the Span and the current Context on the `kamon-context` header going to Kafka. There is no user intervention needed.

The consumer side is automatically unpacking the incoming Context and storing it with each ConsumerRecord, but doesn't create any Span in the process. I'm introducing `KafkaInstrumentation.runWithConsumerSpan` to:
  - Read any parent that might be present in the incoming Context.
  - Create and start a Span that will track the processing time of the incoming record.
  - Store the incoming Context and consumer span in the current context.
  - Run the supplied block of code.
  - Finish the processing span.
  - Cleanup the current context.

The user code will end up looking as follows:

```scala
import KafkaInstrumentation.runWithConsumerSpan

records.iterator().forEachRemaining(record => {
  runWithConsumerSpan(record) {
    // The message processing logic goes here
  }
})
```

On top of that, the consumer Span is tagged with `kafka.pollTime` with the time it took for the consumer to poll records. I'm considering to add a tag with the "lag" measured as `currentTime - recordTime` and a metric tracking that value as well, but I'm not really sure about it. I'll investigate more about the timestamps and fix the tests.

Looking forward to any comments.

mladens:
From limited experience with manualy writing producer-processor-consumer logic, i see no use for a `Poll` span, especially since it provides no other information other than "message was fetched within this span", meaning there are no timing information between polled spans, all message spans have same start time and duration.

Most valuable info here is how much is fetched, and how often is it polling (in order to tweak poll parameters). This can be covered by a metric. Or becomes relevant only after nothing can be concluded from higher level metrics (throughput on processor stopped, is it because consumer stopped fetching, not catching up, ...)

For span processing, it really should not matter which message was it polled with together. Simple `poll-started`, `poll-ended`, `polled-total` tags on consumer span could be enough.

 

`Processor` span syntax is great for cases where people write processors manually, any other case should be covered  library instrumentation. Most common case being kafka streams which will carry on context unpacked from message through entire pipeline creating span per stage. Alpakka would be another example (which we dont have covered) 

Few more considerations..

- broadcasts, are we ok with multiple services being involved downstream in the original trace (started by producer)
- Reprocessing and delayed processing. Huge delays between producing and processing might render a lot of visualisations useless.

dpsoft:
I agree with @mladens,  I think I would add just one more thing to what he said:

- drop poll span
- keep consumer span (tagged with all polling information)
- not mix consumer/pooling info with processing stage span tags
- **a tag `lag` and a metric that also reflects the value**

ivantopo:
Updated the PR with the changes above.

Regarding these comments:
> Few more considerations..
> - broadcasts, are we ok with multiple services being involved downstream in the original trace (started by producer)
> - Reprocessing and delayed processing. Huge delays between producing and processing might render a lot of visualisations useless.
The instrumentation comes with the `continue-trace-on-consumer` setting, which lets you decide whether the consumer side continues the same trace or starts a new one with a link to the producer trace.

The delay is a different concern. I'm not really sure of what to do about it yet. Most likely we will ship the first version of the instrumentation without any metrics for the consumer lag and introduce them later when we have a decent idea of how to proceed. Same goes with Streams.

I'll put together a quick documentation page for this instrumentation and merge/publish tomorrow.

PR #765, opened 2020-04-09
kamon-akka: Remove hookHeavyweightWarning logging

DieBauer:
The logging of an ask pattern timeout can be configured by `kamon.instrumentation.akka.ask-pattern-timeout-warning` 

Default is off, however the `hookHeavyweightWarning` is always triggered. 

By removing this case, the setting is actually respected.


ivantopo:
Thanks a lot for the fix! :tada:

PR #768, opened 2020-04-28
play-gRPC: Instrumentation of play-gRPC router

klvmungai:
Added the instrumentation of the play gRPC router.
Still trying to figure out the best way to test this.

### DIFF
--- a/instrumentation/kamon-play/src/main/scala/kamon/instrumentation/play/PlayServerInstrumentation.scala
+++ b/instrumentation/kamon-play/src/main/scala/kamon/instrumentation/play/PlayServerInstrumentation.scala
@@ -68,6 +68,13 @@ class PlayServerInstrumentation extends InstrumentationBuilder {
   private def hasGenericFutureListener(): Boolean = {
     try { Class.forName("io.netty.util.concurrent.GenericFutureListener") != null} catch { case _ => false }
   }
+
+  /**
+    * When using play gRPC, we can reuse the instrumentation that is provided by the akka HTTP module, we just need to
+    * apply the operation name and make a sampling decision.
+    */
+  onType("play.grpc.internal.PlayRouter$$anon$1")
+    .advise(method("apply"), GenerateGRPCOperationName)
 }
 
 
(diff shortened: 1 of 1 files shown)


instrumentation/kamon-play/src/main/scala/kamon/instrumentation/play/PlayServerInstrumentation.scala
```diff
@@ -68,6 +68,13 @@ class PlayServerInstrumentation extends InstrumentationBuilder {
   private def hasGenericFutureListener(): Boolean = {
     try { Class.forName("io.netty.util.concurrent.GenericFutureListener") != null} catch { case _ => false }
   }
+
+  /**
+    * When using play gRPC, we can reuse the instrumentation that is provided by the akka HTTP module, we just need to
+    * apply the operation name and make a sampling decision.
+    */
+  onType("play.grpc.internal.PlayRouter$$anon$1")
```
ivantopo, line 76: Do you know whether this will work fine across all scala versions? :thinking:

instrumentation/kamon-play/src/main/scala/kamon/instrumentation/play/PlayServerInstrumentation.scala
```diff
@@ -239,3 +246,13 @@ object GenerateOperationNameOnFilterHandler {
   })
 
 }
+
+object GenerateGRPCOperationName {
+  @Advice.OnMethodEnter
+  def enter(@Advice.Argument(0) request: akka.http.scaladsl.model.HttpRequest): Unit = {
+    val span = Kamon.currentSpan()
+    span.name(request.uri.toRelative.toString)
```
ivantopo, line 254: Little nitpicking: we usually use the fluid API when doing several "things" on the same Span!

instrumentation/kamon-play/src/main/scala/kamon/instrumentation/play/PlayServerInstrumentation.scala
Tochemey: Currently play-grpc has been cross-compiled with scala 2.12 and 2.13.

Tochemey:
@ivantopo I did some tests against a real application and everything looks great. The application has been built on 2.13 though. Both REST and gRPC sampling worked as expected. As for the unit test we will have to figure out a way to mimic some tests.

ivantopo:
Hey folks!

I spent some time toying around with SBT so that we can set up some tests for this PR. A continuation of this PR is now on #839. I'm going to close this PR in favor of #839. Please take a look at it :pray:

PR #771, opened 2020-05-03
Configurable operation name mappings for http client instrumentation

### LINKED ISSUE #769: Assign operation names to HTTP client requests from config
It is possible to assign names to HTTP operations from configuration, but only for the server side. @lacarvalho91 pointed out [on Gitter](https://gitter.im/kamon-io/Kamon?at=5ea269f111ffed12682f260d) that it would be great to have this for the client side as well and indeed, that would be great!

lacarvalho91:
Closes #769

I can't seem to get the project to build locally to be able to run the tests. so I'm gonna have to leave that to Travis.

Please let me know if I've missed anything.

### DIFF
--- a/instrumentation/kamon-jdbc/src/test/scala/kamon/instrumentation/jdbc/StatementInstrumentationSpec.scala
+++ b/instrumentation/kamon-jdbc/src/test/scala/kamon/instrumentation/jdbc/StatementInstrumentationSpec.scala
@@ -89,7 +89,7 @@ class StatementInstrumentationSpec extends WordSpec with Matchers with Eventuall
           statement.execute()
           validateNextRow(statement.getResultSet, valueNr = 1, valueName = "foo")
 
-          eventually(timeout(scaled(5 seconds))) {
+          eventually(timeout(scaled(5 seconds)), interval(200 millis)) {
             val span = testSpanReporter().nextSpan().value
             span.operationName shouldBe StatementTypes.GenericExecute
             span.metricTags.get(plain("component")) shouldBe "jdbc"
--- a/instrumentation/kamon-akka-http/src/main/scala/kamon/instrumentation/akka/http/AkkaHttpInstrumentation.scala
+++ b/instrumentation/kamon-akka-http/src/main/scala/kamon/instrumentation/akka/http/AkkaHttpInstrumentation.scala
@@ -1,11 +1,9 @@
 package kamon.instrumentation.akka.http
 
-import akka.http.scaladsl.model.{HttpHeader, HttpRequest, HttpResponse}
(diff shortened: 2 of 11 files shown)


lacarvalho91:
I don't think that latest Travis build failure is to do with anything I've changed? Can we rerun the build?

instrumentation/kamon-instrumentation-common/src/main/scala/kamon/instrumentation/http/HttpInstrumentation.scala
```diff
@@ -0,0 +1,21 @@
...
+
+import kamon.util.Filter
+
+object HttpInstrumentation {
+
+  final case class OperationNameSettings(defaultOperationName: String,
+                                         operationMappings: Map[Filter.Glob, String],
+                                         operationNameGenerator: HttpOperationNameGenerator)
+
+  private[http] def operationName(settings: OperationNameSettings, request: HttpMessage.Request): String = {
```
ivantopo, line 11: Why not make this method a member of the `OperationNameSettings` class? :thinking:

lacarvalho91: yeah don't see why not! i've made that change

lacarvalho91:
thanks for the review @ivantopo! I have addressed your comment 🙂

instrumentation/kamon-jdbc/src/test/scala/kamon/instrumentation/jdbc/StatementInstrumentationSpec.scala
```diff
@@ -89,7 +89,7 @@ class StatementInstrumentationSpec extends WordSpec with Matchers with Eventuall
           statement.execute()
           validateNextRow(statement.getResultSet, valueNr = 1, valueName = "foo")
 
-          eventually(timeout(scaled(5 seconds))) {
+          eventually(timeout(scaled(5 seconds)), interval(200 millis)) {
```
lacarvalho91, line 92: This test seemed really flaky, pr build failed a few times because of it. Seems OK with this change, although could just be coincidence

lacarvalho91:
@ivantopo anymore thoughts on this?

instrumentation/kamon-jdbc/src/test/scala/kamon/instrumentation/jdbc/StatementInstrumentationSpec.scala
ivantopo: Thanks!

ivantopo:
Thanks a lot @lacarvalho91!

PR #773, opened 2020-05-06
Prometheus: append  postfix only if missing

eugenemiretsky:
Append metric name postfix only if missing. 
It help
1) Avoid weird metric names like "my_counter_total_total"
2) Allow naming metrics in a way that results in same metric name with tools that don't rename metrics (statsD Exportorter) 
3) Standardizes with Micrometer

### DIFF
--- a/reporters/kamon-prometheus/src/main/scala/kamon/prometheus/ScrapeDataBuilder.scala
+++ b/reporters/kamon-prometheus/src/main/scala/kamon/prometheus/ScrapeDataBuilder.scala
@@ -51,7 +51,7 @@ class ScrapeDataBuilder(prometheusConfig: PrometheusSettings.Generic, environmen
 
   private def appendCounterMetric(metric: MetricSnapshot.Values[Long]): Unit = {
     val unit = metric.settings.unit
-    val normalizedMetricName = normalizeMetricName(metric.name, unit) + "_total"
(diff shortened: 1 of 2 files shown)


ivantopo:
Hey @eugenemiretsky, thanks a lot for the PR!

The changes look fine but I do have some questions about items 2. and 3. from your list. Could you please share a bit more details of why 2. and 3. were necessary or what problems were you seeing without this changes? 

That info  will help us understand if something else should be taken into account before merging and how would this affect users who already have these issues!

eugenemiretsky:
We have the same service deployed in serveral environment
1) K8s, where the Kamon Prometheus  plugin is used to export  metrics
2) AWS, where we set metric  to a Prometheus statsD exporter (using Kamon Datadog, though we should probably switch to the statsD plugin ). Because of the renaming, we cannot use the same dashboard in all environments - one way around it is to name all our metrics  using the Prometheus naming convention, so we have the same metric names n statsD and as with Kamon Prometheus. Micrometer does something similar to this PR.

eugenemiretsky:
Generally though, I feel like if as a develop I already follow the Prometheus naming convention, it would be weird to see my metrics as my_counter_total_total

eugenemiretsky:
Coming to think of it, would be nice too have Kamon itself follow similar naming convention  (or have an option to turn it on )

ivantopo:
Thanks a lot @eugenemiretsky!

PR #775, opened 2020-05-14
Prometheus: Add exportation of Summary based on Distribution

### LINKED ISSUE #751: Allow selecting distribution metrics to be exposed as summaries instead of histograms
Take a look [here](https://gitter.im/kamon-io/Kamon?at=5acf18021130fe3d36cb986d) and a bit bellow.

As discussed by @StephenKing and @kodemaniak there are a few metrics that could or maybe even should be exposed as summaries instead of histograms, maybe even gauges. I personally prefer the idea of summaries cause they can tell a bit more about behavior than a simple gauge but I'm open to discuss it :). These changes would be most certainly applied to JVM and System metrics when exposed to Prometheus.

TjarkoG:
As discussed in #751 there can be the wish to expose summaries.
This PR adds the possibility to configure metrics that are exposed as
Summaries and a way to do this in addition to histograms or as a
replacement. Default quantiles are provided in reference.conf but can be overwritten.

closes #751

### DIFF
--- a/reporters/kamon-prometheus/src/main/resources/reference.conf
+++ b/reporters/kamon-prometheus/src/main/resources/reference.conf
@@ -58,6 +58,26 @@ kamon.prometheus {
     }
   }
 
+  summary: {
+    # Determines if summaries should be added in addition to or instead of prometheus-histogram
+    exclusive: false
+
+    # Histograms that should be reported as summary. Notation in Glob syntax.
+    metrics: [
+      # example:
+      # "request*",
+      # "*akka*"
+    ]
+
+    quantiles: [
+      # Quantiles that should be reported (will be translated to the according percentiles)
+      0.5,
+      0.75,
+      0.95,
+      0.99,
+      0.999
+    ]
+  }
 
   embedded-server {
 
--- a/reporters/kamon-prometheus/src/main/resources/reference.conf
+++ b/reporters/kamon-prometheus/src/main/resources/reference.conf
@@ -58,10 +58,7 @@ kamon.prometheus {
     }
   }
 
-  summary: {
-    # Determines if summaries should be added in addition to or instead of prometheus-histogram
-    exclusive: false
-
+  summaries: {
     # Histograms that should be reported as summary. Notation in Glob syntax.
     metrics: [
       # example:
--- a/reporters/kamon-prometheus/src/main/resources/reference.conf
+++ b/reporters/kamon-prometheus/src/main/resources/reference.conf
@@ -58,16 +58,16 @@ kamon.prometheus {
     }
   }
 
-  summaries: {
-    # Histograms that should be reported as summary. Notation in Glob syntax.
-    metrics: [
+  summaries {
+    # Names of the histogram, timer and range sampler metrics that must be exported as summaries. All patterns are treared as Glob patterns.
+    metrics = [
       # example:
       # "request*",
       # "*akka*"
     ]
 
-    quantiles: [
-      # Quantiles that should be reported (will be translated to the according percentiles)
+    # Quantiles that should be reported (will be translated to the according percentiles)
+    quantiles = [
       0.5,
       0.75,
       0.95,
--- a/reporters/kamon-prometheus/src/main/scala/kamon/prometheus/ScrapeDataBuilder.scala
+++ b/reporters/kamon-prometheus/src/main/scala/kamon/prometheus/ScrapeDataBuilder.scala
@@ -87,13 +87,9 @@ class ScrapeDataBuilder(prometheusConfig: PrometheusSettings.Generic, environmen
 
   private def appendDistributionMetric(metric: MetricSnapshot.Distributions): Unit = {
     val reportAsSummary = prometheusConfig.summarySettings.metricMatchers.exists(_.accept(metric.name))
-    (reportAsSummary, prometheusConfig.summarySettings.exclusive) match {
-      case (true, true) =>
+    if (reportAsSummary) {
         appendDistributionMetricAsSummary(metric)
-      case (true, false) =>
-        appendDistributionMetricAsSummary(metric)
-        appendDistributionMetricAsHistogram(metric)
-      case _ =>
+    } else {
         appendDistributionMetricAsHistogram(metric)
     }
   }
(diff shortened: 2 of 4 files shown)


TjarkoG:
The tests are failing due to flaky tests on the master branch.
Is there something i can do to make the checks green?

TjarkoG:
Hi
I'm Sorry to ask but we would need this feature for our services and don't want to publish the fork to our private maven if that can be avoided.
Are there any thoughts on this PR?

reporters/kamon-prometheus/src/main/scala/kamon/prometheus/ScrapeDataBuilder.scala
```diff
@@ -107,6 +120,33 @@ class ScrapeDataBuilder(prometheusConfig: PrometheusSettings.Generic, environmen
...
+      if(instrument.value.count > 0) {
+        appendSummaryQuantiles(
+          normalizedMetricName,
+          instrument.tags,
+          instrument.value,
+          unit,
+          prometheusConfig.summarySettings.quantiles
+        )
+        appendTimeSerieValue(normalizedMetricName, instrument.tags, format(instrument.value.max), "_max")
+        appendTimeSerieValue(normalizedMetricName, instrument.tags, format(instrument.value.min), "_min")
```
ivantopo, line 142: I think both `_min` and `_max` are super valuable pieced of data, but they are not part of the Prometheus spec. Do you have any idea of whether having those fields could affect Prometheus' behaviour?

reporters/kamon-prometheus/src/main/scala/kamon/prometheus/ScrapeDataBuilder.scala
```diff
@@ -126,6 +166,25 @@ class ScrapeDataBuilder(prometheusConfig: PrometheusSettings.Generic, environmen
       }
     )
 
+  private def appendSummaryQuantiles(name: String,
+                                     tags: TagSet,
+                                     distribution: Distribution,
+                                     unit: MeasurementUnit,
+                                     quantiles: Seq[java.lang.Double]): Unit = {
+    val percentileIter = distribution.percentilesIterator
+    val percentiles = quantiles.sorted.map { quant =>
```
ivantopo, line 175: It would be simpler to call `Distribution.percentile(...)` to get those values out. Under the hood it kind of does the same you are doing here.

reporters/kamon-prometheus/src/main/resources/reference.conf
```diff
@@ -58,6 +58,26 @@ kamon.prometheus {
     }
   }
 
+  summary: {
```
ivantopo, line 61: Maybe call it `summaries` instead? Just because the previous one is `buckets`, in plural.

reporters/kamon-prometheus/src/main/resources/reference.conf
```diff
@@ -58,6 +58,26 @@ kamon.prometheus {
...
+    # Determines if summaries should be added in addition to or instead of prometheus-histogram
+    exclusive: false
```
ivantopo, line 63: I just found this on the [Prometheus docs](https://prometheus.io/docs/instrumenting/exposition_formats/#comments-help-text-and-type-information):

> Only one TYPE line may exist for a given metric name.

So, a metric should be reported as a summary or a histogram, but never both. We should remove this setting and ensure that all metrics selected to be reported as summaries are only reported as summaries.

reporters/kamon-prometheus/src/main/scala/kamon/prometheus/ScrapeDataBuilder.scala
TjarkoG: That was my first implementation. To my understanding, calling Distribution.percentile(...) would create an iterator for every percentile that is requested and iterate through it until the first is reached that is >= the wanted. 
This would result in iterating over the same percentiles multiple times. Thats why i thought creating one iterator and using it for all percentiles might be better for performance

reporters/kamon-prometheus/src/main/scala/kamon/prometheus/ScrapeDataBuilder.scala
TjarkoG: your right
Couldn't find anything about min and max...
removed both just to be save (user can allways get max with the configuration of the 1.0 quantile)

reporters/kamon-prometheus/src/main/resources/reference.conf
TjarkoG: done

reporters/kamon-prometheus/src/main/resources/reference.conf
TjarkoG: you're right. removed that config value and made it exclusive for every summary

reporters/kamon-prometheus/src/main/scala/kamon/prometheus/ScrapeDataBuilder.scala
ivantopo: Indeed, your code will perform better!

reporters/kamon-prometheus/src/main/resources/reference.conf
```diff
@@ -58,6 +58,23 @@ kamon.prometheus {
     }
   }
 
+  summaries: {
```
ivantopo, line 61: Just a bit of nitpicking on this section and we are good to go:
  - Please, remove the `:` at the starting of each section. We don't use them anywhere else in our configurations and would prefer them to keep a consistent style.
  - Suggested text for the `metrics`: "Names of the histogram, timer and range sampler metrics that must be exported as summaries. All patterns are treared as Glob patterns."
  - For the quantiles section, please put the comment above the `quantiles` config.

TjarkoG: No problem. ^^
I agree with those points and changed them accordingly

ivantopo:
Thanks a lot @TjarkoG, this will be out in the next release, most likely towards the end of the week :tada:

TjarkoG:
Perfect
Thanks allot

Issue #629, opened 2019-12-18
classloader leak with playframework dev run


ornicar:
`kamon.Kamon.counter("foo")` causes a classloader leak in playframework when using the dev workflow with `run`.

Here's a minimal project with instructions to reproduce the issue: https://github.com/ornicar/play-cdi-leak/tree/kamon

My setup:
- Linux 5.4.2
- openjdk version "13.0.1" 2019-10-15
- scala 2.13.1
- playframework 2.8.0
- sbt 1.3.5

[edit] Same issue on playframework side: https://github.com/playframework/playframework/issues/9935

ivantopo:
Hey @ornicar, we released a new version of the SBT plugin that clears up a part of Kanela that was retaining most of the memory so far. It is not a definitive solution, but it should improve the situation. Please give it a try, it is version [2.0.6](https://github.com/kamon-io/sbt-kanela-runner/releases/tag/v2.0.6) of the sbt-kanela-runner plugin.

ornicar:
Hello, thanks for that, but I don't use the kanela runner.

ivantopo:
Hey @ornicar! 

In case this is still bothering you, we just released Kamon `2.1.11` with a few changes that should fix the memory leak you were experiencing with Play in development mode. I tried it with the example you provided and it does fix the issue! Although it is necessary to apply a small change to the application loader to call `Kamon.stop()`:

```scala
class MyApplicationLoader extends ApplicationLoader {
  def load(context: ApplicationLoader.Context): Application = {
    // This block is new and necessary.
    context.lifecycle.addStopHook { () =>
      Kamon.stop()
    }

    new MyComponents(context).application
  }
}
```
Thanks to @jtjeferreira for making this possible!

PR #780, opened 2020-05-21
Add configuration to disable Kamon

jtjeferreira:
Add a configuration flag `kamon.enabled` which is true by default. Should provide an workaround for https://github.com/kamon-io/Kamon/issues/629

### DIFF
--- a/core/kamon-core/src/main/scala/kamon/util/NoOpExecutorService.java
+++ b/core/kamon-core/src/main/scala/kamon/util/NoOpExecutorService.java
@@ -1,0 +1,108 @@
+package kamon.util;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.*;
+
+/**
+ * An executor service that does nothing. Used with empty metrics so that no
+ * threads are created wasting time / space.
+ */
+public class NoOpExecutorService implements ScheduledExecutorService {
+  @Override
+  public ScheduledFuture<?> schedule(Runnable runnable, long l,
+                                     TimeUnit timeUnit) {
+    return null;
+  }
+
+  @Override
+  public <V> ScheduledFuture<V> schedule(Callable<V> vCallable, long l,
+                                         TimeUnit timeUnit) {
+    return null;
+  }
+
+  @Override
+  public ScheduledFuture<?> scheduleAtFixedRate(
+    Runnable runnable, long l, long l1, TimeUnit timeUnit
+  ) {
+    return null;
+  }
+
+  @Override
+  public ScheduledFuture<?> scheduleWithFixedDelay(
+    Runnable runnable, long l, long l1, TimeUnit timeUnit
+  ) {
+    return null;
+  }
+
+  @Override
+  public void shutdown() {
+  }
+
+  @Override
+  public List<Runnable> shutdownNow() {
+    return null;
+  }
+
+  @Override
+  public boolean isShutdown() {
+    return false;
+  }
+
+  @Override
+  public boolean isTerminated() {
+    return false;
+  }
+
+  @Override
+  public boolean awaitTermination(long l, TimeUnit timeUnit)
+    throws InterruptedException {
+    return false;
+  }
+
+  @Override
+  public <T> Future<T> submit(Callable<T> tCallable) {
+    return null;
+  }
+
+  @Override
+  public <T> Future<T> submit(Runnable runnable, T t) {
+    return null;
+  }
+
+  @Override
+  public Future<?> submit(Runnable runnable) {
+    return null;
+  }
+
+  @Override
+  public <T> List<Future<T>> invokeAll(
+    Collection<? extends Callable<T>> callables)
+    throws InterruptedException {
+    return null;
+  }
+
+  @Override
+  public <T> List<Future<T>> invokeAll(
+    Collection<? extends Callable<T>> callables, long l, TimeUnit timeUnit
+  ) throws InterruptedException {
+    return null;
+  }
+
+  @Override
+  public <T> T invokeAny(Collection<? extends Callable<T>> callables)
+    throws InterruptedException, ExecutionException {
+    return null;
+  }
+
+  @Override
+  public <T> T invokeAny(Collection<? extends Callable<T>> callables, long l,
+                         TimeUnit timeUnit)
+    throws InterruptedException, ExecutionException, TimeoutException {
+    return null;
+  }
+
+  @Override
+  public void execute(Runnable runnable) {
+  }
+}
--- a/core/kamon-core/src/main/resources/reference.conf
+++ b/core/kamon-core/src/main/resources/reference.conf
@@ -4,6 +4,7 @@
 
 kamon {
 
+  enabled = true
   environment {
 
     # Identifier for this service.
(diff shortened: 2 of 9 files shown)


core/kamon-core/src/main/scala/kamon/util/NoOpExecutorService.java
```diff
@@ -0,0 +1,108 @@
...
+
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.*;
+
+/**
+ * An executor service that does nothing. Used with empty metrics so that no
+ * threads are created wasting time / space.
+ */
+public class NoOpExecutorService implements ScheduledExecutorService {
```
ihostage, line 11: I think, that `NoopModuleRegistry` and `NoOpExecutorService` should have the same prefix 😄

jtjeferreira: fixed :)

prongs:
@ihostage @ivantopo any plans for merging this? Can we expect this to be part of next release?

ivantopo:
Hey folks :wave: 

I'm wondering: what exactly does it mean to disable Kamon? I had a short exchange with @jtjeferreira where he mentioned that his specific use case was to disable Kamon when running a Play Framework app in dev mode. And, in his use case, there is no Kanela or any sort of instrumentation going on while running on dev mode.

My understanding from the code is that this PR will:
  - Prevent modules from being started. That is, no reporters, no Host/Process/JVM metrics being collected.
  - Make the tracer drop all spans. 

Still, if the users are running with Kanela we will still:
  - Have the overhead of transforming the classes during startup.
  - Perform all the usual work that instrumentation does: context propagation, tracing and metrics collection will still happen.

If we want to have a proper on/off switch for Kamon, I think it should take into account the ability to disable instrumentation completely so that it can be used both in development and when deploying.

That said, I think that we can re-purpose this PR to just cover an on/off switch for Kamon's module loader. If we introduce a `kamon.modules.enabled=yes/no` setting doing exactly what this PR does to the `ModuleRegistry` and pair it with `kamon.trace.sampler=never` then we get the same results.

It looks like we can introduce the `kamon.modules.enabled` setting fairly quickly, it is almost done in this PR. On the other hand, the proper on/off switch will require some changes in Kanela and the bundle and that will certainly take some extra time.

@jtjeferreira what do you think about going in this revised direction for this PR?

jtjeferreira:
I think one of the important fixes is https://github.com/kamon-io/Kamon/pull/780/files#diff-af2b22035380418c6576b5db9556193dR31, that I think will fix #629. However your proposal will not address this...

ivantopo:
Wait a sec :exploding_head: 

The scheduler is what is causing the classes leak? How did you figure that out? Awesome!

I would be super happy to include that fix as well! What I'm trying to avoid is having a feature to disable Kamon, but that only half-disables Kamon (since Kanela keeps working).

jtjeferreira:
> The scheduler is what is causing the classes leak? 

I dont have proof, but for sure the threads created by that scheduler dont go away...

ivantopo:
I went digging into this and indeed, everything related with the module registry is causing the memory leaks.

When running on development mode, application reloads are not a problem. Every time you stop the application (by pressing Enter) and start it again (by running the `run` command) Kamon a new ClassLoader is created, though. All ClassLoaders where Kamon was loaded are kept alive because the scheduler threads are never cleared. 

I had a sample app reloading every 4 seconds and it was working as expected, but every time I did a stop/run cycle a new set of threads were created. You will see it better here:

image

We must fix this now, for sure. 

Before getting there, I wanted to share a bit of context around the issue. This should help us take a good decision that benefits most users: 
  - Over the years I got several requests for enabling/disabling Kamon. When people ask for this, they are (mostly) expecting two things: for the toggle to enable/disable everything (meaning, instrumentation and reporters) and, to be able to switch that toggle at runtime. I know that this is not exactly @jtjeferreira's case, but we should keep the expectations of the broader user base in mind.
  - When we introduced the ability to reconfigure Kamon (via calls to `Kamon.reconfigure(...)`), the goal was to allow Kamon's core components to be "on" all the time. If users wanted to change settings, they could reconfigure Kamon at runtime and see the effects immediately. One of the projects that are on the back of my head that would use this is changing the trace sampling configuration via the Kamon Status page. 
  - We cannot disable instrumentation at runtime. Once the application started and the agent is attached, there is no coming back. Trying to reconfigure Kamon with settings that disable instrumentation will not yield the expected result. Disclaimer so that @dpsoft doesn't kill me: certain instrumentations can be rolled back at runtime and Kanela is able to do it, but that only apply to very specific kinds of instrumentation. It is safer to work on the assumption that we can't disable instrumentation at runtime so that whatever we decide to do, it will work with all instrumentation.
  - We cannot enable instrumentation after the target classes are loaded. Same as with the item above, in some cases it might be possible, but not all.
  - Kamon and Kanela don't share configuration. If you call `Kamon.init` or `Kamon.reconfigure` with a custom configuration object, Kamon will see it but Kanela wont. Kanela lives on the System ClassLoader and has its own shaded version of Typesafe Config. This is something that @dpsoft and I have been talking about for some time already but there is no concrete path to move forward just yet.

Putting all the info together, here is the plan I see:
  1. Making it "official" that instrumentation settings cannot be changed after Kamon has been initialized. This is already happening, but we need to ensure that users are aware of it.
  2. Implement the `kamon.enabled` setting as a hard switch to do what is described on this PR and a bit more:
    - Disable sampling on the tracer
    - Disable module loading
    - Use the noop scheduler (see item 4.)
    - Warn users when they attempt to change this setting at runtime. Once Kamon has been initialized, the `kamon.enabled` setting is saved somewhere. Any calls to `Kamon.reconfigure` that try to change the `kamon.enabled` setting will produce a warning in the logs saying that their attempt to change `kamon.enabled` at runtime doesn't have any effect.
  3. Make the `ModuleRegistry` handle its schedulers dynamically. By this I mean, only start the metric and span tickers if any reporters are running. Also, stop the tickers when reporters are no longer running. After calling `Kamon.stopModules()`, the `ModuleRegistry` should wind down all the modules and schedulers.
  4. Make the Kamon scheduler dynamic. This one is tricky. Every metric instrument in Kamon can scheduler auto updates. The range sampler requires a scheduler by definition. This means that after an application is running, there will be metric instruments pointing to the scheduler. The only way I see for releasing the Kamon scheduler is to clear the metric registry as well. We could introduce a `Kamon.stop()` method for this. I suggest a different method because it wouldn't just stop the modules but drop all metrics that were created up to this point. In theory, users should never stop and init Kamon again. We have to make that clear in the docs so everybody is aware of what to expect.
  5. Ensure that the Play's `GuiceModule` is calling the right method in development mode. See [here](https://github.com/kamon-io/Kamon/blob/master/instrumentation/kamon-play/src/main/scala/kamon/instrumentation/play/GuiceModule.scala#L26-L31) for more details. There we should probably call the new `Kamon.stop()` method, regardless of being in development mode or not.
  6. Make Kanela aware of `kamon.enabled`. This is out of the scope of this particular PR, but part of the overall solution. I think @dpsoft and I can handle it in parallel to this PR.

I believe that items 3., 4. and 5. are absolutely necessary. We don't want users to get Kamon into their apps and start getting memory leaks by default. Using the `kamon.enabled` setting should be about optimizing the flow (because no time will be spent on modifying classes) and avoid sending garbage to metric/trace backends (because the reporters will not even be loaded), not about solving a problem introduced out of the box.

What do you think? /cc @jtjeferreira @dpsoft

jtjeferreira:
Looks good to me. I will give it a try

dpsoft:
Looks good to me.

jtjeferreira:
Hi! 

I gave this a try today, and inspired by your suggestion to introduce a `stop` method, I decided to start with a fresh PR, that adds that method to address the memory leaks. Please have a look at #804.

I think after getting that right, we can finish the enable/disable flag in this PR...

JR-Utily:
Any news about the plan as described by @ivantopo ?

jtjeferreira:
Superseded by #804

Issue #772, opened 2020-05-04
Distribution.percentile(rank), returns a percentile value with a different rank that was requested.


jkwatson:
I'm not sure exactly what the circumstances are that causes this, but we've seen it happen quite often, and it leads to unpredictable results in the data being reported externally.

jkwatson:
Any update on this?

jkwatson:
ping!

PR #784, opened 2020-06-03
Create a gauge for the percentile requested, not the one that was returned

jkwatson:
This is a work around for issue #772 

/cc @breedx-nr


ivantopo:
Hey @jkwatson, let me get that #772 fixed right away. It might not be really necessary to make this patch!

jkwatson:
If #772 is fixed, then this can be closed!

ivantopo:
Closed because #772 was fixed and released as [v2.1.1](https://github.com/kamon-io/Kamon/releases/tag/v2.1.1)

PR #786, opened 2020-06-09
New Relic reporters: Make audit logging an option, rather than having it always enabled

jkwatson:


reporters/kamon-newrelic/src/main/scala/kamon/newrelic/spans/SpanBatchSenderBuilder.scala
```diff
@@ -37,10 +37,14 @@ class SimpleSpanBatchSenderBuilder() extends SpanBatchSenderBuilder {
       logger.error("No Insights Insert API Key defined for the kamon.newrelic.nr-insights-insert-key config setting. " +
         "No spans will be sent to New Relic.")
     }
+    val enableAuditLogging = if (nrConfig.getIsNull("enable-audit-logging")) false else nrConfig.getBoolean("enable-audit-logging")
```
dpsoft, line 40: are you expecting that somebody set `enable-audit-logging` = null? I mean by using `getIsNull`?

jkwatson: I hope not, but I'm not 100% familiar with how typesafe config works, and since this is a new setting, I suspected that people might not have the setting set at all. Is there a better way to handle this?

dpsoft: I think that with `nrConfig.getBoolean("enable-audit-logging")` is enough because the value that you set in the `reference.conf` acts like the default value the can be or not overwritten in the `application.conf` in this way is safe use  `nrConfig.getBoolean("enable-audit-logging")`.

jkwatson: cool. I'll make that change. Thanks for the info!

PR #791, opened 2020-06-18
Update New Relic SDK version and allow ingest URI override

### LINKED ISSUE #789: New Relic reporters: enable overriding the URLs for the span and metric ingest
We should allow New Relic users to override the default ingest endpoints for metrics and spans. This could be useful for users behind proxies, etc.

Please assign this issue to @abhihub

breedx-nr:
This resolves #789.

Updates to the latest version of the NR telemetry SDK (0.6.0) and uses the new style configuration-based builders.  It also provides configuration items for overriding the ingest URIs, which is at the core of #789.

Thanks for your consideration.

### DIFF
--- a/reporters/kamon-newrelic/src/main/scala/kamon/newrelic/metrics/NewRelicMetricsReporter.scala
+++ b/reporters/kamon-newrelic/src/main/scala/kamon/newrelic/metrics/NewRelicMetricsReporter.scala
@@ -5,6 +5,8 @@
 
 package kamon.newrelic.metrics
 
+import java.net.{URI, URL}
+
 import com.newrelic.telemetry.SimpleMetricBatchSender
 import com.newrelic.telemetry.metrics.{MetricBatch, MetricBatchSender}
 import com.typesafe.config.Config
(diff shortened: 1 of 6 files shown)


reporters/kamon-newrelic/src/main/scala/kamon/newrelic/metrics/NewRelicMetricsReporter.scala
```diff
@@ -77,12 +79,18 @@ object NewRelicMetricsReporter {
...
+    val userAgent = s"newrelic-kamon-reporter/${LibraryVersion.version}"
+
+    val senderConfig = MetricBatchSender.configurationBuilder()
+      .apiKey(nrInsightsInsertKey)
+      .auditLoggingEnabled(enableAuditLogging)
+      .secondaryUserAgent(userAgent)
 
-    if (enableAuditLogging) {
-      builder.enableAuditLogging()
+    if(nrConfig.hasPath("metric-ingest-uri")) {
```
jkwatson, line 89: should we add a test case to make sure this doesn't blow up somehow?

breedx-nr:
@dpsoft thanks for the quick review, it's greatly appreciated!  I am going to close this for a bit while we fix something important in our upstream sdk and add a test as @jkwatson suggested.

breedx-nr:
Ok, ready for another look.  Thanks all!

PR #797, opened 2020-06-28
Fix bug that causes loss of context when using cats-effect IO timer

bplommer:
Adds instrumentation needed to preserve context across the boundary created by IO.sleep.


dpsoft:
@bplommer thanks for proposing this PR!

PR #804, opened 2020-07-06
Add Kamon.stop method

jtjeferreira:
A different approach to #780

### DIFF
--- a/core/kamon-core/src/main/scala/kamon/module/ModuleRegistry.scala
+++ b/core/kamon-core/src/main/scala/kamon/module/ModuleRegistry.scala
@@ -39,6 +39,7 @@ import scala.util.control.NonFatal
 class ModuleRegistry(configuration: Configuration, clock: Clock, metricRegistry: MetricRegistry, tracer: Tracer) {
 
   private val _logger = LoggerFactory.getLogger(classOf[ModuleRegistry])
+  private val _moduleRegistryEC: ExecutionContext = ExecutionContext.fromExecutor(newScheduledThreadPool(1, threadFactory("kamon-module-registry", daemon = true)))
   private val _metricsTickerExecutor = newScheduledThreadPool(1, threadFactory("kamon-metrics-ticker", daemon = true))
   private val _spansTickerExecutor = newScheduledThreadPool(1, threadFactory("kamon-spans-ticker", daemon = true))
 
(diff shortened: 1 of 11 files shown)


core/kamon-core/src/main/scala/kamon/module/ModuleRegistry.scala
```diff
@@ -141,6 +141,8 @@ class ModuleRegistry(configuration: Configuration, clock: Clock, metricRegistry:
     val latch = new CountDownLatch(stoppedSignals.size)
     stoppedSignals.foreach(f => f.onComplete(_ => latch.countDown()))
 
+    _metricsTickerExecutor.shutdown()
+    _spansTickerExecutor.shutdown()
```
jtjeferreira, line 145: this is making tests fail! should I separate the `stopModules` from the stopping of the `ModuleRegistry`?

core/kamon-core/src/main/scala/kamon/package.scala
```diff
@@ -15,14 +15,25 @@
...
-import java.util.concurrent.{Executors, ThreadFactory}
-
+import java.util.concurrent.{Executors, ScheduledExecutorService, ScheduledThreadPoolExecutor, ThreadFactory, TimeUnit}
 import com.typesafe.config.{Config, ConfigUtil}
 
 import scala.collection.concurrent.TrieMap
 
 package object kamon {
 
+  def newScheduledThreadPool(corePoolSize: Int, threadFactory: ThreadFactory): ScheduledExecutorService = {
```
jtjeferreira, line 25: @ivantopo this will make the threads stop if there is no more work in the queues. I got this from https://github.com/typelevel/cats-effect/blob/series/2.x/core/jvm/src/main/scala/cats/effect/internals/IOTimer.scala#L72-L98.

Do you think we should make this configurable instead?

core/kamon-core/src/main/scala/kamon/module/ModuleRegistry.scala
```diff
@@ -39,8 +39,9 @@ import scala.util.control.NonFatal
 class ModuleRegistry(configuration: Configuration, clock: Clock, metricRegistry: MetricRegistry, tracer: Tracer) {
 
   private val _logger = LoggerFactory.getLogger(classOf[ModuleRegistry])
-  private val _metricsTickerExecutor = Executors.newScheduledThreadPool(1, threadFactory("kamon-metrics-ticker", daemon = true))
-  private val _spansTickerExecutor = Executors.newScheduledThreadPool(1, threadFactory("kamon-spans-ticker", daemon = true))
+  private val _moduleRegistryEC: ExecutionContext = ExecutionContext.fromExecutor(newScheduledThreadPool(1, threadFactory("kamon-module-registry", daemon = true)))
```
jtjeferreira, line 42: introduced another EC to be used by shutdown logic...

core/kamon-core/src/main/scala/kamon/package.scala
ivantopo: I wouldn't make it configurable unless there was an effect to end users. For now, leave it as it is.

ivantopo (approved):
Awesome work @jtjeferreira! We are finally be leak-free in Play development mode :tada:

PR #808, opened 2020-07-08
+ kamon-okhttp: introduce the new module kamon-okhttp to provide instrumentation for OkHttp

cspinetta:
This change introduces a module to provide manual and automatic instrumentation for `OkHttp` >= `3.13`.

### Included:

* The class `KamonTracingInterceptor` so you can manually register it on a OkHttp instance.
* Automatic instrumentation by using `Kanela` agent.
* Integration tests.
* Added as dependency in `kamon-bundle`.

### Compatibility:

The new module `kamon-okhttp` supports `OkHttp` >= `3.13.0` and <= `4.7.2`.

I tested it by running the tests in cross-version fashion against `OkHttp` `3.13.0` , `3.13.1`, `3.14.0`, `3.14.9`, `4.0.1`, `4.1.1`, `4.2.2`, `4.3.1`, `4.4.1`,`4.5.0`, `4.6.0`, `4.7.2`.

### DIFF
--- a/instrumentation/kamon-okhttp/src/main/resources/reference.conf
+++ b/instrumentation/kamon-okhttp/src/main/resources/reference.conf
@@ -1,0 +1,106 @@
+# ============================================ #
+# kamon okhttp3 client reference configuration #
+# ============================================ #
+
+kamon.instrumentation.okhttp {
+
+  http-client {
+    #
+    # Configuration for HTTP context propagation.
+    #
+    propagation {
+
+      # Enables or disables HTTP context propagation on this HTTP server instrumentation. Please note that if
+      # propagation is disabled then some distributed tracing features will not be work as expected (e.g. Spans can
+      # be created and reported but will not be linked across boundaries nor take trace identifiers from tags).
+      enabled = yes
+
+      # HTTP propagation channel to b used by this instrumentation. Take a look at the kamon.propagation.http.default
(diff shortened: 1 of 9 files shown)


instrumentation/kamon-okhttp/src/main/resources/reference.conf
```diff
@@ -0,0 +1,106 @@
...
+    name-generator = kamon.okhttp3.DefaultNameGenerator
+    # Metrics for okhttp
+    metrics {
+      enabled = true
+    }
+  }
+}
+
+kanela {
+  show-banner = true
```
dpsoft, line 91: please remove `show-banner` and `log-level` in order to pick the default config

instrumentation/kamon-okhttp/src/main/resources/reference.conf
```diff
@@ -0,0 +1,106 @@
...
+  log-level = "ERROR"
+
+  modules {
+    okhttp3-module {
+      name = "OkHttp 3 Instrumentation Module"
+      stoppable = false
```
dpsoft, line 97: the instrumentation is an `Advice`  so we can stop the module if we want, please change `stoppeable` to `true`

cspinetta:
@dpsoft any plan to introduce this change?

PR #811, opened 2020-07-22
fix the module shutdown order, fixes #533

### LINKED ISSUE #533: Give reporters a final chance to flush data before stopping

ivantopo:
With this PR, reporters will always get a chance to report the last bit of data before being stopped. Fixes #533.


ivantopo:
I'm going to merge this one right away since it is all internal changes only. Coming out in 2.1.4 in a few minutes!

PR #812, opened 2020-07-23
add support for optimistic alignment to PeriodSnapshot.accumulator

perenecabuto:
This change adds support to PeriodSnapshot.Accumulator to do not align tickers when kamon.metric.optimistic-tick-alignment is disabled.

*Motivation:*
Here in my team we use Kamon in many projects and we use accumulators to have different metric ticker times.
As we have very heavy metrics load sent to our metrics DB and our API server has many units,
it uses an entropy strategy to do not send all the messages at the same time from all the units.
So we disable kamon.metric.optimistic-tick-alignment and start Kamon with a random delay and expect that the accumulators respect this policy.


perenecabuto:
Hello, @dpsoft.

Do you have any plans to merge this PR and release it? I've made a private fork, so I'd like to remove it and use the official Kamon.

Best,
Felipe.,

PR #822, opened 2020-07-31
Matching the Stale Period of PrometheusPushgatewayReporter accumulator to the one of PrometheusReporter accumulator

asutalo:
Really not sure if this was intended behaviour or not but when PrometheusPushgatewayReporter is instantiated it creates an instance of PeriodSnapshot accumulator using the default Stale Period of 365 days (as opposed to PrometheusReporter which creates it with a period of 5 years). As a result when using the PGWReporter all of our metrics get cleaned as stale and are not sent out.

Note I haven't signed the CLA yet as I am not sure if this was intended behaviour or not.
Perhaps there's something I missed in the config that would resolve this issue?


florianmutter:
@ivantopo maybe? We would like to use kamon in our new project but are unable to to so if prometheus push does not work.

florianmutter:
@SimunKaracic maybe? Anything we can to to get this in?

SimunKaracic:
I'm not sure why we were hesitant to merge this, it looks like a good, sensible change. 

Thanks for the contribution and the patience :D

PR #823, opened 2020-08-03
B3 debug flag fix

klvmungai:
As per [the B3 propagation spec](https://github.com/openzipkin/b3-propagation#debug-flag), the debug flag should only be considered if it has a value of `1`. This PR rectifies the current logic to ensure that the system falls back to the `X-B3-Sampled` header if the debug flag holds an invalid value.


ivantopo:
Hey @klvmungai, @Tochemey, thanks a lot for the PR!

PR #841, opened 2020-08-28
Allows to disable the akka remote sharding metrics

danischroeter:
Allows to disable the akka remote sharding metrics independent from akka-remote context propagation

This is done in the same manner as kamon-system-metrics are configurable. The benefit of this approach is, that the sharding subsystem is not instrumented at all.

@ivantopo wdyt?


ivantopo:
Thanks @danischroeter! Good and simple solution. Getting it in!

PR #843, opened 2020-08-29
Fix logger name in DatadogSpanReporter.

DWvanGeest:
Just a small error I noticed when firing up the span reporter.


dpsoft:
@DWvanGeest thanks for proposing this PR

PR #845, opened 2020-09-03
Fix wrong response size

kh-evgeniy:
The original solutions set response size equals to string size, but wrote bytes from UTF-8 encoded string. Which in most cases okay, but I encountered difficult to find problem when my string has some non-UTF-8 characters(in my case most likely NaN or Inf). When trying to curl `/metrics` url I was getting: `curl: (18) transfer closed with 6060 bytes remaining to read`


mladens (approved):
LGTM

ivantopo:
Thanks @kh-evgeniy!

PR #848, opened 2020-09-03
= kamon-datadog: Improve filtering tests.

jarimatti:
The "filter other tags" test now tests metric tag filtering, not environment tag filtering. This seems to be the intention of the code.


dpsoft:
@jarimatti thanks for proposing this PR!

PR #849, opened 2020-09-04
Fixes #842 - No longer URL-encode Uber propagation header, properly handle URL-encoded headers on read

### LINKED ISSUE #842: The uber Span propagation doesn't work with urlencoded values
Copying @DWvanGeest message from [Gitter](https://gitter.im/kamon-io/Kamon?at=5f47d28349148b41c9778f70):

> noticed something interesting with the uber/jaeger span propagation. The incoming header has url-encoded : separators, because of this line: https://github.com/kamon-io/Kamon/blob/master/core/kamon-core/src/main/scala/kamon/trace/SpanPropagation.scala#L234. But, the reader doesn't seem to url-decode before splitting on : here: https://github.com/kamon-io/Kamon/blob/master/core/kamon-core/src/main/scala/kamon/trace/SpanPropagation.scala#L242. As a result, it seems like the trace ID is not being propagated. Am I missing something? Is this a bug?
> 
> FWIW the Jaeger Java lib doesn't url-encode the string any longer: jaegertracing/jaeger-client-java#721

The idea to solve this issue would be not to urlencode on the writer, and urldecode on the reader.

DWvanGeest:
As mentioned in #842, the official Jaeger Java client [no longer](https://github.com/jaegertracing/jaeger-client-java/pull/721) URL-encodes the span context, and the [docs](https://www.jaegertracing.io/docs/1.19/client-libraries/#baggage) mention that only baggage headers should be URL-encoded.

This PR is backwards-compatible since it handles both URL and non-URL encoded headers on read.

### DIFF
--- a/core/kamon-core-tests/src/test/scala/kamon/trace/UberSpanPropagationSpec.scala
+++ b/core/kamon-core-tests/src/test/scala/kamon/trace/UberSpanPropagationSpec.scala
@@ -32,14 +32,14 @@ class UberSpanPropagationSpec extends WordSpecLike with Matchers with OptionValu
       val headersMap = mutable.Map.empty[String, String]
       uberPropagation.write(testContext(), headerWriterFromMap(headersMap))
 
-      headersMap.get("uber-trace-id").value shouldBe "1234%3A4321%3A2222%3A1"
+      headersMap.get("uber-trace-id").value shouldBe "1234:4321:2222:1"
     }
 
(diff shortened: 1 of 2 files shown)


DWvanGeest:
@ivantopo let me know what you think, when you have a chance.

SimunKaracic:
Thanks for the contribution, and sorry for the wait! :tada:

PR #851, opened 2020-09-08
Run tests on PR 

SimunKaracic:
Perhaps it would be useful to run tests on pull-request, not only on direct push. 

WDYT?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: Test all the things
 
-on:
-  push:
+on: [push, pull-request]
+
 
 jobs:
   ci:


ivantopo:
Good, we needed this!

ivantopo:
image

hey @SimunKaracic, can you please check the error above? I actually thought the name was right :confused:

SimunKaracic:
I tested it out on my fork and it worked :/ 
checking it out now :+1:

ivantopo:
I think it should have been `pull_request` instead.

SimunKaracic:
Yeah I messed up when cleaning up after testing. Pushing fix

PR #854, opened 2020-09-13
+ kamon-armeria module 

lucasamoroso:
Armeria http server and client instrumentation.

I tested it manually  and added tests based on `okhttp` and `akka-http` instrumentation modules tests.

Still there is work to do like the https/grpc server instrumention, but i think this PR is a good Armeria module starting point.

Thanks to @dpsoft for his help :)


instrumentation/kamon-armeria/src/main/scala/kamon/armeria/instrumentation/converters/FutureConverters.scala
```diff
@@ -0,0 +1,19 @@
...
+
+import scala.concurrent.{ExecutionContext, Future, Promise}
+import scala.util.Try
+
+trait FutureConverters {
+  implicit class FutureConverterOps[T](jFuture: JFuture[T])(implicit ec: ExecutionContext) {
+    def toScala: Future[T] = {
+      val promise = Promise[T]()
+      Future {
+        promise.complete(Try(jFuture.get))
```
ikhoon, line 13: `jFuture.get()` is a blocking IO. It will block Armeria event loop. Armenia always return [CompletableFuture](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html) for asynchronous functions. CompletableFuture is compatible that means you can convert it to Scala Future. 
scala-java8-compat could be a good reference. 😀
https://github.com/scala/scala-java8-compat/blob/8123acd42f5646e41f4d13329fc2ef8f554cc63c/src/main/scala/scala/compat/java8/FutureConverters.scala#L88-L92

lucasamoroso: Thanks for the review, i'll work on it 😄

instrumentation/kamon-armeria/src/main/scala/kamon/armeria/instrumentation/client/ArmeriaHttpClientInstrumentation.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  private val httpClientConfig = Kamon.config.getConfig("kamon.instrumentation.armeria.http-client")
+  private val instrumentation = HttpClientInstrumentation.from(httpClientConfig, "armeria-http-client")
+
+  override def execute(delegate: HttpClient, ctx: ClientRequestContext, req: HttpRequest): HttpResponse = {
+    val requestHandler = instrumentation.createHandler(getRequestBuilder(req), Kamon.currentContext)
+
+    try {
+      ctx.log()
+        .whenComplete()
```
dpsoft, line 58: why not simply  `.whenComplete().thenAccept(log => {` instead of `.toScala`

instrumentation/kamon-armeria/src/main/scala/kamon/armeria/instrumentation/client/ArmeriaHttpClientInstrumentation.scala
```diff
@@ -0,0 +1,70 @@
...
+        .toScala
+        .map(reqLog => requestHandler.processResponse(toKamonResponse(reqLog)))
```
dpsoft, line 60: I think that instead of `map` should be `foreach` because there is not a transformacion just a side effect

instrumentation/kamon-armeria/src/main/scala/kamon/armeria/instrumentation/client/ArmeriaHttpClientInstrumentation.scala
lucasamoroso: That's the simplest solution and my first approach but unfortunately we ran into some compatibility issues between java and scala 2.11. I didn't like  the workarround, i tried something like this:
```
.whenComplete().thenAccept(toJavaConsumer(...))
```
```
def toJavaConsumer[T](consumer: (T) => Unit): Consumer[T] ={
    new Consumer[T] {
      override def accept(t: T): Unit = {
        consumer(t)
      }
    }
  }
```

I liked more this approach with `toScala` but if you think/have a better solution i could change the current implementation.

instrumentation/kamon-armeria/src/main/scala/kamon/armeria/instrumentation/client/ArmeriaHttpClientInstrumentation.scala
lucasamoroso: You're right! I will change it.

SimunKaracic (commented):
A few nitpicks, but otherwise looking good.

Do try and split it into smaller commits in the future!

SimunKaracic (approved):
Thanks!

SimunKaracic:
I tried this out with 
https://github.com/line/armeria-examples/tree/master/annotated-http-service
The example breaks, and I can't get the `/docs` endpoint. Stacktrace: 
```
java.lang.NullPointerException
        at kamon.armeria.instrumentation.server.ServeMethodAdvisor$.around(ArmeriaHttpServerInstrumentation.scala:79)
        at com.linecorp.armeria.server.FallbackService.serve(FallbackService.java:40)
        at com.linecorp.armeria.server.HttpServerHandler.handleRequest(HttpServerHandler.java:384)
        at com.linecorp.armeria.server.HttpServerHandler.channelRead(HttpServerHandler.java:249)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
        at com.linecorp.armeria.server.ArmeriaHttpServerRequestHandler.channelRead(ArmeriaHttpServerRequestHandler.scala:36)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
        at com.linecorp.armeria.server.Http1RequestDecoder.channelRead(Http1RequestDecoder.java:215)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
        at io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:103)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
        at io.netty.channel.CombinedChannelDuplexHandler$DelegatingChannelHandlerContext.fireChannelRead(CombinedChannelDuplexHandler.java:436)
        at io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:324)
        at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:296)
        at io.netty.channel.CombinedChannelDuplexHandler.channelRead(CombinedChannelDuplexHandler.java:251)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
        at io.netty.handler.codec.ByteToMessageDecoder.handlerRemoved(ByteToMessageDecoder.java:253)
        at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:508)
        at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:440)
        at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:276)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
        at io.netty.handler.logging.LoggingHandler.channelRead(LoggingHandler.java:271)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
        at io.netty.handler.flush.FlushConsolidationHandler.channelRead(FlushConsolidationHandler.java:152)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
        at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1410)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
        at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:919)
        at io.netty.channel.epoll.AbstractEpollStreamChannel$EpollStreamUnsafe.epollInReady(AbstractEpollStreamChannel.java:792)
        at io.netty.channel.epoll.EpollEventLoop.processReady(EpollEventLoop.java:475)
        at io.netty.channel.epoll.EpollEventLoop.run(EpollEventLoop.java:378)
        at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:989)
        at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
        at java.lang.Thread.run(Thread.java:748)
``` 

I'll try out the other examples as well, and report back.

ivantopo:
Hey @lucasamoroso,

I took a deeper look at this PR today and found three things that are important to address:
  - **Handling the context scope properly**. I noticed that the `ArmeriaHttpServerRequestHandler` is storing the incoming `Context` instance, and then it gets cleared out on `ArmeriaHttpServerResponseHandler`, but as far as I understand, there is no guarantee that those two calls will happen on the same thread. I started scratching the surface of Armeria's threading model and so far, I can't say for sure that the behavior in this PR is appropriate. My guess is that there is a separation between Armeria's threads (running the event loops and all Armeria-specific code) and the threads that will actually run the user code (i.e. the service we are implementing), but I will need more research and testing to be sure.
  - **Attaching Kamon's Context to Armeria's Channels**. I'm not entirely sure if that is a good idea, specially because of the previous point. Also, I noticed that Armeria's RequestContext can carry additional attributes and that makes me wonder: is there any particular reason why the Kamon context is attached to the channels rather than Armeria's RequestContext?
  - **Operation Naming**. We should probably be moving to a more gRPC-like naming scheme, where the operation names (and endpoints) look more like `ServiceClassName.methodName` rather than using the HTTP path.

Now, we love Armeria and we want this thing merged and ready to roll! So here is what we will be doing in the next couple days:
  - Try to understand the threading model inside Armeria and make sure that we are not leaving dirty threads nor opening race conditions with Kamon's Context propagation.
  - Run all [Armeria examples](https://github.com/line/armeria-examples) with this instrumentation and uncover any issues that might come up (@SimunKaracic already dropped some of that above).
  - Investigate further how the Prometheus and Zipkin integrations are working with naming and metering. There might be a couple interesting details to learn from there!

Looking forward to your ideas and feedback :smiley:

lucasamoroso:
@ivantopo,@SimunKaracic Thanks for the review! <br/>  
I'll be doing some testing with [Armeria examples](https://github.com/line/armeria-examples), for now i've no problems with the **Operation Name** refactor but i need to take a closer look at the other comments.<br/> I'll working on this in the next couple of days!

lucasamoroso:
Hi!</br> In the last couple of days, with @dpsoft help, we've been changing the current implementation. I've been doing this in another repo because is simple to me to do some manual testing the repo is https://github.com/lucasamoroso/kamon-armeria-test. </br> I'am currently doing manual testing with this [armeria-examples](https://github.com/lucasamoroso/kamon-armeria-test/tree/master/src/main/java/example) taken from Armeria. Until now everithing is working fine so as soon as i finish this manual testing i'll push the changes here and i think with this changes we'll be able to work with grpc and https (but i've to test it).

dpsoft:
@ivantopo about:
> * ** Operation Naming **. We should probably be moving to a more gRPC-like naming scheme, where the operation names (and endpoints) look more like `ServiceClassName.methodName` rather than using the HTTP path.

 I'm not sure about moving to a `gRPC-like` naming scheme for all `Armeria` services. Is really common practice have gRPC services mixed with plain http services in an Armeria application. I think that we need other strategy for gRPC naming services in addition to the current one.

lucasamoroso:
Hi! <br/> I pushed a reimplementation of this module, i think  is a better approach. I've been doing manual testing with [armeria-examples](https://github.com/lucasamoroso/kamon-armeria-test/tree/master/src/main/java/example) no big issue found. <br/> In the next couple of days i'll work with these things: <br/> 
* There is a problem with some [armeria-examples](https://github.com/lucasamoroso/kamon-armeria-test/tree/master/src/main/java/example)  and the default unhandled operation name, i left a comment in the code.
*  gRPC and https, i think with this new implementation it should be easy to work with this protocols (until now i only tested http).

@ivantopo About this:

 ```Investigate further how the Prometheus and Zipkin integrations are working with naming and metering. There might be a couple interesting details to learn from there!``` <br/>   

I haven't had time to investigate yet but I will.


Looking for a feedback about the work done and if there is something more to do in this PR 😄

instrumentation/kamon-armeria/src/main/java/kamon/armeria/instrumentation/server/ArmeriaHttpServerDecorator.java
```diff
@@ -0,0 +1,73 @@
...
+            requestHandler.buildResponse(KamonArmeriaMessageConverter.toResponse(log), requestCtxScope.context());
+            requestHandler.responseSent();
+          }
+        });
+
+    try (Storage.Scope ignored = Kamon.storeContext(scope.context())) {
+      return unwrap().serve(ctx, req);
+    }
+  }
+}
```
dpsoft, line 71:
```suggestion
@Override
  public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exception {
    final HttpServerInstrumentation.RequestHandler requestHandler = httpServerInstrumentation.createHandler(KamonArmeriaMessageConverter.toRequest(req, serverHost, httpServerInstrumentation.port()));

    ctx.setAttr(SERVER_REQUEST_HANDLER, requestHandler);

    ctx.log()
        .whenComplete()
        .thenAccept(log -> {
          final HttpServerInstrumentation.RequestHandler handler = ctx.attr(SERVER_REQUEST_HANDLER);

            if (HttpStatus.NOT_FOUND.equals(log.responseHeaders().status()) && FALLBACK_SERVICE_NAME.equals(log.serviceName())) {
              requestHandler.span().name(httpServerInstrumentation.settings().unhandledOperationName());
            }
            requestHandler.buildResponse(KamonArmeriaMessageConverter.toResponse(log), handler.context());
            requestHandler.responseSent();
        });

    try (Storage.Scope ignored = Kamon.storeContext(requestHandler.context())) {
      return unwrap().serve(ctx, req);
    }
  }
```

SimunKaracic (changes requested):
Hey, I have a couple of questions: 
1) You are storing the context in 4 different places, but never closing the returning scope. Won't that leave dirty threads around?
2) You are setting the `CLIENT_TRACE_SCOPE_KEY` in Armerias context, but not reading it from anywhere? 

Generally the code is in good shape, but I'd like to answer these before continuing with the PR.

lucasamoroso:
>     1. You are storing the context in 4 different places, but never closing the returning scope. Won't that leave dirty threads around?
 The contexts are stored inside a try-with-resources, i understand that the returned scope will be closed after the execution of the try block. 
>     2. You are setting the `CLIENT_TRACE_SCOPE_KEY` in Armerias context, but not reading it from anywhere?
I'm using this in here https://github.com/kamon-io/Kamon/pull/854/files#diff-051d01483bf35bea603907712f76837cc9fd2e519e321daa3bb86229753865dfR49 but i'm not sure it's needed. For sure we need to propagate the Kamon context before call `unwrap()`, like this:

```
   try (Storage.Scope scope = Kamon.storeContext(Kamon.currentContext())) {
      ctx.setAttr(CLIENT_TRACE_SCOPE_KEY, scope);
      return unwrap().execute(ctx, requestHandler.request());
    } catch (Exception exception) {
      requestHandler.span().fail(exception.getMessage(), exception).finish();
      throw exception;
    }
``` 
But when the request is complete:

```
    ctx.log()
        .whenComplete()
        .thenAccept(log -> {
          try (Storage.Scope ignored = Kamon.storeContext(ctx.attr(CLIENT_TRACE_SCOPE_KEY).context())) {
            Timing.takeTimings(log, requestHandler.span());
            requestHandler.processResponse(KamonArmeriaMessageConverter.toKamonResponse(log));
          }
        });
```

I'm not sure if it's necessary to run 

```
     Timing.takeTimings(log, requestHandler.span());
     requestHandler.processResponse(KamonArmeriaMessageConverter.toKamonResponse(log));
```
 with Kamon context. In the server decorator i've the same question, i hope someone with more experience with Kamon can help me.

instrumentation/kamon-armeria/src/main/java/kamon/armeria/instrumentation/server/ArmeriaHttpServerDecorator.java
```diff
@@ -0,0 +1,80 @@
...
+import kamon.Kamon;
+import kamon.armeria.instrumentation.converters.KamonArmeriaMessageConverter;
+import kamon.context.Storage;
+import kamon.instrumentation.http.HttpServerInstrumentation;
+
+import java.net.InetSocketAddress;
+
+public class ArmeriaHttpServerDecorator extends SimpleDecoratingHttpService {
+  private static final String FALLBACK_SERVICE_NAME = "com.linecorp.armeria.server.FallbackService";
+  private static final AttributeKey<Storage.Scope> SERVER_TRACE_SCOPE_KEY = AttributeKey.valueOf(Storage.Scope.class, "SERVER_TRACE_SCOPE");
```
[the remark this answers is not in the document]
lucasamoroso: Yes, that's because i'm working on something that will need this.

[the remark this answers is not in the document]
lucasamoroso: I don't like the current implementation to handle 404 for routes that doesn't exists, so i'll save the requestHandler in Armeria context and then use that requestHandler to set the unhandled operation name to the span inside requestHandler

lucasamoroso:
Hello!!
I think this pr is ready, I have been making a lot of changes in the last time because the first approach I gave to the instrumentation did not convince us much. I think this new approach is much better and it helped us add grpc and https in a very simple way, I just have a doubt regarding the propagation of the Kamon context that I left in one of the comments in response to @SimunKaracic

ivantopo:
Hey @lucasamoroso, I wanted to get into this PR and tried to rebase it to master locally but it didn't go well. There are some minor changes to tests due to the Scalatest version upgrade, but after finishing those changes the tests are not passing anymore :cry: 

Do you have time to try rebasing and pushing the updates? Otherwise let me know and I can move it forward :muscle:

lucasamoroso:
Hi @ivantopo! Sure I'll be working on this

lucasamoroso:
Hey @ivantopo! 

I did a rebase and fixed the tests. The problem with the tests was that they must extend from `InitAndStopKamonAfterAll` to start and stop Kamon.

ivantopo:
Hey @lucasamoroso, thanks for the update!

Generally speaking it looks fine and tests pass locally, but I couldn't make it work with one of my Armeria applications by publishing locally and trying it out :shrug: 

I spent some time trying to debug it and figured out that the `ArmeriaServerBuilderAdvisor` code is not really running locally (tried both using -javaagent and with the bundle). I did some more digging and realized that if I disable the bridge on `ServerBuilder` then the instrumentation gets picked up:

```
class ArmeriaHttpServerInstrumentation extends InstrumentationBuilder {
  onType("com.linecorp.armeria.server.ServerBuilder")
    .advise(method("build"), classOf[ArmeriaServerBuilderAdvisor])
//    .bridge(classOf[ServerBuilderInternalState])

  onType("com.linecorp.armeria.server.FallbackService")
    .advise(method("handleNotFound"), classOf[HandleNotFoundMethodAdvisor])
}
```

But then I had to patch the server ports for it not to blow up. My guess is that when the bridge is in place the entire ServerBuilder class gets loaded before the advise is applied. The only difference I found is that all Kamon tests are running with Java 8, but the app I'm running is using Java 17. Also tried with Java 11 and it didn't work either.

Have you tried it with newer java versions?

lucasamoroso:
Hi @ivantopo !

I've tried with all java versions but the problem was with the attach mode. [This](https://github.com/kamon-io/Kamon/pull/854/commits/edcb49d09b1d147af99f04aa9e496f98ee8b3d18) commit should fix that issue.

I'll stay tuned for comments 😄

yarosman:
So what is the status of armeria module ?

lucasamoroso:
> So what is the status of armeria module ?

Hi @yarosman!

I've resolved all @ivantopo's comments and I'm waiting for more comments or maybe this can be merged.

Are you interested in using this module?

yarosman:
> > So what is the status of armeria module ?
> 
> Hi @yarosman!
> 
> I've resolved all @ivantopo's comments and I'm waiting for more comments or maybe this can be merged.
> 
> Are you interested in using this module?

Yeap, I wanna try it

yarosman:
And I see there is [v1.15.0](https://armeria.dev/release-notes/1.15.0) available

lucasamoroso:
> And I see there is [v1.15.0](https://armeria.dev/release-notes/1.15.0) available

I've updated armeria dependency and rebase with the upstream

dpsoft (approved):
@lucasamoroso awesome!

dpsoft:
@yarosman @ivantopo is finishing some tests and if everything goes well, we will be publishing the module soon.

instrumentation/kamon-armeria/src/test/resources/application.conf
```diff
@@ -0,0 +1,4 @@
+kamon.instrumentation.armeria.server.tracing.operations.mappings = {
```
ivantopo, line 1: These settings should not be necessary after changing the way we generate server span names

ivantopo (changes requested):
I added a few comments regarding the server side of this instrumentation. Nothing major I think, just a few things to clean up and simplify. 

There is one more thing to dig into: Armeria has its own context propagation mechanisms (see the `makeContextAware` variants in Armeria's `RequestContext` for more details). Should we try to propagate Kamon's context on those places?

instrumentation/kamon-armeria/src/test/scala/kamon/instrumentation/armeria/client/ArmeriaHttpClientTracingSpec.scala
```diff
@@ -0,0 +1,231 @@
...
+      val request = HttpRequest.of(RequestHeaders.of(HttpMethod.GET, path))
+
+      val response = Kamon.runWithContext(Context.of(Span.Key, okSpan)) {
+        client.execute(request).aggregate().get()
+      }
+
+      val span = eventually(timeout(3 seconds)) {
+        val span = testSpanReporter().nextSpan().value
+
+        span.operationName shouldBe path
```
ivantopo, line 52: I was debugging this and found that the operation name is set to the request path, which can have cardinality issues. I think when we remove the operation name generator things will be better since the requests will have the same default names as all other instrumentations.

ivantopo (changes requested):

lucasamoroso:
Hi @ivantopo! 

I'll try to address your comments this week.

About

> Armeria has its own context propagation mechanisms (see the makeContextAware variants in Armeria's RequestContext for more details). Should we try to propagate Kamon's context on those places?

I think we should try to propagate Kamon's context on those places but I'm not sure how much work will be required. I'm thinking to do this in another PR (if it takes a lot of work).
 Let me dig into it after addressing the other comments

instrumentation/kamon-armeria/src/main/java/kamon/instrumentation/armeria/server/ArmeriaHttpServerDecorator.java
```diff
@@ -0,0 +1,72 @@
...
+    this.serverInstrumentationMap = serverInstrumentations;
+  }
+
+  @Override
+  public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exception {
+    final HttpServerInstrumentation httpServerInstrumentation = serverInstrumentationMap.get(req.uri().getPort());
+
+    if (httpServerInstrumentation != null) {
+
+      final HttpServerInstrumentation.RequestHandler requestHandler =
```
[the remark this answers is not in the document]
lucasamoroso: I was trying to implement this suggested change like this

```java
      ctx.log()
          .whenComplete()
          .thenAccept(log -> {
              final Context context = ctx.attr(REQUEST_HANDLER_TRACE_KEY).context();
              requestHandler.span().name(ctx.config().route().patternString());
              requestHandler.buildResponse(KamonArmeriaMessageConverter.toResponse(log),context);
              requestHandler.responseSent();
          });
```

And I've some concerns about this:
- If a route does not exist in the server we will potentially generate N operation names (1 per route that does not exist in the server), because `ctx.config().route().patternString()` will return the request path even if the route does not exist. The current implementation ensures that the operation name will be set to `unhandled`(because of [this](https://github.com/kamon-io/Kamon/pull/854/files#diff-2343eaa073a7618196380a20748e653f9ad2f420f42c35d3825f2f5838bcc01dR131)) in this case. Probably we can check if we already set this operation name to the configured value for unhandled requests. I was wondering which will be a better approach, what do you think?
- The client context, `ClientRequestContext`, does not have a method like `ctx.config().route().patternString()` so we'll need the operation name generator, right? I understand that users could have cardinality issues with the current implementation if they don't set operation mappings but I think there is nothing we can do here

yarosman:
Hello. @ivantopo you wrote that you migrate several services to armeria. Are you going to add kamon supporting ?

ivantopo:
Hey @yarosman. Yes! I'm going to make it my new year's resolution to get this PR merged and released ASAP!

We indeed have almost all of our services on Armeria now, but we did manual instrumentation on them long ago so kind of lost the urgency to do this for our own use case. But now is the time :pray: (I might have said that before, hopefully this will REALLY be the time!)

trustin:
> We indeed have almost all of our services on Armeria now

@ivantopo, how about adding you to the adopters list and get some goods, then? 😉 https://github.com/line/armeria/discussions/4626

oskoi:
@ivantopo Hi! Any ETA to merge PR?
Friendly ping

instrumentation/kamon-armeria/src/main/scala/kamon/instrumentation/armeria/KamonArmeriaHttpOperationNameGenerator.scala
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+package kamon.instrumentation.armeria
+
+import kamon.instrumentation.http.HttpMessage.Request
+import kamon.instrumentation.http.HttpOperationNameGenerator
+
+import scala.collection.concurrent.TrieMap
+
+class KamonArmeriaHttpOperationNameGenerator extends HttpOperationNameGenerator {
```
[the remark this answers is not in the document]
hughsimpson: What's the status on this?

instrumentation/kamon-armeria/src/main/java/kamon/instrumentation/armeria/server/ArmeriaHttpServerDecorator.java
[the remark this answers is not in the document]
hughsimpson: Where we at here?

instrumentation/kamon-armeria/src/main/resources/reference.conf
```diff
@@ -0,0 +1,255 @@
...
+        # The default operation name to be used when creating Spans to handle the HTTP client requests. The HTTP
+        # Client instrumentation will always try to use the HTTP Operation Name Generator configured bellow to get
+        # a name, but if it fails to generate it then this name will be used.
+        default = "http.client.request"
+
+        # FQCN for a HttpOperationNameGenerator implementation, or ony of the following shorthand forms:
+        #   - hostname: Uses the request Host as the operation name.
+        #   - method: Uses the request HTTP method as the operation name.
+        #
+        name-generator = "kamon.instrumentation.armeria.KamonArmeriaHttpOperationNameGenerator"
```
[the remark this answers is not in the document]
hughsimpson: Will it?

instrumentation/kamon-armeria/src/test/scala/kamon/instrumentation/armeria/client/ArmeriaHttpClientTracingSpec.scala
hughsimpson: Seems like a blocker

hughsimpson:
Commented some comments. Lot went on since then - if anybody has some context on those places, would appreciate it - I don't use armeria

instrumentation/kamon-armeria/src/main/java/kamon/instrumentation/armeria/server/ArmeriaHttpServerDecorator.java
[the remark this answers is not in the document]
lucasamoroso: We are using the  `patternString()` method provided by Armeria to generate operation names only in the server side, and we set the operation name as `unhandled` by default (it's configurable) if the route wasn't found.

In the client side we  still use the custom operation name generator because we don't have an equivalent `patternString()` method. at least that is true until the version of Armeria that we are currently using which is not the latest.

instrumentation/kamon-armeria/src/main/resources/reference.conf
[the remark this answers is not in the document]
lucasamoroso: Because we don't have a name generator in Armeria, we need a way to generate operation names in our side. If we could find something like `patternString()` in the client side, like we have in the server side, we could definitely remove this.

instrumentation/kamon-armeria/src/main/scala/kamon/instrumentation/armeria/KamonArmeriaHttpOperationNameGenerator.scala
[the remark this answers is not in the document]
lucasamoroso: Is the same as previous comments, we need a way to generate operation names in the client

instrumentation/kamon-armeria/src/test/scala/kamon/instrumentation/armeria/client/ArmeriaHttpClientTracingSpec.scala
lucasamoroso: Yep, I think that, if the configuration is not properly provided, we could have cardinality issues because of:
- The current implementation of `KamonArmeriaHttpOperationNameGenerator`
- The fact that we don't have a method like `patternString()` in the client side

yarosman:
...

lucasamoroso:
Hi @yarosman, I'm not sure if this PR will reach master, I'll need to revisit the implementation and also I lost track of the new Armeria features (I'm working with Go now) .
Probably I could dedicate some time to this but  I have the impression that there is no enough knowledge about Armeria internals in order to provide support and probably a better implementation and that is why I'm not sure if Kamon wants this PR. Do you need this instrumentation?

Issue #857, opened 2020-09-22
Feature: Make it possible to append custom HTTP headers in the response


pnerg:
I'd like to propose a feature to make it possible to append headers to the HTTP response.  
My particular use case is to maintain a set of legacy headers that must be returned from every application.  
Rather than having every application deal with this Kamon could append the header(s) and I can have a single library doing the job.

My thought would be to allow developers to add a custom `HttpServerResponseHeaderGenerator`via the config.  
E.g. in the config:
```
# FQCN to a class implementing the interface below
kamon.instrumentation.akka.http.server.response.header-generator = "foo.bar.ExampleHttpServerResponseHeaderGenerator"
```

The class must implement this interface for rendering custom headers
```
trait HttpServerResponseHeaderGenerator {
  def headers(context:Context):Map[String, String]
}
```
This allows custom code to extract context information or just generate whatever headers it desires
Would be invoked from the `kamon.instrumentation.http.HttpServerInstrumentation` as part of the building of the response

I'll draft a PR to go with this feature request/ticket

ivantopo:
Hey @pnerg, thanks a lot for bringing this up!

Logic is sound and the PR is actually pretty nice. I only had a small comment about where to put the settings but other than that, it looks really good. This will definitely make it to the next release :muscle:

pnerg:
Noticed the 2.1.7 release was made just a before the PR related to this ticket was merged.
Any plan when a 2.8 release will be made?
Just so I know if I should wait or make my own custom delivery from master as I need this feature.

SimunKaracic:
We planned a release for monday. 
Is that ok?

PR #858, opened 2020-09-22
Added support for adding custom HTTP headers to responses

pnerg:
This PR implements the feature request from issue: #857

In essence following the similar pattern as e.g. custom operator naming allowing the developer to provide a FQCN to a class that generates a Map of headers to add to the response.
Default implementation does nothing, i.e. empty Map

### DIFF
--- a/instrumentation/kamon-instrumentation-common/src/main/resources/reference.conf
+++ b/instrumentation/kamon-instrumentation-common/src/main/resources/reference.conf
@@ -45,7 +45,6 @@ kamon {
           enabled = yes
         }
 
-
         #
         # Configuration for HTTP request tracing.
         #
@@ -104,6 +103,12 @@ kamon {
 
             # HTTP response header name for the server span identifier, or "none" to disable it.
             span-id = none
+
+            # Controls adding custom HTTP headers to the responses
+            #   - none: Disabled, no additional headers are appended to the response
+            #   - fqcn: FQCN for a HttpServerResponseHeaderGenerator implementation
+            #
+            headers-generator = "none"
           }
 
           # Custom mappings between routes and operation names.
--- a/instrumentation/kamon-akka-http/src/main/resources/reference.conf
+++ b/instrumentation/kamon-akka-http/src/main/resources/reference.conf
@@ -51,8 +51,7 @@ kamon.instrumentation.akka.http {
       #
       #enabled = yes
     }
-
-
+  
     #
     # Configuration for HTTP request tracing.
     #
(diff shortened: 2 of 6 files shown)


SimunKaracic (approved):
I gave this a quick look, looks good to me. I'll push for it to be merged soon :+1:

instrumentation/kamon-instrumentation-common/src/main/resources/reference.conf
```diff
@@ -45,6 +45,16 @@ kamon {
...
 
+        # 
+        # Configuration for HTTP response handling
+        #
+        response {
+          # Controls adding custom HTTP headers to the responses
+          #   - none: Disabled, no additional headers are appended to the response
+          #   - fqcn: FQCN for a HttpServerResponseHeaderGenerator implementation
+          #
+          header-generator = "none"
```
ivantopo, line 56: This setting should be moved to the `response-headers` section. That section already has the settings for including the trace and span ids, so this would make a lot more sense there.

Also, just as a nitpicking, I would user `headers-generator` (plural) as the setting name.

ivantopo (changes requested):

instrumentation/kamon-instrumentation-common/src/main/resources/reference.conf
pnerg: I deliberately chose not to put it under `response-headers` due to it being under `tracing` and I didn't really think this feature was a tracing feature.
My concern is mainly with the functionality, where the config is and the name of it is all the same for me...:)
I made a suggestion, it's your source code so I'm fine to move it to different place.

pnerg: Hmm, see if I managed to do this `--amend` thingie correctly. Goal was to have a single commit to merge.
I'm no git wizard, it's all voodoo-hoodoo for me...:)

pnerg:
Seems like there's random test failures.
The `DatadogAPIReporterSpec`seems to be the one failing most of the times. Works on my laptop and I cannot see how my changes would affect a reporter

instrumentation/kamon-instrumentation-common/src/main/resources/reference.conf
ivantopo: Hey @pnerg, I think I overlooked the details here. Most likely you were right on placing the settings but now I can't go back and see how it was before! Can you please tell me where did you mean to have them in the first place? Thanks!

pnerg: that's the downside of squashing commits using `--amend`...history is gone.

Anyways I had an own section `server` in parallel with the `tracing` context like this.
instrumentation-common:
```
kamon.instrumentation.http-server.default { 
  server {
    response.headers-generator = "none"
  }
}
```
and in akka-http:
```
kamon.instrumentation.akka.http {
  server {
    #response.headers-generator = "none"
  }
}
```

Just decide where u guys want the config and I'll stuff it there...:)

ivantopo: It smells a bit bad (at least to me) that the headers generator is under tracing, even though they are not limited to include trace data in the responses. But having all the settings related to tweaking response headers in a single place seems the most ergonomic solution to users so we are going to keep it under the `tracing.response-headers` section.

If we end up introducing a couple more response-related settings we might pull them into a dedicated section, but for now I think we are fine with leaving it as it is. Thanks a lot @pnerg!

PR #859, opened 2020-09-22
Add elasticsearch instrumentation

SimunKaracic:
Add elasticsearch instrumentation for the official java High and Low level REST client (current version, 7.9). 

Simple implementation, records a span for each sync and async request with 2 tags (endpoint, method), 
and creates histogram of request sizes.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -122,6 +122,7 @@ lazy val instrumentation = (project in file("instrumentation"))
     `kamon-kafka`,
     `kamon-mongo`,
     `kamon-cassandra`,
+    `kamon-elasticsearch`,
     `kamon-annotation`,
     `kamon-annotation-api`,
     `kamon-system-metrics`,
@@ -319,6 +320,22 @@ lazy val `kamon-cassandra` = (project in file("instrumentation/kamon-cassandra")
     )
   ).dependsOn(`kamon-core`, `kamon-instrumentation-common`, `kamon-testkit` % "test", `kamon-executors`)
 
+lazy val `kamon-elasticsearch` = (project in file("instrumentation/kamon-elasticsearch"))
+  .disablePlugins(AssemblyPlugin)
+  .enablePlugins(JavaAgent)
+  .settings(instrumentationSettings)
+  .settings(
+    fork in (Test,run) := true,
+    libraryDependencies ++= Seq(
+      kanelaAgent % "provided",
+      "org.elasticsearch.client" % "elasticsearch-rest-client" % "7.9.1" % "provided",
+      "org.elasticsearch.client" % "elasticsearch-rest-high-level-client" % "7.9.1" % "provided",
+      scalatest % "test",
+      logbackClassic % "test",
+      "com.dimafeng" %% "testcontainers-scala" % "0.38.3" % "test",
+      "com.dimafeng" %% "testcontainers-scala-elasticsearch" % "0.38.3" % "test"
+    )
+  ).dependsOn(`kamon-core`, `kamon-instrumentation-common`, `kamon-testkit` % "test")
 
 
 lazy val `kamon-annotation-api` = (project in file("instrumentation/kamon-annotation-api"))
@@ -630,6 +647,7 @@ val `kamon-bundle` = (project in file("bundle/kamon-bundle"))
     `kamon-kafka` % "shaded",
     `kamon-mongo` % "shaded",
     `kamon-cassandra` % "shaded",
+    `kamon-elasticsearch` % "shaded",
     `kamon-annotation` % "shaded",
     `kamon-annotation-api` % "shaded",
     `kamon-system-metrics` % "shaded",
--- a/instrumentation/kamon-elasticsearch/src/main/java/kamon/instrumentation/elasticsearch/AsyncElasticsearchRestClientInstrumentation.java
+++ b/instrumentation/kamon-elasticsearch/src/main/java/kamon/instrumentation/elasticsearch/AsyncElasticsearchRestClientInstrumentation.java
@@ -1,0 +1,28 @@
+package kamon.instrumentation.elasticsearch;
+
+import kamon.Kamon;
+import kamon.trace.Span;
+import kanela.agent.libs.net.bytebuddy.asm.Advice;
+import org.elasticsearch.client.Request;
+import org.elasticsearch.client.ResponseListener;
+
+public class AsyncElasticsearchRestClientInstrumentation {
+
+    @Advice.OnMethodEnter
+    public static void enter(
+            @Advice.Argument(0) Request request,
+            @Advice.Argument(value = 1, readOnly = false) ResponseListener responseListener) {
+        final String operationName =
+                RequestNameConverter.convert(
+                        HighLevelElasticsearchClientInstrumentation.requestClassName.get(),
+                        "AsyncRequest");
+
+        Span span = Kamon.clientSpanBuilder(operationName, "elasticsearch.client")
+                .tag("elasticsearch.http.endpoint", request.getEndpoint())
+                .tag("elasticsearch.http.method", request.getMethod())
+                .start();
+
+        RequestSizeHistogram.record(request.getEntity());
+        responseListener = new InstrumentedListener(responseListener, span);
+    }
+}
--- a/instrumentation/kamon-elasticsearch/src/main/resources/reference.conf
+++ b/instrumentation/kamon-elasticsearch/src/main/resources/reference.conf
@@ -1,0 +1,23 @@
+# ================================================== #
+# kamon Elasticsearch client reference configuration #
+# ================================================== #
+
+kamon.instrumentation.elasticsearch {
+}
+
+kanela {
+  modules {
+    elasticsearch-driver {
+
+      name = "Elasticsearch Client"
+      description = "Provides tracing of client calls made with the official Elasticsearch Client library."
+      instrumentations = [
+        "kamon.instrumentation.elasticsearch.ESInstrumentation"
+      ]
+
+      within = [
+        "org.elasticsearch.client..*"
+      ]
+    }
+  }
+}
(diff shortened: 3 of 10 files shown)


build.sbt
```diff
@@ -319,6 +320,22 @@ lazy val `kamon-cassandra` = (project in file("instrumentation/kamon-cassandra")
...
+  .settings(
+    fork in (Test,run) := true,
+    libraryDependencies ++= Seq(
+      kanelaAgent % "provided",
+      "org.elasticsearch.client" % "elasticsearch-rest-client" % "7.9.1" % "provided",
+      "org.elasticsearch.client" % "elasticsearch-rest-high-level-client" % "7.9.1" % "provided",
+      scalatest % "test",
+      logbackClassic % "test",
+      "com.dimafeng" % "testcontainers-scala_2.12" % "0.38.3" % "test",
+      "com.dimafeng" % "testcontainers-scala-elasticsearch_2.12" % "0.38.3" % "test"
```
ivantopo, line 336: Why is the scala version included in these two artifacts? That will probably break cross testing.

instrumentation/kamon-elasticsearch/src/main/java/kamon/instrumentation/elasticsearch/highlevel/SyncClientInstrumentation.java
```diff
@@ -0,0 +1,18 @@
+package kamon.instrumentation.elasticsearch.highlevel;
+
+import kanela.agent.libs.net.bytebuddy.asm.Advice;
+
+public class SyncClientInstrumentation {
```
ivantopo, line 5: Is there any particular reason why `AsyncClientInstrumentation` and `SyncClientInstrumentation` aren't just the same advice? Also, a more specific name would help :)

instrumentation/kamon-elasticsearch/src/main/scala/kamon/instrumentation/elasticsearch/ESInstrumentation.scala
```diff
@@ -0,0 +1,41 @@
...
+  }
+}
+
+object RequestSizeHistogram {
+  private val histogram = Kamon.histogram("elastic.request.size").withoutTags()
+
+  def record(entity: HttpEntity): Unit = {
+    Option(entity)
+      .map(_.getContentLength)
+      .filter(_ >= 0)
```
ivantopo, line 38: In what cases could the content be negative or zero?

instrumentation/kamon-elasticsearch/src/main/scala/kamon/instrumentation/elasticsearch/RequestNameConverter.scala
```diff
@@ -0,0 +1,12 @@
+package kamon.instrumentation.elasticsearch
+
+object RequestNameConverter {
+  def convert(className: String, fallback: String): String = {
+    // converts class name to something more appropriate
+    // e.g. ListTasksRequest to list.tasks.requests
+    Option(className)
+      .getOrElse(fallback)
+      .replaceAll("([^_A-Z])([A-Z])", "$1.$2")
+      .toLowerCase
```
ivantopo, line 10: Few things that would be better here:
  - Add the `elasticsearch` prefix to all names.
  - Drop the `Request` at the end of the name.
  - Cache these transformed operation names. The number of possible options is bound so it wouldn't be a problem.

build.sbt
```diff
@@ -319,6 +320,22 @@ lazy val `kamon-cassandra` = (project in file("instrumentation/kamon-cassandra")
...
+    )
+  ).dependsOn(`kamon-core`, `kamon-instrumentation-common`, `kamon-testkit` % "test", `kamon-executors`)
```
ivantopo, line 338: Why the dependency on executors?

instrumentation/kamon-elasticsearch/src/main/java/kamon/instrumentation/elasticsearch/AsyncRestClientInstrumentation.java
```diff
@@ -0,0 +1,29 @@
...
+public class AsyncRestClientInstrumentation {
+
+    @Advice.OnMethodEnter
+    public static void enter(
+            @Advice.Argument(0) Request request,
+            @Advice.Argument(value = 1, readOnly = false) ResponseListener responseListener) {
+        final String operationName =
+                RequestNameConverter.convert(
+                        AsyncClientInstrumentation.requestClassName.get(),
+                        "elastic.async.request");
```
ivantopo, line 19: Please change this and all other mentions of `elastic` to `elasticsearch`

ivantopo (changes requested):

instrumentation/kamon-elasticsearch/src/main/scala/kamon/instrumentation/elasticsearch/ESInstrumentation.scala
SimunKaracic: Just being cautious. From the docs:

```
     * @return  the number of bytes of the content, or
     *          a negative number if unknown. If the content length is known
     *          but exceeds {@link java.lang.Long#MAX_VALUE Long.MAX_VALUE},
     *          a negative number is returned.
```

instrumentation/kamon-elasticsearch/src/main/java/kamon/instrumentation/elasticsearch/highlevel/SyncClientInstrumentation.java
SimunKaracic: D'oh, not reason. 
Will merge them with a prettier name.

build.sbt
SimunKaracic: Copy paste artifact, removing...

build.sbt
SimunKaracic: Yeah after some trying out, I'll include specific version for each of those libraries, depending on the version of scala used, e.g.
```
{
        if(scalaBinaryVersion.value == "2.11") {
          "com.dimafeng" % "testcontainers-scala-elasticsearch_2.11" % "0.38.3" % "test"
        } else if(scalaBinaryVersion.value == "2.12") {
          "com.dimafeng" % "testcontainers-scala-elasticsearch_2.12" % "0.38.3" % "test"
        } else {
          "com.dimafeng" % "testcontainers-scala-elasticsearch_2.13" % "0.38.3" % "test"
        }
      }
```

SimunKaracic:
`kamon-mongo` tests failing...

build.sbt
```diff
@@ -319,6 +320,38 @@ lazy val `kamon-cassandra` = (project in file("instrumentation/kamon-cassandra")
...
+    fork in (Test,run) := true,
+    libraryDependencies ++= Seq(
+      kanelaAgent % "provided",
+      "org.elasticsearch.client" % "elasticsearch-rest-client" % "7.9.1" % "provided",
+      "org.elasticsearch.client" % "elasticsearch-rest-high-level-client" % "7.9.1" % "provided",
+      scalatest % "test",
+      logbackClassic % "test",
+      {
+        if(scalaBinaryVersion.value == "2.11") {
+          "com.dimafeng" % "testcontainers-scala_2.11" % "0.38.3" % "test"
```
ElfoLiNk, line 337: Just use %% `"com.dimafeng" %% "testcontainers-scala"`

SimunKaracic: I swear I tried that and it didn't work... I must have forgotten for use `sbt reload` or something. 

Thx for the feedback :+1:

instrumentation/kamon-elasticsearch/src/main/java/kamon/instrumentation/elasticsearch/AsyncElasticsearchRestClientInstrumentation.java
```diff
@@ -15,7 +15,7 @@ public static void enter(
         final String operationName =
                 RequestNameConverter.convert(
                         HighLevelElasticsearchClientInstrumentation.requestClassName.get(),
-                        "elasticsearch.async");
+                        "AsyncRequest");
```
ivantopo, line 18: This one should be `elasticsearch/AsyncRequest`, same goes for `elasticsearch/SyncRequest`.

instrumentation/kamon-elasticsearch/src/main/resources/reference.conf
```diff
@@ -0,0 +1,23 @@
...
+    elasticsearch-driver {
+
+      name = "Elasticsearch client"
+      description = "Elasticsearch Client Instrumentation"
+      instrumentations = [
+        "kamon.instrumentation.elasticsearch.ESInstrumentation"
+      ]
+
+      within = [
+        "org.elasticsearch..*"
```
ivantopo, line 19: Can this be narrowed down?

instrumentation/kamon-elasticsearch/src/main/resources/reference.conf
```diff
@@ -0,0 +1,23 @@
...
+
+kamon.instrumentation.elasticsearch {
+}
+
+kanela {
+  modules {
+    elasticsearch-driver {
+
+      name = "Elasticsearch client"
+      description = "Elasticsearch Client Instrumentation"
```
ivantopo, line 13: Name: Elasticsearch Client
Description: Provides tracing of client calls made with the official Elasticsearch Client library.

ivantopo (changes requested):

instrumentation/kamon-elasticsearch/src/main/java/kamon/instrumentation/elasticsearch/AsyncElasticsearchRestClientInstrumentation.java
SimunKaracic: "AsyncRequest" and "SyncRequest" are fallbacks for classnames, like "ListTasksRequest" (which might be null).
So they resulting operation name is "elasticsearch/ClassName".

instrumentation/kamon-elasticsearch/src/main/resources/reference.conf
SimunKaracic: Narrowed to "org.elasticseearch.client..*"

Issue #827, opened 2020-08-06
deadlock on jvm start


bradrust:
Occasionally my instances fail to start.   They hang right after the kanela banner.  I am including a jstack output which identifies a deadlock.   

kanela-agent: 1.0.6
kamon depedencies/bundles: 2.1.4

### stdout

`OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended

 _  __                _        ______
| |/ /               | |       \ \ \ \
| ' / __ _ _ __   ___| | __ _   \ \ \ \
|  < / _` | '_ \ / _ \ |/ _` |   ) ) ) )
| . \ (_| | | | |  __/ | (_| |  / / / /
|_|\_\__,_|_| |_|\___|_|\__,_| /_/_/_/

==============================
Running with Kanela, the Kamon Instrumentation Agent :: (v1.0.6)`

### jstack output

`2020-08-06 18:24:15
Full thread dump OpenJDK 64-Bit Server VM (14.0.2+12 mixed mode, sharing):

Threads class SMR info:
_java_thread_list=0x00007f29e4002010, length=12, elements={
0x00007f2a3002e800, 0x00007f2a30ac7800, 0x00007f2a30ac9800, 0x00007f2a30acf800,
0x00007f2a30ad1800, 0x00007f2a30ad3800, 0x00007f2a30ad5800, 0x00007f2a30ad7800,
0x00007f2a30b11000, 0x00007f2a3178a800, 0x00007f2a31b37800, 0x00007f29e4001000
}

"main" #1 prio=5 os_prio=0 cpu=1637.37ms elapsed=44.96s tid=0x00007f2a3002e800 nid=0x45 waiting on condition  [0x00007f2a39920000]
   java.lang.Thread.State: WAITING (parking)
        at jdk.internal.misc.Unsafe.park(java.base@14.0.2/Native Method)
        - parking to wait for  <0x00000007db14ad28> (a java.util.concurrent.locks.ReentrantLock$NonfairSync)
        at java.util.concurrent.locks.LockSupport.park(java.base@14.0.2/LockSupport.java:211)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(java.base@14.0.2/AbstractQueuedSynchronizer.java:714)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(java.base@14.0.2/AbstractQueuedSynchronizer.java:937)
[...]

PR #862, opened 2020-09-25
Replace expiringMap with caffeine

SimunKaracic:
Replacing it because of the deadlock bug described in #827 
Not triggered in Kamon, but no reason to keep a library that we know has such a defect

### DIFF
--- a/instrumentation/kamon-annotation/src/main/java/kamon/annotation/instrumentation/cache/AnnotationCache.java
+++ b/instrumentation/kamon-annotation/src/main/java/kamon/annotation/instrumentation/cache/AnnotationCache.java
@@ -38,9 +38,9 @@ import java.util.concurrent.TimeUnit;
 
 public final class AnnotationCache {
 
-  private static ConcurrentMap<MetricKey, Object> metrics = buildCache();
+  private static Map<MetricKey, Object> metrics = buildCache();
 
-  private static ConcurrentMap<MetricKey, Object> buildCache() {
+  private static Map<MetricKey, Object> buildCache() {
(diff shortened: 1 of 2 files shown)


SimunKaracic:
`kamon-jdbc` tests failing. 
`kamon-cassandra`, `kamon-mongo` and that module are flip-flopping, and thus preventing green builds.

instrumentation/kamon-annotation/src/main/java/kamon/annotation/instrumentation/cache/AnnotationCache.java
```diff
@@ -24,29 +26,26 @@
...
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
 
 public final class AnnotationCache {
 
-  private static Map<MetricKey, Object> metrics = buildCache();
+  private static ConcurrentMap<MetricKey, Object> metrics = buildCache();
```
dpsoft, line 41: why not use the interface `Map<MetricKey, Object>` instead of the implementation `ConcurrentMap<MetricKey, Object>` in the variable declaration?

dpsoft (approved):

instrumentation/kamon-annotation/src/main/java/kamon/annotation/instrumentation/cache/AnnotationCache.java
SimunKaracic: Must have slipped through, nice catch!

PR #863, opened 2020-09-28
fix: ignore doJsonPut failure result

ryota0624:
i fixed `httpClient.doJsonPut` result `Try`'s Failure case ignored.
threw exception is catched by caller method and Failure case is output to log by this fix.


reporters/kamon-datadog/src/main/scala/kamon/datadog/DatadogSpanReporter.scala
```diff
@@ -12,6 +12,8 @@ import kamon.util.{ EnvironmentTags, Filter }
 import org.slf4j.LoggerFactory
 import play.api.libs.json.{ JsObject, Json }
 
+import scala.util.{ Failure, Success }
```
SimunKaracic, line 15: Remove unnecessary `Success` import

SimunKaracic (approved):

SimunKaracic (approved):
Thanks! :+1:

ivantopo:
Thanks a lot for the fix @ryota0624!

PR #865, opened 2020-09-30
Fix spans leaking between tests in cassandra instrumentation

SimunKaracic:
Before these fixes, failures would occur ~1/15 times.

To test this, I ran the tests 100 times without any failures.
Hopefully, that's the last of them in this module.


instrumentation/kamon-cassandra/src/test/scala/kamon/instrumentation/instrumentation/CassandraClientTracingInstrumentationSpec.scala
```diff
@@ -16,30 +16,34 @@
...
     with BeforeAndAfterAll
+    with BeforeAndAfterEach
     with MetricInspection.Syntax
     with InstrumentInspection.Syntax
     with Reconfigure
     with OptionValues
     with TestSpanReporter {
 
+  override protected def afterEach(): Unit = testSpanReporter().clear()
+  override protected def beforeEach(): Unit = testSpanReporter().clear()
```
SimunKaracic, line 45: only one of them is necessary

Issue #662, opened 2020-03-17
Hikari pool instrumentation causes NPE when connection fails, effectively hiding the real exception


eyalfa:
whenever Hikari fails to connect, we get an NPE instead of the actual exception:

> 
> Caused by: java.lang.NullPointerException: null
> 	at kamon.instrumentation.jdbc.HikariPoolCreatePoolEntryMethodAdvice$.exit(HikariInstrumentation.scala:117)
> 	at com.zaxxer.hikari.pool.HikariPool.createPoolEntry(HikariPool.java:510)
> 	at com.zaxxer.hikari.pool.HikariPool.checkFailFast(HikariPool.java:562)
> 	at com.zaxxer.hikari.pool.HikariPool.<init>(HikariPool.java:115)
> 	at com.zaxxer.hikari.HikariDataSource.<init>(HikariDataSource.java:81)
> 	at com.datorama.pluto.PlutoContext.$anonfun$getDBConnection$1(PlutoContext.scala:49)
> 	at scala.collection.mutable.HashMap.getOrElseUpdate(HashMap.scala:82)
> 	at com.datorama.pluto.PlutoContext.getDBConnection(PlutoContext.scala:38)
> 	at com.datorama.pluto.sql.Jdbc$FetchSupport.connect(Jdbc.scala:71)
> 	at com.datorama.pluto.sql.Jdbc$RsState$1.$anonfun$x$11$1(Jdbc.scala:173)
> 	at com.datorama.pluto.plan.execution.Span$.wrap(TaskExecutor.scala:338)
> 	at com.datorama.pluto.sql.Jdbc$RsState$1.measured(Jdbc.scala:166)
> 	at com.datorama.pluto.sql.Jdbc$RsState$1.liftedTree1$1(Jdbc.scala:171)
> 	at com.datorama.pluto.sql.Jdbc$RsState$1.<init>(Jdbc.scala:171)
> 	at com.datorama.pluto.sql.Jdbc$.$anonfun$streamArrowMessages$2(Jdbc.scala:266)
> 	at scala.util.Try$.apply(Try.scala:209)
> 	at com.datorama.pluto.sql.Jdbc$.$anonfun$streamArrowMessages$1(Jdbc.scala:266)
> 	at akka.stream.impl.UnfoldResourceSource$$anon$1.preStart(UnfoldResourceSource.scala:35)
> 	at akka.stream.impl.fusing.GraphInterpreter.init(GraphInterpreter.scala:306)
[...]

PR #867, opened 2020-10-02
Ensure pool entry is not null when instrumenting it

ecyshor:
If an exception is thrown in the createPoolEntry method in the HikariPool then null will be returned (https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/HikariPool.java#L507).
This causes a NPE in the instrumentation which swallows the original exception eventually.

Fixes: https://github.com/kamon-io/Kamon/issues/662


SimunKaracic:
Thanks for the contribution :tada: :+1:

PR #868, opened 2020-10-02
Assert tests throw exception

SimunKaracic:
Fix tests broken by #863


SimunKaracic:
Aaaand we're back to intermittent test failures! But kamon-datadog tests are passing again :) :tada:

PR #869, opened 2020-10-02
Make jaeger endpoint configurable

klvmungai:
Update the jaeger exporter to have a configurable endpoint. This is similar to what zipkin has.


SimunKaracic:
Relevant tests are passing, config in line with what zipkin does. 

Thanks for the contribution, merging :tada:

Issue #871, opened 2020-10-05
java.lang.UnsatisfiedLinkError: Unable to load library 'DiskArbitration'


oschrenk:
I'm trying to setup Kamon 2.1.4 following [Get Started](https://kamon.io/get-started/)

but I am getting

```
[error] Exception in thread "Host Metrics" java.lang.UnsatisfiedLinkError: Unable to load library 'DiskArbitration':
[error] dlopen(libDiskArbitration.dylib, 9): image not found
[error] dlopen(libDiskArbitration.dylib, 9): image not found
[error] Native library (darwin/libDiskArbitration.dylib) not found
...
```
( see full log [here](https://gist.github.com/oschrenk/3198beea6e30f41db76a1545b6ca295a) )

when I run `sbt run` (using `sbt-kanela-runner` `2.0.6`).

My best guess is that this is because I am using macOS 11 Big Sur Beta (work laptop is in repair, and I have to use my private machine) and that I  am hitting https://github.com/oshi/oshi/issues/1289

My current workaround (thanks @liff via gitter) is to disable host metrics and process metrics.

SimunKaracic:
Fixed with version `2.0.8`

PR #876, opened 2020-10-19
Update oshi-core to 5.2.5

SimunKaracic:
Intellij formatted the file as well, I can revert that if needed


ivantopo:
Should this fix https://github.com/kamon-io/Kamon/issues/871?

SimunKaracic:
Yes, the fix was part of a release older than this one, and i bumped us to the latest stable one.

PR #892, opened 2020-11-09
Replace play-json with nanojson, leave play-json tests for now.

DWvanGeest:
This is a follow-up from https://github.com/kamon-io/Kamon/discussions/888. The PR removes the dependency on `play-json` and instead writes JSON using https://github.com/mmastrac/nanojson, which is already required by the `kamon-status-page` module.

Since `play-json` is widely used in the Scala community, there's potential for dependency conflicts when a user pulls in both `kamon-datadog` and their own `play-json`. This PR avoids that situation.

A few questions for the maintainers:
- Should we shade the `nanojson` dep, like `kamon-status-page` does?
- I left the tests as they are, using `play-json`. Should we put in the work to port them to `nanojson`? I haven't yet for a few reasons: 
  - to make sure the JSON output was exactly the same while changing the underlying implementation, 
  - `play-json` has some niceties when working in Scala that `nanojson` does not, so replacing it there would be a bit annoying (and I'm lazy :-D )
  - `play-json` can be moved to a test dep so conflicts will no longer be an issue

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -468,8 +468,9 @@ lazy val `kamon-datadog` = (project in file("reporters/kamon-datadog"))
   .settings(
     libraryDependencies ++= Seq(
       okHttp,
-      "com.typesafe.play" %% "play-json" % "2.7.4",
+      "com.grack" % "nanojson" % "1.1",
 
+      "com.typesafe.play" %% "play-json" % "2.7.4" % "test",
       scalatest % "test",
       slf4jApi % "test",
       slf4jnop % "test",
--- a/build.sbt
+++ b/build.sbt
@@ -70,7 +70,7 @@ lazy val `kamon-status-page` = (project in file("core/kamon-status-page"))
       ShadeRule.rename("fi.iki.elonen.**"       -> "kamon.lib.@0").inAll,
     ),
     libraryDependencies ++= Seq(
-      "com.grack"     %  "nanojson"  % "1.1"   % "provided,shaded",
+      "com.grack"     %  "nanojson"  % "1.6"   % "provided,shaded",
       "org.nanohttpd" %  "nanohttpd" % "2.3.1" % "provided,shaded"
     )
   ).dependsOn(`kamon-core`)
@@ -464,11 +464,14 @@ lazy val reporters = (project in file("reporters"))
 
 
 lazy val `kamon-datadog` = (project in file("reporters/kamon-datadog"))
-  .disablePlugins(AssemblyPlugin)
+  .enablePlugins(AssemblyPlugin)
   .settings(
+    assemblyShadeRules in assembly := Seq(
+      ShadeRule.rename("com.grack.nanojson.**"  -> "kamon.lib.@0").inAll,
+    ),
     libraryDependencies ++= Seq(
       okHttp,
-      "com.grack" % "nanojson" % "1.1",
+      "com.grack" % "nanojson" % "1.6" % "provided,shaded",
 
       "com.typesafe.play" %% "play-json" % "2.7.4" % "test",
       scalatest % "test",
--- a/build.sbt
+++ b/build.sbt
@@ -464,14 +464,11 @@ lazy val reporters = (project in file("reporters"))
 
 
 lazy val `kamon-datadog` = (project in file("reporters/kamon-datadog"))
-  .enablePlugins(AssemblyPlugin)
+  .disablePlugins(AssemblyPlugin)
   .settings(
-    assemblyShadeRules in assembly := Seq(
-      ShadeRule.rename("com.grack.nanojson.**"  -> "kamon.lib.@0").inAll,
-    ),
     libraryDependencies ++= Seq(
       okHttp,
-      "com.grack" % "nanojson" % "1.6" % "provided,shaded",
+      "com.grack" % "nanojson" % "1.6",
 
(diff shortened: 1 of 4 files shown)


reporters/kamon-datadog/src/main/scala/kamon/datadog/DatadogSpanReporter.scala
```diff
@@ -101,12 +101,19 @@ class DatadogSpanReporter(@volatile private var configuration: Configuration) ex
   private val logger = LoggerFactory.getLogger(classOf[DatadogSpanReporter])
 
   override def reportSpans(spans: Seq[Span.Finished]): Unit = if (spans.nonEmpty) {
-    val spanList: List[Seq[JsObject]] = spans
+    val spanLists: Array[Array[JsonObject]] = spans
       .map(span => configuration.translator.translate(span, configuration.envTags, configuration.tagFilter).toJson())
-      .groupBy { _.\("trace_id").get.toString() }
+      .groupBy { jsonObj => Option(jsonObj.getNumber("trace_id")).get.toString() }
```
SimunKaracic, line 106: this `.get` is gonna always succeed because there's always a `trace_id`, right?

SimunKaracic (commented):
Good call on first changing only the code
It would be nice to refactor the tests too, but I'll let it slide if you insist on being lazy :D

> Should we shade the nanojson dep, like kamon-status-page does?

I think we should, but I just published this locally, and I don't see nanojson included in the jar ¯\_(ツ)_/¯


All in all, great PR, thanks for the contrib! :tada:

reporters/kamon-datadog/src/main/scala/kamon/datadog/DatadogSpanReporter.scala
DWvanGeest: Yep that's my understanding. At any rate the logic hasn't changed from what it was previously.

build.sbt
```diff
@@ -468,8 +468,9 @@ lazy val `kamon-datadog` = (project in file("reporters/kamon-datadog"))
   .settings(
     libraryDependencies ++= Seq(
       okHttp,
-      "com.typesafe.play" %% "play-json" % "2.7.4",
+      "com.grack" % "nanojson" % "1.1",
```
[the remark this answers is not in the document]
DWvanGeest: :+1: done.

SimunKaracic:
LGTM, I'll merge tomorrow when the tests finish running :D

build.sbt
```diff
@@ -464,11 +464,14 @@ lazy val reporters = (project in file("reporters"))
 
 
 lazy val `kamon-datadog` = (project in file("reporters/kamon-datadog"))
-  .disablePlugins(AssemblyPlugin)
+  .enablePlugins(AssemblyPlugin)
```
SimunKaracic, line 467: Hmm, I'm not sure if this is something we can change, we'll wait for input from @ivantopo

DWvanGeest:
Laziness aside, the two JSON libs have different requirements for the different contexts. In the actual code, we want something fast, reliable, small, and non-conflicting. In the tests, we really only care about correctness and ease of use - a difficult-to-use JSON lib will discourage test-writing or make the tests difficult to understand. So, I think there's an actual case for leaving `play-json` in that role. It's also nice to have to use a different JSON lib in tests because it ensures compatibility (i.e. it checks that nanojson isn't doing something non-compliant).

build.sbt
DWvanGeest: Yeah I wasn't sure either. But, it won't publish without it (I tried locally). The shading stuff relies on that plugin, looks like, so I follwed `kamon-status-page`'s lead.

DWvanGeest:
> LGTM, I'll merge tomorrow when the tests finish running, and someone else checks out the sbt changes

Awesome, thanks for the quick turnaround and feedback on this!

SimunKaracic:
> Laziness aside, the two JSON libs have different requirements for the different contexts. In the actual code, we want something fast, reliable, small, and non-conflicting. In the tests, we really only care about correctness and ease of use - a difficult-to-use JSON lib will discourage test-writing or make the tests difficult to understand. So, I think there's an actual case for leaving `play-json` in that role. It's also nice to have to use a different JSON lib in tests because it ensures compatibility (i.e. it checks that nanojson isn't doing something non-compliant).

Well when you put it that way, I agree that it's better to leave tests as is!

DWvanGeest:
> I think we should, but I just published this locally, and I don't see nanojson included in the jar ¯\_(ツ)_/¯

When I published `kamon-datadog` locally with the shading changes, I do see it included in the JAR in a shaded fashion (under `/kamon/lib`).

SimunKaracic:
> > I think we should, but I just published this locally, and I don't see nanojson included in the jar ¯_(ツ)_/¯
> 
> When I published `kamon-datadog` locally with the shading changes, I do see it included in the JAR in a shaded fashion (under `/kamon/lib`).

Ok, we're almost there, please just remove the shading part, it's not necessary, and we can merge :tada:

DWvanGeest:
> Ok, we're almost there, please just remove the shading part, it's not necessary, and we can merge

:+1: done!

SimunKaracic:
Done!

Oh, and for the future, you don't have to repush to rerun tests, the reviewer can (and should) do that for you

PR #898, opened 2020-11-12
Akka typed actor tag and auto-grouping changes

SimunKaracic:
This PR makes changes based on the fact that typed actors get converted to ActorAdapter class, rendering some tags and existing behaviour pointless.

# Changes 
* akka.actor.class tag is not used when the actor is typed, because all typed actor would have the same tag (`ActorAdapter`)
* operation name now only contains the message class, instead of both the message and actor class
* auto-grouping is disabled when the actor is typed, because all typed actors on the same level would be in the same group. Typed actors can still be targeted by group filters

### DIFF
--- a/instrumentation/kamon-akka/src/common/scala/kamon/instrumentation/akka/AkkaMetrics.scala
+++ b/instrumentation/kamon-akka/src/common/scala/kamon/instrumentation/akka/AkkaMetrics.scala
@@ -1,6 +1,7 @@
 package kamon.instrumentation.akka
 
 import kamon.Kamon
+import kamon.instrumentation.akka.instrumentations.ActorCellInfo
 import kamon.metric.InstrumentGroup
 import kamon.tag.TagSet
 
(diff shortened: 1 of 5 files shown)


instrumentation/kamon-akka/src/common/scala/kamon/instrumentation/akka/AkkaMetrics.scala
```diff
@@ -85,14 +87,16 @@ object AkkaMetrics {
...
-    new RouterInstruments(TagSet.builder()
+  def forRouter(path: String, system: String, dispatcher: String, routerClass: Class[_], routeeClass: String): RouterInstruments = {
+    val tags = TagSet.builder()
       .add("path", path)
       .add("system", system)
       .add("dispatcher", dispatcher)
-      .add("routerClass", routerClass)
       .add("routeeClass", routeeClass)
-      .build())
+    if (!ActorCellInfo.isTyped(routerClass)) tags.add("routerClass", routerClass.getName)
```
ivantopo, line 96: Shouldn't this be the routee rather than the router class?

ivantopo (changes requested):

instrumentation/kamon-akka/src/common/scala/kamon/instrumentation/akka/AkkaMetrics.scala
SimunKaracic: Good catch, both the router and the routee class can be "ActorAdapter". I've pushed a small fix for that

But this specific case, no, it should be router. `routerClass` is the tag I'm refactoring

PR #900, opened 2020-11-16
Bring back buffer pool metrics

### LINKED ISSUE #893: Missing 'jvm.memory.buffer-pool.usage' metric in 2.x
When we migrated Kamon from 1.x to 2.x, we observed `jvm.memory.buffer-pool.usage` metric is now missing. We were using metrics for `direct` and `mapped` buffer pools. I couldn't find it in the [documentation](https://kamon.io/docs/latest/instrumentation/system/jvm-metrics/), so I assume it is either moved in some other category or it is missing in 2.x.

Is there an other way to get values of this metric in some other way or is this missing in 2.x? And if it is missing in 2.x can it be added soon?

The Kamon dependencies we use:
```
"io.kamon" %% "kamon-bundle" % "2.1.8"
"io.kamon" %% "kamon-influxdb" % "2.1.8"
"io.kamon" %% "kamon-prometheus" % "2.1.8"
"io.kamon" % "kanela-agent" % "1.0.7"
```
Other version details:
```
scalaVersion: 2.12.8, 
sbtVersion: 1.4.2
openjdk version "11.0.2" 2019-01-15
```

SimunKaracic:
Fixes #893 by bringing back buffer pool metrics.

Reference implementation: 
https://github.com/kamon-io/kamon-system-metrics/blob/724bb63b1bf47b0a1b23d0c9820452444095a04c/src/main/scala/kamon/system/jvm/MemoryUsageMetrics.scala#L100

### DIFF
--- a/instrumentation/kamon-system-metrics/src/main/scala/kamon/instrumentation/system/jvm/JvmMetrics.scala
+++ b/instrumentation/kamon-system-metrics/src/main/scala/kamon/instrumentation/system/jvm/JvmMetrics.scala
@@ -19,7 +19,7 @@ package kamon.instrumentation.system.jvm
 import java.lang.management.ManagementFactory
 
 import kamon.Kamon
-import kamon.instrumentation.system.jvm.JvmMetrics.MemoryUsageInstruments.MemoryRegionInstruments
+import kamon.instrumentation.system.jvm.JvmMetrics.MemoryUsageInstruments.{BufferPoolInstruments, MemoryRegionInstruments}
 import kamon.instrumentation.system.jvm.JvmMetricsCollector.{Collector, MemoryPool}
 import kamon.metric.{Gauge, Histogram, InstrumentGroup, MeasurementUnit}
 import kamon.tag.TagSet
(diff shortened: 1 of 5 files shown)


instrumentation/kamon-system-metrics/src/main/scala/kamon/instrumentation/system/jvm/JvmMetrics.scala
```diff
@@ -168,6 +185,22 @@ object JvmMetrics {
           register(MemoryPoolMax, region)
         )
       })
+
+    def bufferPoolInstruments(poolName: String): BufferPoolInstruments = {
+      _memoryBuffersCache.getOrElseUpdate(poolName, {
+        val commonTags = tags
+          .withTag("pool", poolName)
+        val usageTags = commonTags.withTag("measure", "used")
+        val capacityTags = commonTags.withTag("measure", "capacity")
```
ivantopo, line 194: Do not include these tags. The `used` or `capacity` details are already defined in the metric name.

instrumentation/kamon-system-metrics/src/main/scala/kamon/instrumentation/system/jvm/JvmMetrics.scala
```diff
@@ -124,6 +124,22 @@ object JvmMetrics {
     description = "Total number od classes currently loaded"
   )
 
+  val BufferPoolCount = Kamon.gauge(
+    name = s"jvm.memory.buffer-pool.count",
+    description = "Estimated number of buffers in the pool"
+  )
+  val BufferPoolUsage = Kamon.gauge(
+    name = s"jvm.memory.buffer-pool.usage",
```
ivantopo, line 132: This should be changed to `used` to be in line with other similar metrics. I would also recommend add a proper `usage` metric that tracks the percentage of usage (capacity vs actual use).

ivantopo (changes requested):

instrumentation/kamon-system-metrics/src/main/scala/kamon/instrumentation/system/jvm/JvmMetrics.scala
SimunKaracic: Renamed. 

Making a `usage` metric doesn't make sense, since the `BufferPoolMXBean` can only tell us the total capacity of the pool, and the amount currently used by the JVM.

Docs for `getMemoryUsed`: 
```
Returns an estimate of the memory that the Java virtual machine is using for this buffer pool. 
The value returned by this method may differ from the estimate of the total capacity of the buffers in this pool. 
This difference is explained by alignment, memory allocator, and other implementation specific reasons.
```

PR #926, opened 2021-01-20
Allow to define custom router name generator for play server

yarosman:
Current operation name are transformed play.api.routing.HandlerDef.path to predefined format and you can't change it
```
/foo/user/$user<regexp>/blah to /foo/user/user/blah
```

And now you can define custom operation name generator with format like in the micrometer, and you will have the same template format between java/scala services
```
/foo/user/$user<regexp>/blah to /foo/user/{user}/blah instead of /foo/user/user/blah
```


SimunKaracic:
Hello, please add a small writeup of what this PR is trying to achieve, and I'll get to checking out the PR soon!

yarosman:
> Hello, please add a small writeup of what this PR is trying to achieve, and I'll get to checking out the PR soon!

I added description )

yarosman:
@SimunKaracic did you have time to review this pr ?

SimunKaracic:
Not yet, we'll get to it this week!

instrumentation/kamon-play/src/main/resources/reference.conf
```diff
@@ -139,6 +139,10 @@ kamon.instrumentation.play.http {
         }
       }
     }
+
+    name-generator = "kamon.instrumentation.play.DefaultRouterOperationNameGenerator"
+
+    grpc-name-generator = "kamon.instrumentation.play.DefaultGrpcRouterNameGenerator"
```
ivantopo, line 145: Hey @yarosman!

Generally speaking I think the PR is great and I'm sorry it took so long to write down this feedback. All this time I have been concerned that if we add these two new settings then users will have three different name generators:
  - `operations.name-generator` coming from instrumentation-common (I notice that it was not included in the reference conf for Play but you can see it here [here](https://github.com/kamon-io/Kamon/blob/master/instrumentation/kamon-instrumentation-common/src/main/resources/reference.conf#L128-L132)
  - `name-generator` introduced by this PR, which should work right after the one above and will override any name given by settings in instrumentation-common.
  - `grpc-name-generator` which I think (not 100% sure) ends up overriding any work that might have been done by both of the above.

I think we should have a single way to provide these name generators, and I think there should have been a framework-specific way to generate those names instead of having it all abstracted up in instrumentation-common. Still, this is a nice improvement for Play users! 

Here is what I think we can do:
- Move these settings into `kamon.instrumentation.play.http.server.extra` so that we group together anything "extra" that is play-specific in there.
- Add a comment to the settings explaining what do they do and mentioning the interfaces that should be implemented on each case. Similar to the name-generator comment [on instrumentation-common](https://github.com/kamon-io/Kamon/blob/master/instrumentation/kamon-instrumentation-common/src/main/resources/reference.conf#L128-L132). Also please mention that these interfaces are internal APIs that might change in the future.

That should be it!

ivantopo (changes requested):

instrumentation/kamon-play/src/main/resources/reference.conf
yarosman: Hello, @ivantopo !
I moved configuration to `extra` block and comments was added about using play-internal api

ivantopo (approved):
Thanks a lot @yarosman!

PR #928, opened 2021-01-27
Deflate compression

agemooij:


agemooij:
Apologies. Preparing to create this PR but pushed save too early

PR #929, opened 2021-01-27
Datadog 'Deflate' compression

arjun-1:
This PR introduces 'Deflate' compression when `POST`ing to the Datadog API. The Datadog API is known to have a payload limit (see https://docs.datadoghq.com/api/latest/metrics/#submit-metrics). By using compression, users can `POST` bodies which are larger in size than this limit.

Compression has been made optional, under the `datadog.api.compression` key.

We ran into the Datadog limit when posting metrics for one of our services. This was resolved by using this 'Deflate' compression functionality. Note that the Datadog API doesn't support gzip compression, as we attempted this compression method before.


SimunKaracic:
Great PR, merge-ing as is!

PR #934, opened 2021-02-01
Akka http upgrade

### LINKED ISSUE #810: Hostname operation name generator produces empty names
in our setups, the following config results in empty names and always falls back to the default operations name of `http.client.request`

```
kamon.instrumentation.akka.http.client.tracing.operations.name-generator = "hostname"
```
We have to supply custom generators that consider the requests `Host` header as well like

```
class ClientRequestOperationNameGenerator extends HttpOperationNameGenerator {
  override def name(request: HttpMessage.Request): Option[String] =
    Some(request.host).filter(_.nonEmpty) orElse request.read(Host.name)
}
```

Is the current behaviour intended or would you be open to change/accept a PR for this?

SimunKaracic:
Update akka-http to `10.2.3`, fix #810 on the way


instrumentation/kamon-akka-http/src/main/scala/kamon/instrumentation/akka/http/AkkaHttpClientInstrumentation.scala
```diff
@@ -33,7 +33,10 @@ class AkkaHttpClientInstrumentation extends InstrumentationBuilder {
     * Simply modifies the requests as they are submitted. This does not cover connection pooling, just requests sent
     * via the Http.singleRequest mechanism.
     */
+
   onType("akka.http.scaladsl.HttpExt")
+    .advise(method("singleRequest"), classOf[HttpExtSingleRequestAdvice])
+    // akka-http 10.1.x compatibility
```
ivantopo, line 39: I'm not totally sure about this. If `singleRequest` ends up calling `singleRequestImpl` in 10.1.x, wouldn't this be causing the instrumentation create two spans for all client requests?

ivantopo (changes requested):

ivantopo (approved):

Issue #913, opened 2020-12-05
Warning immensely helpful when operating on finished spans


vishramachandran:
It helps when a warning log or exception is printed when operating on finished spans.

In asynchronous workflows, a span that is finished may be propagated down unintentionally. In other cases, `Kamon.runWithSpan(span) { ...}` may be used with default `finish=true` param, and asynchronous flow could be started from inside. Any operations on such finished span are **silently** lost. Finding this is hard enough. Debugging how the missing information in the span came about is another painful exercise.

A simple warning message that "you are trying to update a finished span" is sufficient to help debug such situations. Right now, nothing is emitted in logs.

SimunKaracic:
A warning there would be helpful, and it shouldn't be too hard to implement!

Feel free to contribute that, if you can't we'll get to it soon

PR #936, opened 2021-02-02
Warn on finished span in runWithSpan

SimunKaracic:


ivantopo:
Given that this issue is related to the comments on https://github.com/kamon-io/Kamon/issues/913, I wouldn't limit the warnings to `runWithSpan`. There are many ways we could be interacting with finished Spans, so I would rather put these things on the tag methods on the Span class itself.

And this opens two questions:
  - What do we do with Empty and Remote spans. I'm guessing they should also get warning.
  - Should the warning be there all the time or should we have a specific flag for it? I'm inclined to believe this should be behind some sort of debug flag.

SimunKaracic:
Added flag, and a bunch of log messages everywhere. 
Not sure if this is how you wanted it to look @ivantopo , so please check it out and let me know if you like it

PR #938, opened 2021-02-08
Add configurable endpoint for Prometheus scraping

SimunKaracic:
Introduces new setting to keep backwards compatibility for kamon-prometheus endpoint

### DIFF
--- a/reporters/kamon-prometheus/src/main/resources/reference.conf
+++ b/reporters/kamon-prometheus/src/main/resources/reference.conf
@@ -82,7 +82,7 @@ kamon.prometheus {
     # Hostname and port used by the embedded web server to publish the scraping enpoint.
     hostname = "0.0.0.0"
     port = 9095
-    metrics_path = "/metrics"
+    metrics-path = "/metrics"
 
     # embedded server impl
     # choose between
--- a/reporters/kamon-prometheus/src/main/resources/reference.conf
+++ b/reporters/kamon-prometheus/src/main/resources/reference.conf
@@ -82,7 +82,7 @@ kamon.prometheus {
     # Hostname and port used by the embedded web server to publish the scraping enpoint.
     hostname = "0.0.0.0"
     port = 9095
-    path = "/metrics"
+    metrics_path = "/metrics"
 
     # embedded server impl
     # choose between
--- a/reporters/kamon-prometheus/src/main/resources/reference.conf
+++ b/reporters/kamon-prometheus/src/main/resources/reference.conf
@@ -82,6 +82,7 @@ kamon.prometheus {
     # Hostname and port used by the embedded web server to publish the scraping enpoint.
     hostname = "0.0.0.0"
     port = 9095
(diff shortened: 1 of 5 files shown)


reporters/kamon-prometheus/src/main/resources/reference.conf
```diff
@@ -82,6 +82,7 @@ kamon.prometheus {
     # Hostname and port used by the embedded web server to publish the scraping enpoint.
     hostname = "0.0.0.0"
     port = 9095
+    path = "/metrics"
```
Philippus, line 85: Maybe use `metrics-path` so it goes nicely with https://prometheus.io/docs/prometheus/latest/configuration/configuration/

reporters/kamon-prometheus/src/main/scala/kamon/prometheus/embeddedhttp/EmbeddedHttpServer.scala
```diff
@@ -19,6 +19,6 @@ package kamon.prometheus.embeddedhttp
 import com.typesafe.config.Config
 import kamon.prometheus.ScrapeSource
 
-abstract class EmbeddedHttpServer(hostname: String, port: Int, scrapeSource: ScrapeSource, config: Config) {
+abstract class EmbeddedHttpServer(hostname: String, port: Int, path: String, scrapeSource: ScrapeSource, config: Config) {
```
Philippus, line 22: will this not break backwards compatibility for users that already defined their own `EmbeddedHttpServer` ?

SimunKaracic: I was not aware that people extended that! Fixed!

reporters/kamon-prometheus/src/main/resources/reference.conf
SimunKaracic: Nice catch, will do!

Philippus: `metrics-path` is more encouraged HOCON style I think. Sorry to bother you. :)

SimunKaracic: Don't apologize, feedback is always welcome.

But I think I'm gonna stick with `_` on this one, Prometheus configuration uses metrics_path, so it makes sense if we're going for compatibility

reporters/kamon-prometheus/src/main/resources/reference.conf
```diff
@@ -82,6 +82,7 @@ kamon.prometheus {
     # Hostname and port used by the embedded web server to publish the scraping enpoint.
     hostname = "0.0.0.0"
     port = 9095
+    metrics_path = "/metrics"
```
ivantopo, line 85: Please call it `metrics-path`, we don't user underscore on any of our setting names.

ivantopo (changes requested):

PR #947, opened 2021-02-16
Kafka Consumer instrumentation issue

mladens:
Handle case where consumer has no group.id set.
In one particular case, `globalConsumer` used by  KafkaStreams when restoring global state.


SimunKaracic (approved):
LGTM! :rocket:

PR #948, opened 2021-02-18
Basic spring boot support

SimunKaracic:
Add basic spring-boot support for servers using spring-mvc.

If you see a comment, it's either a reminder for me or a comment for you

I've tested this with multiple, albeit simple apps. 
It worked with mvn and gradle packaging :tada:

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -348,6 +348,7 @@ lazy val `kamon-spring` = (project in file("instrumentation/kamon-spring"))
       // check that dependencies are shaded
       kanelaAgent % "provided",
       "org.springframework.boot" % "spring-boot-starter-web" % "2.4.2" % "provided",
+      "org.springframework.boot" % "spring-boot-starter-webflux" % "2.4.2" % "provided",
 
       okHttp % "test",
       "com.h2database" % "h2" % "1.4.200" % "test",
--- a/build.sbt
+++ b/build.sbt
@@ -123,6 +123,7 @@ lazy val instrumentation = (project in file("instrumentation"))
     `kamon-mongo`,
     `kamon-cassandra`,
     `kamon-elasticsearch`,
+    `kamon-spring`,
     `kamon-annotation`,
     `kamon-annotation-api`,
     `kamon-system-metrics`,
@@ -334,6 +335,25 @@ lazy val `kamon-elasticsearch` = (project in file("instrumentation/kamon-elastic
       logbackClassic % "test",
       "com.dimafeng" %% "testcontainers-scala" % "0.38.3" % "test",
       "com.dimafeng" %% "testcontainers-scala-elasticsearch" % "0.38.3" % "test"
+    )
+  ).dependsOn(`kamon-core`, `kamon-instrumentation-common`, `kamon-testkit` % "test")
+
+lazy val `kamon-spring` = (project in file("instrumentation/kamon-spring"))
+  .disablePlugins(AssemblyPlugin)
+  .enablePlugins(JavaAgent)
+  .settings(instrumentationSettings)
+  .settings(
+    fork in (Test,run) := true,
+    libraryDependencies ++= Seq(
+      // check that dependencies are shaded
+      kanelaAgent % "provided",
+      "org.springframework.boot" % "spring-boot-starter-web" % "2.4.2" % "provided",
+
+      okHttp % "test",
+      "com.h2database" % "h2" % "1.4.200" % "test",
+      "org.springframework.boot" % "spring-boot-starter-data-jpa" % "2.4.2" % "test",
+      scalatest % "test",
+      logbackClassic % "test",
     )
   ).dependsOn(`kamon-core`, `kamon-instrumentation-common`, `kamon-testkit` % "test")
 
@@ -656,6 +676,7 @@ val `kamon-bundle` = (project in file("bundle/kamon-bundle"))
     `kamon-mongo` % "shaded",
     `kamon-cassandra` % "shaded",
     `kamon-elasticsearch` % "shaded",
+    `kamon-spring` % "shaded",
     `kamon-annotation` % "shaded",
     `kamon-annotation-api` % "shaded",
     `kamon-system-metrics` % "shaded",
--- a/build.sbt
+++ b/build.sbt
@@ -333,8 +333,8 @@ lazy val `kamon-elasticsearch` = (project in file("instrumentation/kamon-elastic
       "org.elasticsearch.client" % "elasticsearch-rest-high-level-client" % "7.9.1" % "provided",
       scalatest % "test",
       logbackClassic % "test",
-      "com.dimafeng" %% "testcontainers-scala" % "0.38.3" % "test",
-      "com.dimafeng" %% "testcontainers-scala-elasticsearch" % "0.38.3" % "test"
+      "com.dimafeng" %% "testcontainers-scala" % "0.39.1" % "test",
+      "com.dimafeng" %% "testcontainers-scala-elasticsearch" % "0.39.1" % "test"
     )
   ).dependsOn(`kamon-core`, `kamon-instrumentation-common`, `kamon-testkit` % "test")
 
--- a/instrumentation/kamon-spring/src/main/java/kamon/instrumentation/spring/server/CallableWrapper.java
+++ b/instrumentation/kamon-spring/src/main/java/kamon/instrumentation/spring/server/CallableWrapper.java
@@ -1,0 +1,13 @@
+package kamon.instrumentation.spring.server;
+
+import kanela.agent.libs.net.bytebuddy.asm.Advice;
+import java.util.concurrent.Callable;
+
+public class CallableWrapper {
+    @Advice.OnMethodEnter()
+    // raw use of parametrized class?
+    public static void enter(@Advice.Argument(value = 0, readOnly = false) Callable callable) {
+        callable = CallableContextWrapper.wrap(callable);
+    }
+
+}
--- a/instrumentation/kamon-spring/src/main/java/kamon/instrumentation/spring/server/CallableWrapper.java
+++ b/instrumentation/kamon-spring/src/main/java/kamon/instrumentation/spring/server/CallableWrapper.java
@@ -1,6 +1,7 @@
 package kamon.instrumentation.spring.server;
 
 import kanela.agent.libs.net.bytebuddy.asm.Advice;
+
 import java.util.concurrent.Callable;
 
 public class CallableWrapper {
--- a/instrumentation/kamon-spring/src/main/java/kamon/instrumentation/spring/server/CallableWrapper.java
+++ b/instrumentation/kamon-spring/src/main/java/kamon/instrumentation/spring/server/CallableWrapper.java
@@ -5,8 +5,7 @@ import java.util.concurrent.Callable;
 
 public class CallableWrapper {
     @Advice.OnMethodEnter()
-    // raw use of parametrized class?
-    public static void enter(@Advice.Argument(value = 0, readOnly = false) Callable callable) {
+    public static void enter(@Advice.Argument(value = 0, readOnly = false) Callable<?> callable) {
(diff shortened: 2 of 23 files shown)


build.sbt
```diff
@@ -348,6 +348,7 @@ lazy val `kamon-spring` = (project in file("instrumentation/kamon-spring"))
       // check that dependencies are shaded
       kanelaAgent % "provided",
       "org.springframework.boot" % "spring-boot-starter-web" % "2.4.2" % "provided",
```
ivantopo, line 350: @SimunKaracic do you know down to what version are we backwards compatible?

instrumentation/kamon-spring/src/main/scala/kamon/instrumentation/spring/client/ClientInstrumentation.scala
```diff
@@ -0,0 +1,106 @@
...
+import org.springframework.web.reactive.function.client.{ClientRequest, ClientResponse}
+import reactor.core.publisher.Mono
+
+import java.util.function.Consumer
+import scala.collection.JavaConverters.mapAsScalaMapConverter
+import scala.collection.mutable
+
+object ClientInstrumentation {
+  private val instrumentation: HttpClientInstrumentation =
+    HttpClientInstrumentation.from(ConfigFactory.empty(), "spring.client")
```
ivantopo, line 16: Why an empty config here? I think we should do something similar to what we do with all other instrumentations: have a dedicated path in the config for this specific implementation and fallback to the defaults.

instrumentation/kamon-spring/src/main/scala/kamon/instrumentation/spring/client/ClientInstrumentation.scala
```diff
@@ -0,0 +1,106 @@
...
+  def wrapResponse(mono: Mono[ClientResponse], handler: RequestHandler[ClientRequest]): Mono[ClientResponse] = {
+    // This is unnecessarily complicated because scala 2.11 support
+    mono.doOnSuccess {
+      new Consumer[ClientResponse] {
+        override def accept(t: ClientResponse): Unit = {
+          handler.processResponse(new HttpMessage.Response {
+            override def statusCode: Int = t.rawStatusCode()
+          })
+        }
+      }
```
ivantopo, line 93: Why do we have this in here? It looks a little bit weird, almost as if it was test code that was left over, or like it is missing to call finish on the Span.

instrumentation/kamon-spring/src/main/scala/kamon/instrumentation/spring/client/SpringWebClientInstrumentation.scala
```diff
@@ -0,0 +1,11 @@
+package kamon.instrumentation.spring.client
+
+import kanela.agent.api.instrumentation.InstrumentationBuilder
+
+class SpringWebClientInstrumentation extends InstrumentationBuilder {
+  /*
+   * Instruments spring WebClient, mutating the request and registering callbacks.
+   */
+  onSubTypesOf("org.springframework.web.reactive.function.client.ExchangeFunction")
```
ivantopo, line 9: Are there many `ExchangeFunction` implementations out there? :thinking:

instrumentation/kamon-spring/src/main/java/kamon/instrumentation/spring/server/CallableWrapper.java
```diff
@@ -0,0 +1,13 @@
+package kamon.instrumentation.spring.server;
+
+import kanela.agent.libs.net.bytebuddy.asm.Advice;
+import java.util.concurrent.Callable;
+
+public class CallableWrapper {
+    @Advice.OnMethodEnter()
+    // raw use of parametrized class?
```
ivantopo, line 8: I bet you can use `Callable<?>` and it will work out?

instrumentation/kamon-spring/src/main/resources/reference.conf
```diff
@@ -0,0 +1,25 @@
+# ================================================== #
+# kamon Spring reference configuration #
+# ================================================== #
+
+kamon.instrumentation.spring {
+   server {
+   }
```
ivantopo, line 7: This is missing all the settings for the client and server instrumentation. The same as we do for Akka HTTP.

instrumentation/kamon-spring/src/main/resources/reference.conf
```diff
@@ -0,0 +1,25 @@
...
+    spring {
+      name = "Spring instrumentation"
+      description = "Provides tracing and context propagation for server requests in Spring MVC"
+      instrumentations = [
+        "kamon.instrumentation.spring.server.SpringMVCInstrumentation",
+        "kamon.instrumentation.spring.client.SpringWebClientInstrumentation"
+      ]
+
+      within = [
+        "org.springframework.web..*",
```
ivantopo, line 21: If possible, narrow this down to more specific packages.

instrumentation/kamon-spring/src/main/scala/kamon/instrumentation/spring/server/HasServerInstrumentation.scala
```diff
@@ -0,0 +1,28 @@
...
+}
+
+object HasServerInstrumentation {
+
+  class Mixin(var serverInstrumentation: HttpServerInstrumentation) extends HasServerInstrumentation {
+    override def getServerInstrumentation(request: HttpServletRequest): HttpServerInstrumentation = {
+      if (serverInstrumentation == null) {
+        // should I just create an empty config since, you know
+        // there's no settings?
+        val config = Kamon.config().getConfig("kamon.instrumentation.spring.server")
```
ivantopo, line 19: To answer the question: no. There is a comment on the `reference.conf` file about it. Just follow the same practice we have on the Akka HTTP instrumentation.

instrumentation/kamon-spring/src/main/scala/kamon/instrumentation/spring/server/SpringMVCInstrumentation.scala
```diff
@@ -0,0 +1,95 @@
...
+
+object RenderAdvice {
+  @Advice.OnMethodEnter()
+  def enter(): Span =
+    Kamon.serverSpanBuilder("view.render", "spring.server").start()
+
+  @Advice.OnMethodExit()
+  def exit(@Advice.Enter span: Span): Unit =
+    span.finish()
+}
```
ivantopo, line 54: Will this Span be a child of the actual request Span? I'm almost sure it should be, and that it shouldn't be marked as server. Probably should be an internal Span.

instrumentation/kamon-spring/src/main/scala/kamon/instrumentation/spring/server/SpringMVCInstrumentation.scala
```diff
@@ -0,0 +1,95 @@
...
+    .advise(
+      method("startCallableProcessing")
+        // is this [_] ok?
+        .and(withArgument(0, classOf[Callable[_]])), classOf[CallableWrapper])
+}
+
+object ProcessHandlerAdvice {
+  @Advice.OnMethodExit(onThrowable = classOf[Throwable], suppress = classOf[Throwable])
+  def exit(@Advice.Argument(3) throwable: Throwable): Unit =
+    Option(throwable).map(Kamon.currentSpan().fail)
```
ivantopo, line 42: Are you sure that when a Span fails it will still be finished somewhere else in the instrumentation?

instrumentation/kamon-spring/src/main/scala/kamon/instrumentation/spring/server/SpringMVCInstrumentation.scala
```diff
@@ -0,0 +1,95 @@
...
+    val requestHandler = Option(request.getAttribute("kamon-handler"))
+      .getOrElse({
+        val serverInstrumentation = dispatcherServlet.getServerInstrumentation(request)
+        val handler = serverInstrumentation.createHandler(InstrumentationUtils.requestReader(request))
+        request.setAttribute("kamon-handler", handler)
+        handler
+      }).asInstanceOf[RequestHandler] // cast because .getAttribute returns AnyRef
+    val scope = Kamon.storeContext(requestHandler.context)
+
+    // should I be tupling this like this?
```
ivantopo, line 70: I remember discussing it with @dpsoft long time ago and the conclusion after lots of testing is that it is fine to do it and we would only result to using `@Advice.Local` in ultra performance sensitive situations.

instrumentation/kamon-spring/src/main/scala/kamon/instrumentation/spring/server/SpringMVCInstrumentation.scala
```diff
@@ -0,0 +1,95 @@
...
+  def exit(@Advice.Enter enter: (RequestHandler, Scope),
+           @Advice.Argument(0) request: HttpServletRequest,
+           @Advice.Argument(1) response: HttpServletResponse): Unit = {
+    val (handler, scope) = enter
+
+    // Extract name from path if it matches a pattern
+    // otherwise, use `http.server.request`
+    Option(request.getAttribute(HandlerMapping.BEST_MATCHING_PATTERN_ATTRIBUTE))
+      .map(_.toString)
+      .foreach(handler.span.name)
```
ivantopo, line 84: Can the operation be named on the enter advice? I'm not sure if the adaptive sampler will be able to get a decent operation name.

ivantopo (changes requested):

build.sbt
SimunKaracic: No idea, but I'll be decrementing and seeing where we fail

instrumentation/kamon-spring/src/main/scala/kamon/instrumentation/spring/server/SpringMVCInstrumentation.scala
SimunKaracic: Yes it will be a child. Changing to internal span :eyes:

instrumentation/kamon-spring/src/main/scala/kamon/instrumentation/spring/server/SpringMVCInstrumentation.scala
SimunKaracic: Yes!

dpsoft: I think that is good idea, when were are in the hot path, make an explicit null check instead of using an Option in order to avoid an unnecessary allocation. :)

instrumentation/kamon-spring/src/main/scala/kamon/instrumentation/spring/server/SpringMVCInstrumentation.scala
SimunKaracic: It cannot be named on the onEnter advice since the pattern is not available there.
The earliest that I found it was in `applyPreHandle`, so now we delay the sampling decision until we exit that method, and then name the span and take a sampling decision

instrumentation/kamon-spring/src/main/java/kamon/instrumentation/spring/server/CallableWrapper.java
SimunKaracic: Worked!

instrumentation/kamon-spring/src/main/scala/kamon/instrumentation/spring/client/SpringWebClientInstrumentation.scala
SimunKaracic: Exactly one, so I'll narrow it down

build.sbt
SimunKaracic: We work up to `2.2.0` for sure! 
We probably work for older versions than that, but the examples start breaking when we go further back.

PR #953, opened 2021-03-07
Allow specification of excluded endpoints for tracing+metrics

hughsimpson:
Being able to exclude specific endpoints from tracing+metrics tracking can reduce noise (e.g. I would like to be able to exclude health & metrics endpoints in my apps). Might be desirable to permit distinguishing 'exclude from metrics' from 'exclude from tracing' but for now I've assumed that they would be similar concerns, and aggregated them into a single 'excludes' list

### DIFF
--- a/instrumentation/kamon-instrumentation-common/src/main/scala/kamon/instrumentation/http/HttpServerInstrumentation.scala
+++ b/instrumentation/kamon-instrumentation-common/src/main/scala/kamon/instrumentation/http/HttpServerInstrumentation.scala
@@ -32,6 +32,7 @@ import kamon.trace.Trace.SamplingDecision
 import kamon.util.Filter
 import org.slf4j.LoggerFactory
 
+import scala.collection.JavaConverters._
 import scala.util.Try
 
 
@@ -218,12 +219,12 @@ object HttpServerInstrumentation {
       }
 
     override def createHandler(request: HttpMessage.Request, deferSamplingDecision: Boolean): RequestHandler = {
+      val excluded = settings.excludes(request)
 
-      val incomingContext = if(settings.enableContextPropagation)
+      val incomingContext = if(settings.enableContextPropagation && !excluded)
         _propagation.read(request)
       else Context.Empty
-
-      val requestSpan = if(settings.enableTracing)
+      val requestSpan = if(settings.enableTracing && !excluded)
         buildServerSpan(incomingContext, request, deferSamplingDecision)
       else Span.Empty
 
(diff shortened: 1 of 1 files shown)


hughsimpson:
Failing tests appear to be entirely unrelated. See other pr for some fixes

hughsimpson:
On reflection, splitting the two is probably better

SimunKaracic:
This looks interesting, but it's touching some core stuff, so we'll need to take a closer look at it. 
In the meantime, can you add one test that verifies this solution?

hughsimpson:
Sure. This one was a little tentative anyway, will update pr tomorrow

pnerg:
I achieved the same/similar behaviour by overriding the sampling decision.
This way I could get rid of useless incoming requests (e.g. health) as well as outgoing requests such as lookups in Consul.

**Custom sampler:**
In my case I wrap the `RandomSampler` by proxying it with a filter.  
If we pass my filter then we use the `RandomSampler`. As it was how I wanted it to work.
```
/**
  * Custom decision point if the span for a particular URI/path is to be sampled
  */
class PathFilterSampler extends Sampler with PathFilter {
  
  // uses the random sampler in case we pass the URI filtering
  // configured using the same means as if the sampler would be created by Kamon itself
  // https://github.com/kamon-io/Kamon/blob/master/kamon-core/src/main/scala/kamon/trace/Tracer.scala#L363
  private val sampler = RandomSampler(Kamon.config().getDouble("kamon.trace.random-sampler.probability"))

  override def decide(operation: Sampler.Operation): Trace.SamplingDecision = {
    if(accept(operation.operationName())) sampler.decide(operation) else Trace.SamplingDecision.DoNotSample
  }
}
```
**Path filter utility**
```
object PathFilter {
  /** 
    * Singleton of the filter to be used
    */
  private val filter:Option[Filter] = Try(Kamon.filter("kamon.filters.path-filter")).toOption
}

trait PathFilter {
  import PathFilter._

  /**
    * Filter the provided path/URI.
    * If no filter is configured the response defaults to ''true''
    */
  def accept(path:String):Boolean = filter.map(_.accept(path)).getOrElse(true)
}
```

**Custom config for the filter:**
```
kamon.filters {
  "path-filter" {
    includes = ["**"]
    #exclude calls to Consul and all non-functional paths from the span measurements and metric generation
    excludes = ["regex:.*/v1/.*", "regex:.*/app-details", "regex:.*/health", "regex:.*/health/.*-status"]
  }
}
```
**Override the sampler**:
```
kamon.trace.sampler = "foo.bar.kamon.PathFilterSampler"
```

PR #954, opened 2021-03-07
bring tracing support for Cassandra 4.x driver, fixes #798

### LINKED ISSUE #798: Instrument the Cassandra Driver 4.x series
We are about to release support for the 3.x Cassandra driver series via https://github.com/kamon-io/Kamon/pull/763, but 4.x has been out for long already!

A good thing about the new driver (and partly, the old one) is that the connection pool implementation and default settings are suitable for most scenarios. To me it seems like the most important feature to bring to the 4.x driver is request tracing and general session-level metrics. It might be totally fine to ignore the connection pool metrics for a first release.

ivantopo:
This PR brings query tracing for the Cassandra Driver 4.x. A few notes on differences to the 3.x support:
  - There are no connection pool metrics, nor session utilization metrics.
  - There is no detailed tracing of executions, only high-level queries are traced.

From the experience we had with the 3.x driver I would say that we should forget about trying to add connection pool metrics, but we should add basic monitoring of the Throttler implementations and optional executions tracing. Those two don't seem too hard to achieve, but I will not include them on this PR.


instrumentation/kamon-cassandra/build.sbt
```diff
@@ -0,0 +1,30 @@
...
+)
+
+configs(Cassandra3xTest, Cassandra4xTest)
+inConfig(Cassandra3xTest)(Defaults.testSettings)
+inConfig(Cassandra4xTest)(Defaults.testSettings)
+
+test in Test :=  {
+  (test in Cassandra3xTest).value
+  (test in Cassandra4xTest).value
+}
```
SimunKaracic, line 30: Make your editor add newline on save!

PR #961, opened 2021-03-20
Add Storage.CrossThreadLocal

jatcwang:
This storage implements the "unoptimized" version of thread local storage which allows for the scope to be closed in another thread, as it re-resolves the thread local context in `close`

Rerunning the benchmark:

```
[info] # JMH version: 1.21
[info] # VM version: JDK 1.8.0_282, OpenJDK 64-Bit Server VM GraalVM CE 21.0.0.2, 25.282-b07-jvmci-21.0-b06
[info] # VM invoker: /usr/lib/jvm/java-8-graalvm/jre/bin/java
[info] # VM options: -XX:MaxInlineLevel=24 -XX:MaxInlineSize=270
[info] # Warmup: 3 iterations, 10 s each
[info] # Measurement: 5 iterations, 10 s each
[info] # Timeout: 10 min per iteration
[info] # Threads: 2 threads, will synchronize iterations
[info] # Benchmark mode: Average time, time/op
[info] # Benchmark: kamon.bench.ThreadLocalStorageBenchmark.crossThreadLocal
[info] # Run progress: 0.00% complete, ETA 00:02:40
[info] # Fork: 1 of 1
[info] Picked up JAVA_TOOL_OPTIONS: -XX:MaxInlineLevel=24 -XX:MaxInlineSize=270
[info] # Warmup Iteration   1: 13.610 ns/op
[info] # Warmup Iteration   2: 13.379 ns/op
[info] # Warmup Iteration   3: 14.535 ns/op
[info] Iteration   1: 14.325 ns/op
[info] Iteration   2: 14.308 ns/op
[info] Iteration   3: 14.341 ns/op
[info] Iteration   4: 14.323 ns/op
[info] Iteration   5: 14.383 ns/op
[info] Result "kamon.bench.ThreadLocalStorageBenchmark.crossThreadLocal":
[info]   14.336 ±(99.9%) 0.110 ns/op [Average]
[info]   (min, avg, max) = (14.308, 14.336, 14.383), stdev = 0.029
[info]   CI (99.9%): [14.226, 14.446] (assumes normal distribution)
[info] # JMH version: 1.21
[info] # VM version: JDK 1.8.0_282, OpenJDK 64-Bit Server VM GraalVM CE 21.0.0.2, 25.282-b07-jvmci-21.0-b06
[info] # VM invoker: /usr/lib/jvm/java-8-graalvm/jre/bin/java
[info] # VM options: -XX:MaxInlineLevel=24 -XX:MaxInlineSize=270
[info] # Warmup: 3 iterations, 10 s each
[info] # Measurement: 5 iterations, 10 s each
[info] # Timeout: 10 min per iteration
[info] # Threads: 2 threads, will synchronize iterations
[info] # Benchmark mode: Average time, time/op
[info] # Benchmark: kamon.bench.ThreadLocalStorageBenchmark.fastThreadLocal
[info] # Run progress: 50.00% complete, ETA 00:01:20
[info] # Fork: 1 of 1
[info] Picked up JAVA_TOOL_OPTIONS: -XX:MaxInlineLevel=24 -XX:MaxInlineSize=270
[info] # Warmup Iteration   1: 12.372 ns/op
[info] # Warmup Iteration   2: 12.327 ns/op
[info] # Warmup Iteration   3: 12.380 ns/op
[info] Iteration   1: 12.363 ns/op
[info] Iteration   2: 12.541 ns/op
[info] Iteration   3: 12.448 ns/op
[info] Iteration   4: 13.155 ns/op
[info] Iteration   5: 13.782 ns/op
[info] Result "kamon.bench.ThreadLocalStorageBenchmark.fastThreadLocal":
[info]   12.858 ±(99.9%) 2.323 ns/op [Average]
[info]   (min, avg, max) = (12.363, 12.858, 13.782), stdev = 0.603
[info]   CI (99.9%): [10.534, 15.181] (assumes normal distribution)
[info] # Run complete. Total time: 00:02:40
[info] Benchmark                                     Mode  Cnt   Score   Error  Units
[info] ThreadLocalStorageBenchmark.crossThreadLocal  avgt    5  14.336 ± 0.110  ns/op
[info] ThreadLocalStorageBenchmark.fastThreadLocal   avgt    5  12.858 ± 2.323  ns/op
```

TBH, I don't think the "gain" of the optimized threadlocal implementation is worth keeping (the large variability of the fastThreadLocal is especially interesting). But I'm biased as I primarily work with the Typelevel stack so I'd like to minimize the steps required to get it working with Kamon, especially when failing to do so will lead to horrible cross-thread context contamination

### DIFF
--- a/core/kamon-core-bench/src/main/scala/kamon/bench/ThreadLocalStorageBenchmark.scala
+++ b/core/kamon-core-bench/src/main/scala/kamon/bench/ThreadLocalStorageBenchmark.scala
@@ -16,9 +16,9 @@
 
 package kamon.bench
 
-import java.util.concurrent.TimeUnit
+import kamon.context.Storage.CrossThreadLocal
 
-import kamon.context.Storage.Scope
+import java.util.concurrent.TimeUnit
 import kamon.context.{Context, Storage}
 import org.openjdk.jmh.annotations._
 
@@ -28,7 +28,7 @@ class ThreadLocalStorageBenchmark {
   val TestKey: Context.Key[Int] = Context.key("test-key", 0)
   val ContextWithKey: Context = Context.of(TestKey, 43)
 
-  val TLS: Storage =  new OldThreadLocal
+  val CrossTLS: Storage =  new CrossThreadLocal
   val FTLS: Storage =  new Storage.ThreadLocal
 
 
@@ -36,11 +36,11 @@ class ThreadLocalStorageBenchmark {
   @BenchmarkMode(Array(Mode.AverageTime))
   @OutputTimeUnit(TimeUnit.NANOSECONDS)
   @Fork
-  def currentThreadLocal: Context = {
-    val scope = TLS.store(ContextWithKey)
-    TLS.current()
+  def crossThreadLocal: Context = {
+    val scope = CrossTLS.store(ContextWithKey)
+    CrossTLS.current()
     scope.close()
-    TLS.current()
+    CrossTLS.current()
   }
 
   @Benchmark
@@ -53,29 +53,4 @@ class ThreadLocalStorageBenchmark {
     scope.close()
     FTLS.current()
   }
-}
-
-
-class OldThreadLocal extends Storage {
-  private val tls = new java.lang.ThreadLocal[Context]() {
-    override def initialValue(): Context = Context.Empty
-  }
-
-  override def current(): Context =
-    tls.get()
-
-  override def store(context: Context): Scope = {
-    val newContext = context
-    val previousContext = tls.get()
-    tls.set(newContext)
-
-    new Scope {
-      override def context: Context = newContext
-      override def close(): Unit = tls.set(previousContext)
-    }
-  }
-}
-
-object OldThreadLocal {
-  def apply(): OldThreadLocal = new OldThreadLocal()
 }
--- a/core/kamon-core/src/main/scala/kamon/ContextStorage.scala
+++ b/core/kamon-core/src/main/scala/kamon/ContextStorage.scala
@@ -141,8 +141,10 @@ object ContextStorage {
     * instrumentation follows them around.
     */
   private val _contextStorage: Storage = {
-    if(sys.props("kamon.context.debug") == "true")
+    if (sys.props("kamon.context.debug") == "true")
       Storage.Debug()
+    else if (sys.props("kamon.context.crossThread") == "true")
+      Storage.CrossThreadLocal()
     else
       Storage.ThreadLocal()
   }
--- a/core/kamon-core/src/main/scala/kamon/ContextStorage.scala
+++ b/core/kamon-core/src/main/scala/kamon/ContextStorage.scala
@@ -141,11 +141,18 @@ object ContextStorage {
     * instrumentation follows them around.
     */
   private val _contextStorage: Storage = {
+    val storageTypeStr = Option(sys.props("kamon.context.storageType"))
+
     if (sys.props("kamon.context.debug") == "true")
       Storage.Debug()
-    else if (sys.props("kamon.context.crossThread") == "true")
-      Storage.CrossThreadLocal()
-    else
-      Storage.ThreadLocal()
+    else {
+      storageTypeStr match {
+        case None => Storage.CrossThreadLocal()
+        case Some("debug") => Storage.Debug()
+        case Some("sameThreadScope") => Storage.ThreadLocal()
+        case Some("default") => Storage.CrossThreadLocal()
+        case Some(other) => throw new IllegalArgumentException(s"Unrecognized kamon.context.storageType value: $other")
+      }
+    }
   }
 }
--- a/core/kamon-core/src/main/scala/kamon/context/Storage.scala
+++ b/core/kamon-core/src/main/scala/kamon/context/Storage.scala
@@ -92,12 +92,13 @@ object Storage {
     * Wrapper that implements an optimized ThreadLocal access pattern ideal for heavily used ThreadLocals. It is faster
     * to use a mutable holder object and always perform ThreadLocal.get() and never use ThreadLocal.set(), because the
     * value is more likely to be found in the ThreadLocalMap direct hash slot and avoid the slow path of
-    * ThreadLocalMap.getEntryAfterMiss(). Closing of the returned Scope **MUST** be called in the same thread as
-    * store(..) was originally called.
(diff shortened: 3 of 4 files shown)


core/kamon-core-bench/src/main/scala/kamon/bench/ThreadLocalStorageBenchmark.scala
```diff
@@ -54,28 +54,3 @@ class ThreadLocalStorageBenchmark {
     FTLS.current()
   }
 }
-
-
-class OldThreadLocal extends Storage {
```
dpsoft, line 59: @jatcwang please don't remove the `OldThreadLocal`(is our baseline in the microbenchmark), maybe is a good idea change the name to something like `BaselineThreadLocal` and in the test add `CrossThreadLocal` in order to measure against the baseline WDYT?

dpsoft:
@jatcwang looks good to me! 

btw, in my local machine I got this bench results: the variability could be generated by some background process, if I don't close chrome my benchmarks are garbage. :(
```
[info] # JMH version: 1.25
[info] # VM version: JDK 1.8.0_252, OpenJDK 64-Bit Server VM, 25.252-b09
[info] # VM invoker: /Users/dparra/.sdkman/candidates/java/8.0.252.hs-adpt/jre/bin/java
[info] # VM options: <none>
[info] # Warmup: 5 iterations, 10 s each
[info] # Measurement: 10 iterations, 10 s each
[info] # Timeout: 10 min per iteration
[info] # Threads: 2 threads, will synchronize iterations
...
```
```
[info] Benchmark                                         Mode  Cnt   Score   Error  Units
[info] ThreadLocalStorageBenchmark.currentThreadLocal    avgt   50  17.712 ± 0.425  ns/op
[info] ThreadLocalStorageBenchmark.fastThreadLocal       avgt   50  13.102 ± 0.255  ns/op
```

dpsoft (approved):

jatcwang:
Cool. @dpsoft can you retrigger CI. Seems like a flaky test

dpsoft:
I just ran again the tests!

core/kamon-core/src/main/scala/kamon/ContextStorage.scala
```diff
@@ -141,8 +141,10 @@ object ContextStorage {
     * instrumentation follows them around.
     */
   private val _contextStorage: Storage = {
-    if(sys.props("kamon.context.debug") == "true")
+    if (sys.props("kamon.context.debug") == "true")
```
SimunKaracic, line 144: Now that the choice is between a couple of actual implementations, this should probably be loaded from the configuration file, not the environment

core/kamon-core/src/main/scala/kamon/context/Storage.scala
```diff
@@ -104,6 +136,8 @@ object Storage {
     * different stack traces for every store/close operation pair. Do not use this for any reason other than debugging
     * Context propagation issues (like, dirty Threads) in a controlled environment.
     *
+    * Note: Similar to the default ThreadLocal storage, the scope must be closed in the same thread as where store(..)
+    * was called, otherwise you are potentially modifying the context of another thread.
```
SimunKaracic, line 140: :+1:

core/kamon-core/src/main/scala/kamon/context/Storage.scala
```diff
@@ -57,11 +57,43 @@ object Storage {
...
+    def apply(): Storage.CrossThreadLocal =
+      new Storage.CrossThreadLocal()
+  }
+
   /**
     * Wrapper that implements an optimized ThreadLocal access pattern ideal for heavily used ThreadLocals. It is faster
     * to use a mutable holder object and always perform ThreadLocal.get() and never use ThreadLocal.set(), because the
     * value is more likely to be found in the ThreadLocalMap direct hash slot and avoid the slow path of
-    * ThreadLocalMap.getEntryAfterMiss().
+    * ThreadLocalMap.getEntryAfterMiss(). Closing of the returned Scope **MUST** be called in the same thread as
```
SimunKaracic, line 95: Nice addition to the docs!

SimunKaracic (commented):
Great PR, nice job. 

I would only change the `sys.props` thing, otherwise it's good to go!

core/kamon-core/src/main/scala/kamon/ContextStorage.scala
ivantopo: Folks, I'm thinking that we should make `CrossThreadLocal` the default context storage implementation. The reasons why I think that are:
- It will not affect the behavior of any of the manual/automatic instrumentation. We always aimed to close scopes on the same thread where we created them, and in those situations this implementation will produce the same results. In 99% of the cases we need to close the scope in the same thread we created it anyways, and Cats/Monix implementations are an exception to that common case so we will continue to recommended closing scopes on the same thread, unless it is very clear that there is a reason for doing otherwise, as it is for Cats/Monix.
- It makes it easier to get started. If there is something I learned over the years is that users almost never read the fine print :smile:.. if users need to add a JVM property but that property is controlled from the outside world (IDE, command line parameters or container orchestrators), there will be several cases of things "not working" in certain environments. I would rather go with an option that makes it work by default for everyone.
- In the specific cases where someone needs that little boost in performance (I'm counting to be very few cases here) they can add the extra JVM options to use the optimized TLS.

What do you think about that?

SimunKaracic: This all sounds reasonable, and simplifying semantics is always good!
I'm just scared because this is a "breaking" change, and I have no intuition for how often this code is called. 
If you're sure that we won't see a big performance drop, let's do it, and just make sure to document clearly and make not of it in the release notes.

jatcwang: I totally agree @ivantopo (Makes my life easier that's for sure ;)). Happy to make the change if @SimunKaracic agree too

SimunKaracic: I agree :handshake:

jatcwang: This is done.

jatcwang:
Can someone please retrigger the CI? Seems like a flakey test

SimunKaracic:
Thank you for the contribution! :D

PR #969, opened 2021-03-24
(status page) Add missing api key state

the-overengineer:
Add a special state for Kamon APM reporter that guides the user in case
their API key is not configured or is invalid.

Update Typescript and underscore to allow for modern dev practices.

The link itself will change once we have the new onboarding. We could wait to merge this until then.

### DIFF
--- a/core/kamon-status-page/src/main/vue/src/components/ModuleList.vue
+++ b/core/kamon-status-page/src/main/vue/src/components/ModuleList.vue
@@ -21,6 +21,7 @@
         :key="reporter.name"
         :module="reporter"
         :is-missing-key="isMisconfiguredApmModule(reporter)"
+        @show:api-key="$emit('show:api-key')"
       />
       <div v-if="!hasApmModule" class="apm-suggestion">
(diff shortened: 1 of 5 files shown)


core/kamon-status-page/src/main/vue/src/components/ModuleList.vue
```diff
@@ -16,7 +16,12 @@
...
 
-      <module-status-card v-for="reporter in reporterModules" :key="reporter.name" :module="reporter" />
+      <module-status-card
+        v-for="reporter in reporterModules"
+        :key="reporter.name"
+        :module="reporter"
+        :is-missing-key="isMisconfiguredApmModule(reporter)"
+      />
       <div v-if="!hasApmModule" class="apm-suggestion">
         <a href="https://kamon.io/apm/?utm_source=kamon&utm_medium=status-page&utm_campaign=kamon-status" target="_blank">
```
the-overengineer, line 26: While touching, want me to change this, @ivantopo?

PR #970, opened 2021-03-25
Add order config

SimunKaracic:


dpsoft:
related to https://github.com/kamon-io/kanela/pull/118

Issue #973, opened 2021-03-29
Add support for an OpenTelemetry trace/span reporter


pnerg:
Kamon needs to add support for exporting traces according to the [OpenTelemetry Protocol (OTLP)](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md).
 
This ticket is added as a placeholder as I intend to donate my ongoing implementation [kamon-otel-trace-exporter](https://github.com/pnerg/kamon-otel-trace-exporter) once I have it in decent shape.

SimunKaracic:
We eagerly await your donation, feel free to open a WIP pull request or ask for help!

pnerg:
@SimunKaracic just did an integration test with the exporter and the released w3c context propagation....resulted in #975 :)

Apart from that the code is in working condition. Will make a PR soonish, the only thing I'll leave out is support for TLS.
Think that can come in a later release. Guess there's very few apps actually using TLS for reporting trace data.

PR #977, opened 2021-03-31
Added support for opentelemetry trace exporter

pnerg:
# Overview
This PR adds an exporter for exporting traces/spans according to [OpenTelemetry Protocol (OTLP)](https://github.com/open-telemetry/opentelemetry-proto) as requested in #973.
The code has been tested against an instance of the [OpenTelemetry Collector](https://opentelemetry.io/docs/collector/)

# What is not supported from the OTEL spec
* metrics (should be done at some point as a separate exporter in the same sbt sub-module)
* logs (spec still in draft and probably not for Kamon anyways)

# What is not implemented in this first draft
Some things were left out for sake of getting something out there.

## Support for custom trust store for managing TLS
The exporter supports HTTP (default) and HTTPS but there is no way to add custom trust store or keychains

## No labels/tags/keyvalues are dropped 
The spec allows for dropping meta-information in case if. 
  * the meta-data is too long
  * there's too many instances of meta-data (e.g tags)
  
But dropping meta-information requires that the exporter notes how many were dropped.
So for sake of simplicity the exporter includes all environment/span tags

# Assumptions made on the OTLP model
The data model for exporting spans looks like:
```
ExportServiceRequest - -> *  ResourceSpans
ResourceSpans -->* InstrumentationLibrarySpans
ResourceSpans -->1 Resource
InstrumentationLibrarySpans --> * Span
InstrumentationLibrarySpans --> 1 InstrumentationLibrary
```
1- is 1-1
*- is one to many

So one `ExportServiceRequest` can have multiple`ResourceSpans`which in turn can have multiple `InstrumentationLibrarySpans`which finally has multiple `Spans`

I've assumed that always is a single resource (i.e. the application itself) and there is only one instrumentation library (kamon) and all spans belong in that chain.
So the meta-information in the `Resource` belonging to `ResourceSpans` comes largely from optional Kamon env tags.
The meta-information in the instrumentation library is info on Kamon.

# Local testing
I've used the OTEL collector in a container and provided it with config to just print all traces it receives
otel-collector.yaml
```yaml
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:55690

exporters:
  logging:
    loglevel: debug

processors:
  batch:

service:
  pipelines:
    traces:
      receivers: [otlp]
      processors: [batch]
      exporters: [logging]
```
Then just run this command  (requires Docker)
```bash
docker run --rm -p 55690:55690 \
-v "${PWD}/otel-collector.yaml":/otel-local-config.yaml \
--name otelcol otel/opentelemetry-collector \
--config otel-local-config.yaml; \
```

## Example printout from OTEL collector
Note the _Resource labels_ as they're from the Kamon env tags apart from `telemetry.sdk.*` which we always add.
The _InstrumentationLibrary kamon 2.1.14_ is the `InstrumentationLibrary` meta-info provided
```
Resource labels:
     -> service.name: STRING(otlp-test-app)
     -> telemetry.sdk.name: STRING(kamon)
     -> telemetry.sdk.language: STRING(scala)
     -> telemetry.sdk.version: STRING(2.1.14)
     -> notset: STRING(/Users/peter.nerg)
     -> env: STRING(staging)
     -> service.instance.id: STRING(xxx-yyy)
     -> service.namespace: STRING(namespace-1)
     -> service.version: STRING(x.x.x)
InstrumentationLibrarySpans #0
InstrumentationLibrary kamon 2.1.14
Span #0
    Trace ID       : 4a7e482ebab331570cfbfc070dd5e70f
    Parent ID      : 
    ID             : 2c47181be9a542a2
    Name           : /purchase-item/{id}
    Kind           : SPAN_KIND_SERVER
    Start time     : 2021-03-30 12:48:37.244977523 +0000 UTC
    End time       : 2021-03-30 12:48:37.593127187 +0000 UTC
    Status code    : STATUS_CODE_OK
    Status message : 
Attributes:
     -> method: STRING(POST)
Span #1
    Trace ID       : 4a7e482ebab331570cfbfc070dd5e70f
    Parent ID      : 2c47181be9a542a2
    ID             : 4ccac7b8965abd99
    Name           : add-order
    Kind           : SPAN_KIND_CLIENT
    Start time     : 2021-03-30 12:48:37.288737377 +0000 UTC
    End time       : 2021-03-30 12:48:37.592803111 +0000 UTC
    Status code    : STATUS_CODE_OK
    Status message : 
Attributes:
     -> method: STRING(put)
Span #2
    Trace ID       : 4a7e482ebab331570cfbfc070dd5e70f
    Parent ID      : 2c47181be9a542a2
    ID             : c9c8937819d35d41
    Name           : check-credits
    Kind           : SPAN_KIND_CLIENT
    Start time     : 2021-03-30 12:48:37.258310723 +0000 UTC
    End time       : 2021-03-30 12:48:37.284974414 +0000 UTC
    Status code    : STATUS_CODE_OK
    Status message : 
Attributes:
     -> method: STRING(query)

```

### DIFF
--- a/reporters/kamon-opentelemetry/src/main/resources/reference.conf
+++ b/reporters/kamon-opentelemetry/src/main/resources/reference.conf
@@ -1,0 +1,43 @@
+# ======================================== #
+# kamon-otlp reference configuration       #
+# ======================================== #
+
+kamon.otel.trace {
+  # Hostname and port where the OTLP Server is running
+  host = "localhost"
+  port = 4317
+
+  # Decides whether to use HTTP or HTTPS when connecting to the OTel server
+  protocol = "http"
+
+  # default to support the ENV:s as described at
+  # https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md
+  endpoint = ${kamon.otel.trace.protocol}"://"${kamon.otel.trace.host}":"${kamon.otel.trace.port}
+  endpoint = ${?OTEL_EXPORTER_OTLP_ENDPOINT}
+  endpoint = ${?OTEL_EXPORTER_OTLP_TRACES_ENDPOINT}
+
+  # Enable or disable including tags from kamon.environment.tags as resource labels in the exported trace
+  # Any keys containing a hyphen (-) will be converted to a dot (.) as that is the standard.
+  # e.g. service-version becomes service.version
+  # reference: https://github.com/kamon-io/Kamon/blob/master/core/kamon-core/src/main/resources/reference.conf#L23
+  include-environment-tags = no
+
+  # Arbitrary key-value pairs that further identify the environment where this service instance is running.
+  # These are added as KeyValue labels to the Resource part of the exported traces
+  # Requires 'include-environment-tags' to be set to 'yes'
+  #
+  # kamon.environment.tags {
+  #   service-version = "x.x.x"
+  #   service-namespace = "ns"
+  #   service-instance.id = "xxx-yyy"
+  # }
+}
+
+kamon.modules {
+  otel-trace-reporter {
+    enabled = true
+    name = "OpenTelemetry Trace Reporter"
+    description = "Sends trace data to a OpenTelemetry server via gRPC"
+    factory = "kamon.otel.OpenTelemetryTraceReporter$Factory"
+  }
+}
--- a/reporters/kamon-opentelemetry/src/main/scala/kamon/otel/SpanConverter.scala
+++ b/reporters/kamon-opentelemetry/src/main/scala/kamon/otel/SpanConverter.scala
@@ -1,0 +1,211 @@
+/*
+ * Copyright 2013-2021 The Kamon Project <https://kamon.io>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package kamon.otel
+
+import com.google.protobuf.ByteString
+import io.opentelemetry.proto.common.v1.{AnyValue, InstrumentationLibrary, KeyValue}
+import io.opentelemetry.proto.resource.v1.Resource
+import io.opentelemetry.proto.trace.v1.{InstrumentationLibrarySpans, ResourceSpans, Status, Span => ProtoSpan}
+import kamon.tag.{Lookups, Tag}
+import kamon.tag.Tag.unwrapValue
+import kamon.trace.Span.{Kind, TagKeys}
+import kamon.trace.{Identifier, Span}
+import org.slf4j.LoggerFactory
+
+import java.time.Instant
+import java.util.Collections
+import java.util.concurrent.TimeUnit
+
+/**
+ * Converts Kamon spans to OpenTelemetry proto Spans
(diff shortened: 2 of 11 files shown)


pnerg:
First test run was green, now something related to MongoDB broke.

pnerg:
Hi guys, any progress/comments on this PR or did I DoS attack you?
Even if the PR is quite big it is at least isolated as I've added a new functionality and not touched the existing code.
So let's go brave and hit that "approve" button, you know you want to do that...:wink:
"There is no problem in the code, let it be approved"/Jedi mind trick

SimunKaracic:
Don't worry, I'll take a hard look at everything you did tomorrow ;)

SimunKaracic:
Looking great, no major complaints for me. I'll have to dive a bit deeper though, so on monday expect either smaller feedback or just a straight up merge

reporters/kamon-opentelemetry/src/main/scala/kamon/otel/SpanConverter.scala
```diff
@@ -0,0 +1,211 @@
...
+      .setStartTimeUnixNano(toEpocNano(span.from))
+      .setEndTimeUnixNano(toEpocNano(span.to))
+      .setStatus(getStatus(span))
+      .addAllAttributes(attributes.asJava)
+      .addAllLinks(links.asJava)
+
+    //TODO add set traceState once we have something to set. Need w3c context
+    //It is a trace_state in w3c-trace-context format: https://www.w3.org/TR/trace-context/#tracestate-header
+    //See also https://github.com/w3c/distributed-tracing for more details about this field.
+    //builder.setTraceState()
```
SimunKaracic, line 99: Can this be done with the current support for w3c trace context?

reporters/kamon-opentelemetry/src/main/resources/reference.conf
```diff
@@ -0,0 +1,43 @@
...
+}
+
+kamon.modules {
+  otel-trace-reporter {
+    enabled = true
+    name = "OpenTelemetry Trace Reporter"
+    description = "Sends trace data to a OpenTelemetry server via gRPC"
+    factory = "kamon.otel.OpenTelemetryTraceReporter$Factory"
+  }
+}
```
SimunKaracic, line 43: Fix these endlines please!

SimunKaracic:
Other than those few small nitpicks, this is great. There's still work to do, but what is there is readable and well tested.
GJ!

reporters/kamon-opentelemetry/src/main/scala/kamon/otel/SpanConverter.scala
pnerg: Unsure where to get it from, the TraceState is stored in the Context
`context.withEntry(Span.Key, span).withEntry(TraceStateKey, traceState)`
https://github.com/kamon-io/Kamon/blob/master/core/kamon-core/src/main/scala/kamon/trace/SpanPropagation.scala#L56

But as the reporter only sees the `Span.Finished` which does not contain any W3C specific data.   
IMO one would need to update the `Span.Finished` case class to contain some meta fields or similar containing the `TraceState` key/value

pnerg:
Yep, there's a few things that can be added in the continuation. That's why I added the notes in this PR on what is yet to be done. Also left a few TODOs in the code to mark where more features can be added.
IMO this makes for a decent first version it at least covers the most important parts of the OTLP spec.

reporters/kamon-opentelemetry/src/main/scala/kamon/otel/SpanConverter.scala
SimunKaracic: We'll leave it for a later update then!

SimunKaracic:
Thanks for the contribution!
Will try to release this today!

pnerg:
Any plans for when you'll do a new release?
#impatient

SimunKaracic:
Now where would the fun be in just _doing things on time_? 

Jokes aside, thanks for the patience :tada:

PR #1000, opened 2021-04-16
Add tapir instrumentation

SimunKaracic:


jatcwang:
Question: This will conflict with other http library integration right? So if you're using this with e.g. akka-http backend you should turn off the akka-http server instrumentation?

SimunKaracic:
There should be no conflicts, you'll just get different name spans on your endpoints if you're using both together!
It's working only for akka-http backends right now :)

PR #1002, opened 2021-04-23
Build core modules for Scala 3

bplommer:
CI is failing with a problem starting sbt - possibly an issue with `csbt` / a result of needing to upgrade to SBT 1.5.0

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -33,6 +33,7 @@ lazy val `kamon-core` = (project in file("core/kamon-core"))
   .settings(
     buildInfoKeys := Seq[BuildInfoKey](version),
     buildInfoPackage := "kamon.status",
+    crossScalaVersions += "3.0.0-RC3",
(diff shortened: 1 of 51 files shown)


SimunKaracic:
I'm gonna close this for now, but I'll reference it when we start upgrading to Scala 3. 
I also think support for Scala 3 is probably going to come in few smaller PR's, making incremental changes and making sure we don't break anything.
First step is updating all scala 2.13.x versions to 2.13.4 so we can see if there's any immediate problems.
If closing this PR didn't destroy your enthusiasm @bplommer, feel free to start with that.

With all of that out of the way, thanks for the contribution, it's always great to see community involvement :heart:

btomala:
Hi Guys, I was looking for information about Kamon compatibility with Scala 3. This is what I found. Connot find anything on official webpage. What are the plans related to upgrade to Scala 3? Is it on road map already?

Issue #1008, opened 2021-04-26
Kamon emits invalid bytecode when instrumenting sqlite versions > 3.32.3.2


Falmarri:
```
java.lang.VerifyError: Expecting a stackmap frame at branch target 20
Exception Details:
  Location:
    org/sqlite/jdbc3/JDBC3Statement.executeBatch()[I @17: goto
  Reason:
    Expected stackmap frame at this location.
  Bytecode:
    0000000: b201 6c2a b601 6fa7 000e 1301 4d5f b801
    0000010: 53a7 0003 014c 2a4d 2cb6 0002 2cb4 0024
    0000020: c600 0a2c b400 259a 0009 03bc 0aa7 00ac
    0000030: 2cb4 0025 bc0a 4e2c b400 04b6 0005 3a04
    0000040: 1904 593a 05c2 0336 0615 062d bea2 006a
    0000050: 2c2c b400 2415 0632 c000 71b5 0007 1904
    0000060: 2cb6 0008 2d15 0619 042c 01b6 002a 4f19
    0000070: 042c b600 2b57 a700 3b3a 07bb 002c 59bb
    0000080: 002d 59b7 002e 122f b600 3015 06b6 0031
    0000090: 1232 b600 3019 07b6 0033 b600 30b6 0034
    00000a0: 2db7 0035 bf3a 0819 042c b600 2b57 1908
    00000b0: bf84 0601 a7ff 952c b600 36a7 000c 3a09
    00000c0: 2cb6 0036 1909 bf19 05c3 a700 0b3a 0a19
    00000d0: 05c3 190a bf2d a700 034d 014e a700 064e
    00000e0: 014d b201 6c2b 2db6 0170 a700 0d13 014d
    00000f0: 5fb8 0153 a700 032d c600 052d bf2c b0  
  Exception Handler Table:
    bci [80, 111] => handler: 121
    bci [80, 111] => handler: 165
    bci [121, 167] => handler: 165
    bci [70, 183] => handler: 190
    bci [190, 192] => handler: 190
    bci [70, 202] => handler: 205
    bci [205, 210] => handler: 205
    bci [0, 10] => handler: 10
    bci [24, 217] => handler: 223
    bci [226, 237] => handler: 237
  Stackmap Table:
    same_locals_1_stack_item_frame(@10,Object[#114])
    same_locals_1_stack_item_frame(@21,Object[#341])
    append_frame(@22,Object[#341])
    append_frame(@24,Object[#112])
[...]

PR #1009, opened 2021-04-27
Fix sqlite-jdbc insrumentation for > 3.32.3.2 #1008

Falmarri:
This is more of a temporary workaround than a fix. It disables
instrumentation on the non-prepared statement versions of queries


SimunKaracic:
While this looks good, please be patient and give us a couple days to look into this. 
We've seen issues like this before, and perhaps there is something else that we could do to fix this, so this workaround is not necessary.
If we dont come up with something after that, we'll probably merge this!

Falmarri:
Sounds good. No rush. I'm going to just remove the sqlite classes from the "within" kanela config in the short term.

I wonder if there's a change that can be made upstream to sqlite to fix this as well if it can't be addressed here

SimunKaracic:
I've opened up a PR with a simpler fix #1023 , which will get merged if we don't find a Kanela only solution. 

Tanks for the PR anyway @Falmarri , it's always nice to see people contributing <3

Falmarri:
No worries. I was hoping there was a better fix that didn't just disable this

PR #1013, opened 2021-05-04
Parsing sql for operation name

Falmarri:


SimunKaracic:
Thanks for the PR :tada: 
The original discussion:
https://gitter.im/kamon-io/Kamon?at=609069f84d0065337e4f8116

We'll take a look at it and drop feedback here, but we're swamped for the next couple of days so you'll have to be a bit patient!

SimunKaracic:
So, this looks great, and we'd like to merge it. 
However, we still need to figure out if it has a large performance penalty. 
For now, what we should do is hide this feature behind a flag, merge it, and show people how to disable/enable it. 

What do you say, are you up for it? :D

Falmarri:
Yeah I'll mess with it a bit as I get some time. The one thing I wanted to do was to add a cache of statements, but wasn't sure how to add guava and have it be shaded (i don't think I'm shading the sql parser either). But maybe that can be done after we see the performance impact

SimunKaracic:
I think guava might be a bit overkill, a simple TrieMap might be better. 
But yeah, let's see the performance impact first. 
Add a flag to the reference.conf, fixup the merge conflicts, and we'll take it from there

SimunKaracic (commented):
Looks good for now, I'll test it later today/tomorrow and come back with some more feedback (if necessary)

SimunKaracic:
This is about 90% done! 
@Falmarri  do you have time to finish it? 
If you don't I'd be glad to take this over from you

Falmarri:
@SimunKaracic Feel free to finish it. I'm not sure when I'll have time but it probably won't be for a week or so.

PR #1015, opened 2021-05-05
Added the possibility to filter out context tags not to propagate

pnerg:
This adds the possibility to filter (Glob) out tags NOT to propagate.  
Applications may be using the context for storing tags which one does not want to automatically propagate.
In my case I have certain tags I handle in custom readers/writers and therefore don't want the default propagation to manage them.

By default the filter is empty, i.e. all tags are propagated just as it works now


SimunKaracic:
Docs, tests, and clean code :+1: 
Thanks for the PR

Issue #1014, opened 2021-05-05
akka-http : Duplicate context headers when propagating a trace


pnerg:
# Issue
When debugging headers sent between applications I noticed an odd behaviour. 
Seems like context related headers are duplicated during the propagation.
I see double `b3` and `context-tags` headers.

Kamon version used: `"io.kamon" %% "kamon-bundle" % "2.1.17"`

# Tests to prove behaviour
To illustrate the behaviour I created a small app (see below) that supports two paths 
* /forward - prints headers and sends a GET to /hello
* /hello - prints headers and just responds with a string

## Sending a request via Postman
```
GET http://localhost:9669/forward
X-B3-TraceId: aaaaaaaaaaaaaa
X-B3-SpanId: 1111111111111112
```

## Log from test app
```
[info] --------------
[info] Got request for '/forward' parsed/generated context traceId[aaaaaaaaaaaaaa], spanId[39760408a4dc2da0], parentSpanId[1111111111111112]
[info] Timeout-Access: <function1>
[info] X-B3-SpanId: 1111111111111112
[info] X-B3-TraceId: aaaaaaaaaaaaaa
[info] User-Agent: PostmanRuntime/7.26.8
[info] Accept: */*
[info] Postman-Token: 1e33d481-8b0a-4b28-ab09-b15313b89da2
[info] Host
[info] Accept-Encoding: gzip, deflate, br
[info] Connection: keep-alive
[info] --------------
[info] Got request for '/hello' parsed/generated context traceId[aaaaaaaaaaaaaa], spanId[0cc92d692023e180], parentSpanId[5931484589822481]
[info] Timeout-Access: <function1>
[info] Host
[info] X-B3-Sampled: 1
[info] X-B3-ParentSpanId: 39760408a4dc2da0
[info] X-B3-SpanId: 5931484589822481
[info] X-B3-TraceId: aaaaaaaaaaaaaa
[info] context-tags: upstream.name=area51-server;
[info] X-B3-Sampled: 1
[info] X-B3-ParentSpanId: 5931484589822481
[info] X-B3-SpanId: 507f6b65a922d102
[info] X-B3-TraceId: aaaaaaaaaaaaaa
[info] context-tags: upstream.name=area51-server;
[info] User-Agent: akka-http/10.2.4
```
### Observation 1 : Wrong spanID in context
The first log to print parsed context states:
```
[...]

PR #1018, opened 2021-05-13
Stop doubly instrumenting outgoing requests

SimunKaracic:
The no longer instrumented method just delegates to the PoolMaster


SimunKaracic:
Should fix #1014. I've recreated the tests from that issue locally and I no longer see double headers!

PR #1019, opened 2021-05-14
Add source mappings from custom configurations

steinybot:
The [published sources](https://repo1.maven.org/maven2/io/kamon/kamon-akka_2.13/2.1.18/kamon-akka_2.13-2.1.18-sources.jar) contain nothing but a MANIFEST.MF.

I suspect that this is because of how the custom configurations have been done.

This works with `kamon-akka/publishLocal`.


SimunKaracic:
Thanks for the improvement, we should really do the same for all packages!

PR #1023, opened 2021-05-17
Update instrumentation and sqlite version

SimunKaracic:
Initial fix for #1008. 
Waiting for more input from @dpsoft before merging, this might not be necessary if he can find a Kanela only solution


dpsoft:
@SimunKaracic please merge, I need resolve the kanela release process (bintray is dead) and not sure how long it will take...

SimunKaracic:
Good luck with that! The best that I found is this, which claims we should be able to publish to maven central (i didn't try it though):
https://github.com/olafurpg/sbt-ci-release#how-do-i-publish-sbt-plugins

PR #1026, opened 2021-05-27
introduce Span Reporting delay

ivantopo:
This PR introduces the `kamon.trace.span-reporting-delay` setting to allow users to keep Spans in memory for a period of time before reporting them, in case the they want to change a Trace's sampling decision programmatically. 

The idea is to allow people to do tail-based sampling, at least at the local service level. The reporting delay is zero by default because it doesn't make sense to delay Spans in case folks explicitly want to change sampling decisions.


SimunKaracic (approved):
LGTM :rocket:

PR #1029, opened 2021-05-28
Maven BOM for Kamon

ihostage:
I think that Kamon should have a BOM like many other multi-module frameworks and libraries. 😉


SimunKaracic:
So, if I understand this correctly, `kamon-bundle` will now include a new subproject, `kamon-bom`.
LGTM, thanks for the contribution :tada: :rocket:

Issue #1021, opened 2021-05-17
Subsequent Akka ask spans do not share the same parent


steinybot:
Expected:

The spans created by:
```
val future = for {
  _ <- first ? "Hello?"
  _ <- second ? "Goodbye!"
} yield ()
```

Should have the same parent.

Actual:

The span from the first ask is the parent of the second.

See this [unit test](https://github.com/steinybot/Kamon/blob/akka-ask/instrumentation/kamon-akka/src/test-common/scala/kamon/instrumentation/akka/AskPatternInstrumentationSpec.scala#L85-L109) which reproduces the problem.

steinybot:
I'm not sure if this is a good workaround but it works:
```
        val future = for {
          _ <- Kamon.runWithSpan(span) {
            first ? "Hello?"
          }
          _ <- Kamon.runWithSpan(span) {
            second ? "Goodbye!"
          }
        } yield ()
```

ivantopo:
Hey @steinybot!

I remember working on this long ago. Not specifically for the ask pattern, but for Futures which is what seems to be kicking-in here. I think it was around the time of this: https://github.com/kamon-io/kamon-futures/pull/12. I do remember having the explicit intent of moving from "all future transformations are siblings" in Kamon 1.x to "all future transformations are a children chain" as it is in Kamon 2.x. I also remember it being super hard to get it done, but I made it happen!

But time seems to prove that it was wrong to do so :sob:. I'm not sure if people are expecting these Spans to be siblings because of historical behavior from Kamon 1.x, or because it is actually more natural to be like that. 

I'm starting to think that we should let users decide whether they want one behavior or the other, probably via some new setting, but that seems like non-trivial work. The more feedback we can get before touching this the better :pray: 

What are your thoughts on that? Should Spans from chained futures be siblings or children of each other?

Also invoking the folks I know might be interested in the topic: @jatcwang @jtjeferreira @SimunKaracic @nadavwr @danielkman @yiminyuan @danischroeter. Please feel free to add more people who might be interested in the topic.

nadavwr:
For comprehensions, much like Haskell's do notation (where this is more apparent) kinda emulates imperative calls:

    do putStr "What is your first name? "
       first <- getLine
       putStr "And your last name? "
       last <- getLine
       let full = first ++ " " ++ last
       putStrLn ("Pleased to meet you, " ++ full ++ "!")

[...]

PR #1033, opened 2021-06-03
add a convenience function for span creation

ivantopo:
This PR brings a simple function to create a Span to wrap the execution of code blocks, with two special behaviors:
- The function stores the created Span in the current context while the function is executed.
- The function will always finish the Span after the wrapped code finishes execution, with special handling for Scala Futures and CompletionStage instances. If the wrapped code returns a Future or CompletionStage, the created Span will be finished when the returned Future/CompletionStage completes.

This function should be enough to replace the soon-to-be-deprecated functions in `ScalaFutureInstrumentation`. More info here: https://github.com/kamon-io/Kamon/issues/1021.

Also, this simplifies cases where people want a simple Span for a block of code and end up having to write something like this:

```
Kamon.runWithSpan(Kamon.spanBuilder(....).start()) {

}
```

### DIFF
--- a/core/kamon-core-tests/src/test/scala/kamon/trace/QuickSpanCreationSpec.scala
+++ b/core/kamon-core-tests/src/test/scala/kamon/trace/QuickSpanCreationSpec.scala
@@ -34,6 +34,21 @@ class QuickSpanCreationSpec extends WordSpec with Matchers with OptionValues wit
       }
     }
 
+    "finish and fail Spans if an exception is thrown while running the wrapped code" in {
+      intercept[Throwable] {
+        span("failSpanWithThrowable") {
+          throw new Throwable()
+          "I'm never going to finish nicely"
+        }
+      }
+
+      eventually(timeout(5 seconds)) {
+        val reportedSpan = testSpanReporter().nextSpan().value
+        reportedSpan.operationName shouldBe "failSpanWithThrowable"
+        reportedSpan.hasError shouldBe true
+      }
+    }
+
     "apply the component tag, if provided" in {
       span("finishSimpleSpanWithComponent", "customComponentTag") {
         "I'm done right away"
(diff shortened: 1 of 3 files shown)


SimunKaracic (approved):
LGTM, merge when ready

core/kamon-core-tests/src/test/scala/kamon/trace/QuickSpanCreationSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+      intercept[Throwable] {
+        span("failSpanWithThrowable") {
+          throw new Throwable()
+          "I'm never going to finish nicely"
+        }
+      }
+
+      eventually(timeout(5 seconds)) {
+        val reportedSpan = testSpanReporter().nextSpan().value
+        reportedSpan.operationName shouldBe "failSpanWithThrowable"
```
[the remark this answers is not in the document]
ivantopo: Done.

core/kamon-core/src/main/scala/kamon/Tracing.scala
```diff
@@ -85,6 +92,80 @@ trait Tracing { self: Configuration with Utilities with ContextStorage =>
...
+        case cs: CompletionStage[_] =>
+          CompletionStageSpanFinisher.finishWhenDone(cs, span).asInstanceOf[A]
+
+        case other =>
+          span.finish()
+          other
+      }
+    } catch {
+      case NonFatal(t) =>
+        span.finish()
```
AlexITC, line 163: Shouldn't `.fail(t)` be invoked before finishing the span?

the-overengineer: It's being handled by `runWithSpan` for this case, so no worries :)
https://github.com/kamon-io/Kamon/blob/ce27fb5d4f5e69dd110dedf091f27b4a6705e8ad/core/kamon-core/src/main/scala/kamon/ContextStorage.scala#L119

PR #1034, opened 2021-06-03
Deprecate scala future instrumentation

ivantopo:
Deprecates the Future Chaining instrumentation, but keeps it in and enabled. More context in https://github.com/kamon-io/Kamon/issues/1021, requires https://github.com/kamon-io/Kamon/pull/1033 to be merged first.

### DIFF
--- a/instrumentation/kamon-scala-future/src/main/resources/reference.conf
+++ b/instrumentation/kamon-scala-future/src/main/resources/reference.conf
@@ -23,9 +23,9 @@ kamon.instrumentation.futures.scala {
 kanela.modules {
 
   scala-future {
-    name = "Scala Future Intrumentation"
+    name = "Scala Future Instrumentation"
     order = 1
-    description = "Provides automatic context propagation to the thread executing a Scala Future's body and callbacks"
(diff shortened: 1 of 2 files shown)


instrumentation/kamon-scala-future/src/main/resources/reference.conf
```diff
@@ -23,9 +23,9 @@ kamon.instrumentation.futures.scala {
 kanela.modules {
 
   scala-future {
-    name = "Scala Future Intrumentation"
+    name = "Scala Future Instrumentation"
     order = 1
-    description = "Provides automatic context propagation to the thread executing a Scala Future's body and callbacks"
+    description = "Deprecated since Kamon 2.1.21, will be removed in Kamon 2.3.0. Provides automatic context propagation to the thread executing a Scala Future's body and callbacks"
```
SimunKaracic, line 28: 2.3.0, not 2.2.0?

SimunKaracic (approved):
Other than the small version confusion, LGTM

instrumentation/kamon-scala-future/src/main/resources/reference.conf
ivantopo: Yeap. We will keep it on Kamon but disabled, then remove it when 2.3.0 comes out.

PR #1035, opened 2021-06-03
Future instrumentation rollback

ivantopo:
This PR rolls back the Future Chaining Instrumentation we introduced on Kamon 2.x, and goes back to the same behavior we had on Kamon 1.x. All the Kamon 2.x instrumentation is kept, but deprecated and disabled by defaults, but users could enable them manually if necessary via configuration.

This is expected to be released with Kamon 2.2.0, and requires #1033 and #1034 to be merged first.

For reference, if anybody wanted to enable the Future Chaining instrumentation in 2.2.0+ they would need to add these settings:

```
kanela.modules {
  executor-service {
    exclude += "scala.concurrent.impl.*"
  }

  scala-future {
    enabled = true
  }
  
  akka-http {
    instrumentations += "kamon.instrumentation.akka.http.FastFutureInstrumentation"
  }  
}
```

All the code related to the Future Chaining instrumentation will be removed in Kamon 2.3.0.

### DIFF
--- a/instrumentation/kamon-akka-http/src/main/resources/reference.conf
+++ b/instrumentation/kamon-akka-http/src/main/resources/reference.conf
@@ -238,6 +238,8 @@ kanela.modules {
     name = "Akka HTTP Instrumentation"
     description = "Provides context propagation, distributed tracing and HTTP client and server metrics for Akka HTTP"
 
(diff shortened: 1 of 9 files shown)


SimunKaracic (approved):
LGTM, great job :tada:

PR #1036, opened 2021-06-07
delete the extracted Kanela file after exiting

ivantopo:
This should ensure that the temporary Kanela agent file extracted by the Bundle will be deleted after the JVM exits. Still need to test this out, though. I don't even get a single temporary file on my machine!


SimunKaracic:
LGTM!

PR #1040, opened 2021-06-15
Truncate 128-bit trace IDs before reporting to DataDog

### LINKED ISSUE #1032: Datadog reporter doesn't handle 128-bit trace IDs
If your application receives a 128-bit trace ID, say from an `x-b3-trace-id` header, and then the application tries to report a span with it, reporting will fail because the [Datadog agent expects a 64-bit Int](https://github.com/DataDog/datadog-agent/blob/140a4ee164261ef2245340c50371ba989fbeb038/pkg/trace/api/version.go#L36) for the trace ID. I discovered this a while ago and have since worked around it using a custom `KamonDataDogTranslator`, but I think the failure mode was that the DD agent discarded the reported trace because the trace ID was too large - it showed up in the DD agent logs.

Regardless, it seems like [DdSpanReporter](https://github.com/kamon-io/Kamon/blob/master/reporters/kamon-datadog/src/main/scala/kamon/datadog/DatadogSpanReporter.scala#L39) should use the lower 64 bits of a 128-bit trace ID. Lower, because, as a result of testing, I've determined that the official Datadog agent does the same when encountering a 128 bit trace ID in a B3 header, and you want Kamon to be compatible with a mixed environment using the Datadog agent as well.

dvgica:
Fixes #1032.

### DIFF
--- a/reporters/kamon-datadog/src/main/scala/kamon/datadog/DatadogSpanReporter.scala
+++ b/reporters/kamon-datadog/src/main/scala/kamon/datadog/DatadogSpanReporter.scala
@@ -36,7 +36,9 @@ trait KamonDataDogTranslator {
 
 object KamonDataDogTranslatorDefault extends KamonDataDogTranslator {
   def translate(span: Span.Finished, additionalTags: TagSet, tagFilter: Filter): DdSpan = {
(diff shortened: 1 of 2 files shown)


dvgica:
It seems like the test failure is unrelated to my change, but let me know if not.

PR #1042, opened 2021-06-15
try to bind Status Page on a different port if the configured port is busy

ivantopo:
This ensures that no errors will be thrown when users start more than one Kamon-instrumented application on the same host and network.


SimunKaracic:
Tried and tested, works great

Issue #986, opened 2021-04-07
W3C propagates the wrong parent spanID


pnerg:
Looks like the W3C context propagation uses a new spanID as the parent.

Did a small test app that received a request and forwarded it again to itself just to see what headers are propagated.
So the flow is `postman -> app (/forward) -> app (/hello)`  and response all the way back in the chain

## Test 1 - Sent _traceparent_ header to app
Sent the header `traceparent: Some(00-bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb-0000000000000000-01)`
The printout is like
```
Got request for '/forward' with w3c header [Some(00-bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb-0000000000000000-01)] and parsed/generated context traceId[bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb], spanId[b4e99c809b29b7fa], parentSpanId[0000000000000000]
Got request for '/hello' with w3c header [Some(00-BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB-A4B5DFB07D82F430-01)] and parsed/generated context traceId[], spanId[230e25c602c0b2bc], parentSpanId[]
```
We see from the printout that the W3C propagation has read the headers and created a new local spanID (b4e99c809b29b7fa) as expected.
Trace and parent span identifiers are set in the local context.  
However when the request is sent to the next part (/hello) we can see that the header is an odd parentSpanID (A4B5DFB07D82F430). This should have been 'b4e99c809b29b7fa'

The "empty" trace/span ids in the log is due to #985

## Test 2 - No headers are sent to the app
The printout is like
```
Got request for '/forward' with w3c header [None] and parsed/generated context traceId[1aaf4ba3617d76f2441f0fc3879a1a3f], spanId[b16fd11b8393437c], parentSpanId[]
Got request for '/hello' with w3c header [Some(00-1AAF4BA3617D76F2441F0FC3879A1A3F-45B2C2A25AE39FB3-00)] and parsed/generated context traceId[], spanId[a547524af1953855], parentSpanId[]
```
Same situation, we see a new context is created with trace/span identifiers. But again when the context is propagated we see the sent parentId is '45B2C2A25AE39FB3' but the span that sent the request was 'b16fd11b8393437c'

## Conclusion
I'm assuming a new spanID has been created *before* passing the context down to the 
[...]

PR #1045, opened 2021-06-21
W3C span propagation id fix

SimunKaracic:
Propagate the current span id, not the parent span id, according to the spec. 

When reading from headers, don't generate a new ID for the Remote span, but use the existing one.

This is the thread where @dguggemos reported the bug -> https://gitter.im/kamon-io/Kamon?at=60cc4e2cb31731135437e0aa

@pnerg If you have the time, check this out. You initially changed this to use the parentID instead of the spanID. According to the W3C spec, the current span ID should be propagated. 
From the spec https://www.w3.org/TR/trace-context/#a-traceparent-is-received: 
```Update parent-id: The value of property parent-id MUST be set to a value representing the ID of the current operation.```

https://github.com/kamon-io/Kamon/issues/986 I believe this issue originated from the fact that akka-http was being double instrumented.

### DIFF
--- a/core/kamon-core-tests/src/test/scala/kamon/trace/W3CTraceContextSpanPropagationSpec.scala
+++ b/core/kamon-core-tests/src/test/scala/kamon/trace/W3CTraceContextSpanPropagationSpec.scala
@@ -14,7 +14,7 @@ class W3CTraceContextSpanPropagationSpec extends WordSpecLike with Matchers with
       val headersMap = mutable.Map.empty[String, String]
       traceContextPropagation.write(testContext(), headerWriterFromMap(headersMap))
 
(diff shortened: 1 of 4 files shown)


the-overengineer:
We talked about this a bit offline, looks good to me after looking at the spec. Please make sure this doesn't do something terribly wrong to Zipkin (or we at least need to check whether we need to adjust the joining of spans flag in the config/document it), though. The `Span.Remote` is just a _reference_ with the same ID, so this should work, but issues might arise due to only one of the sides having the correct parent span ID.

the-overengineer (approved):

SimunKaracic:
There's only one issue. When using w3c, we need to make sure that `kamon.trace.identifier-scheme` is set to `double`. This is already documented in the configuration, but we should probably warn if it's set to `single`. I'll try and get it done now, but if it's too complicated I'll create a task for it for *_the future_*

PR #1053, opened 2021-07-12
Change adaptive sample groups to use Globs

SimunKaracic:


core/kamon-core/src/main/scala/kamon/trace/AdaptiveSampler.scala
```diff
@@ -275,7 +274,7 @@ object AdaptiveSampler {
 
         Settings.Group(
           name = groupName,
-          operations = groupConfig.getStringList("operations").asScala.toSeq,
+          operations = groupConfig.getStringList("operations").asScala.toSeq.map(Filter.fromGlob),
```
the-overengineer, line 277: Don't like this, for two reasons:
1. We should allow people to use regular expressions, if they want to
2. This is not backwards compatible! We know some people have already set it up to raw strings.

What I suggest we do is:
1. Copy logic from https://github.com/kamon-io/Kamon/blob/4b397cbc242aaebbdc5dff5e28eee6cf0b5fac4d/core/kamon-core/src/main/scala/kamon/util/Filter.scala#L106
2. Change it so that the default is **not** glob, but it's exact match (a new type of filter, I guess, `Filter.StringMatch` or something)
3. _Immediately_ announce exact match as deprecated
4. Move to default-to-glob in the next minor version (2.3.0, I guess)

Thoughts, @SimunKaracic , @ivantopo?

the-overengineer (changes requested):

core/kamon-core/src/main/scala/kamon/trace/AdaptiveSampler.scala
SimunKaracic: Assuming you mean this under 1. https://github.com/kamon-io/Kamon/blob/4b397cbc242aaebbdc5dff5e28eee6cf0b5fac4d/core/kamon-core/src/main/scala/kamon/util/Filter.scala#L121

That all sounds like a great idea

PR #1057, opened 2021-09-27
Scala 3 Initial support

dpsoft:
* update scalatest to 3.2.9
* update to scala 2.13.6
* bump kanela to 1.0.12


ivantopo:
Rebased from master and pushed. All tests passed locally, let's see what GA says :smile:

PR #1063, opened 2021-10-18
Reproducer: Span operation name duplicated when more than 1 endpoint defined per route

seglo:
I'm using `sbt-guardrail` to generate akka-http `Routes` from an OpenAPI spec. I'm instrumenting these endpoints with Kamon's akka-http integration. I noticed that the way `sbt-guardrail` generates the akka-http `Route` sometimes causes an issue with how operation names are generated for spans. Guardrail will define `Routes` by their `path` and then define the HTTP verb as a nested `Directive`. This appears to cause confusion with how paths are matched in the akka-http/Kamon akka-http server instrumentation code.

i.e.)

```scala
val route = 
      path("some_endpoint") {
        post(complete(OK))
      } ~
      path("some_endpoint") {
        get(complete(OK))
      }
```

Reproducer test fails with:

```
[info] AkkaHttpServerTracingSpec:
[info] the Akka HTTP server instrumentation with HTTP/1
[info] - should return the correct operation name with overloaded route *** FAILED ***
[info]   "/some_endpoint[/some_endpoint]" was not equal to "/some_endpoint[]" (AkkaHttpServerTracingSpec.scala:89)
```

### DIFF
--- a/instrumentation/kamon-akka-http/src/main/scala-2.11/kamon/instrumentation/akka/http/AkkaHttpServerInstrumentation.scala
+++ b/instrumentation/kamon-akka-http/src/main/scala-2.11/kamon/instrumentation/akka/http/AkkaHttpServerInstrumentation.scala
@@ -1,7 +1,6 @@
 package kamon.instrumentation.akka.http
 
 import java.util.concurrent.Callable
-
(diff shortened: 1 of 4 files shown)


seglo:
Excellent. TYVM @ivantopo!

PR #1071, opened 2021-11-04
Fix some typos

Philippus:
Applying this PR will fix some typos I came across when looking at the reference.conf file.


ivantopo:
Thanks a lot @Philippus, I'm actually surprised by how many times we (probably me) made the same mistake :man_facepalming:

PR #1075, opened 2021-11-11
Add jvm threads states metrics

getArtemUsername:
Hi, I think that states of threads could be usefull for gross analyzing in cloud environments such as k8s or similar, for example for fine tuning pools or understanding what happens while cpu throttling.

### DIFF
--- a/instrumentation/kamon-system-metrics/src/main/scala/kamon/instrumentation/system/jvm/JvmMetrics.scala
+++ b/instrumentation/kamon-system-metrics/src/main/scala/kamon/instrumentation/system/jvm/JvmMetrics.scala
@@ -17,10 +17,9 @@
 package kamon.instrumentation.system.jvm
 
 import java.lang.management.ManagementFactory
-
 import kamon.Kamon
 import kamon.instrumentation.system.jvm.JvmMetrics.MemoryUsageInstruments.{BufferPoolInstruments, MemoryRegionInstruments}
-import kamon.instrumentation.system.jvm.JvmMetricsCollector.{Collector, MemoryPool}
+import kamon.instrumentation.system.jvm.JvmMetricsCollector.{Collector, MemoryPool, ThreadState}
 import kamon.metric.{Gauge, Histogram, InstrumentGroup, MeasurementUnit}
 import kamon.tag.TagSet
 
@@ -107,6 +106,11 @@ object JvmMetrics {
   val ThreadsDaemon = Kamon.gauge(
     name = "jvm.threads.daemon",
     description = "Tracks the current number of daemon threads on the JVM"
+  )
+  
+  val ThreadsStates = Kamon.gauge(
+    name = "jvm.threads.states",
+    description = "Tracks the current number of threads states"
   )
 
   val ClassesLoaded = Kamon.gauge(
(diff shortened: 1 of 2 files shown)


instrumentation/kamon-system-metrics/src/main/scala/kamon/instrumentation/system/jvm/JvmMetricsCollector.scala
```diff
@@ -259,4 +268,27 @@ object JvmMetricsCollector {
...
       _invalidChars.replaceAllIn(name.toLowerCase, "-")
   }
+  
+  sealed trait ThreadState
+  object ThreadState {
+    case object New extends ThreadState { override def toString: String = "new"}
+    case object Runnable extends ThreadState { override def toString: String = "runnable" }
+    case object Blocked extends ThreadState { override def toString: String = "blocked" }
+    case object Waiting extends ThreadState { override def toString: String = "waiting" }
+    case object TimedWaiting extends ThreadState { override def toString: String = "timed_waiting" }
```
ivantopo, line 278: Can you please change `timed_waiting` with `timed-waiting`. This way the metric would match exactly the same output as the corresponding metric in Micrometer :smile:

instrumentation/kamon-system-metrics/src/main/scala/kamon/instrumentation/system/jvm/JvmMetrics.scala
```diff
@@ -108,6 +107,11 @@ object JvmMetrics {
     name = "jvm.threads.daemon",
     description = "Tracks the current number of daemon threads on the JVM"
   )
+  
+  val ThreadsStates = Kamon.gauge(
+    name = "jvm.threads.states",
+    description = "Tracks the current number of threads states"
```
ivantopo, line 113:
```suggestion
    description = "Tracks the current number of threads on each possible state"
```

ivantopo (changes requested):
Hey @getArtemUsername, thanks a lot for this addition! The metric and logic looks all fine, there is only a little change on names and descriptions. I'll merge this as soon as those changes are in :smile:

instrumentation/kamon-system-metrics/src/main/scala/kamon/instrumentation/system/jvm/JvmMetricsCollector.scala
getArtemUsername: Thanks for review. I've changed it.

instrumentation/kamon-system-metrics/src/main/scala/kamon/instrumentation/system/jvm/JvmMetricsCollector.scala
```diff
@@ -259,4 +268,27 @@ object JvmMetricsCollector {
...
+    def find(state: String): ThreadState =
+      _threadStateMapping.getOrElse(state, Unknown)
+    
+    private val _threadStateMapping: Map[String, ThreadState] = Map(
+      "NEW"           -> New,
+      "RUNNABLE"      -> Runnable,
+      "BLOCKED"       -> Blocked,
+      "WAITING"       -> Waiting,
+      "TIMED_WAITING" -> TimedWaiting,
+      "TERMINATED"    -> Terminated,
```
ivantopo, line 291: Please remove this trailing comma, otherwise it will not compile on Scala 2.11

ivantopo (changes requested):
There is a tiny issue with compiling on Scala 2.11, I think removing the trailing comma should be enough.

instrumentation/kamon-system-metrics/src/main/scala/kamon/instrumentation/system/jvm/JvmMetricsCollector.scala
getArtemUsername: done :smile:

getArtemUsername:
@ivantopo Do I need to do something else to be merged? Looks like it's not in 2.4.0 :thinking:

ivantopo:
Oh man, I even included it in the release notes. I'm sorry. In my head this was merged already :sob: 

I'll release 2.4.1 right after, including this PR.

getArtemUsername:
Thanks :smile:

PR #1076, opened 2021-11-12
Kamon-redis: make it work for jedis 2.x

jtjeferreira:
In Jedis 2.x the type is `redis.clients.jedis.Protocol.Command`
while in Jedis 3.x is `redis.clients.jedis.commands.ProtocolCommand`
(and `redis.clients.jedis.Protocol.Command` is the only subclass of `ProtocolCommand`)

However we dont need to match on it since we just use toString...


jtjeferreira:
I was getting double the spans like this
image

0b23492 fixes it

ivantopo:
Hey @jtjeferreira, so this will work for multiple Jedis versions?

jtjeferreira:
> Hey @jtjeferreira, so this will work for multiple Jedis versions?

for jedis 2.x and 3.x. However the tests only test it for 3.x

Issue #1077, opened 2021-11-12
Any reason why otel TraceService is package private?


MaciejSzewczyszyn:
The trait and implementation are defined as:
```
private[otel] trait TraceService extends Closeable { ... }
private[otel] object GrpcTraceService { ... }
private[otel] class GrpcTraceService(channel:ManagedChannel, traceService:TraceServiceFutureStub) extends TraceService { ... }
```
While the reporter itself is public:
```
class OpenTelemetryTraceReporter(traceServiceFactory:Config=>TraceService)(implicit ec:ExecutionContext) extends SpanReporter { ... }
```

In my app, I need to override `reportSpans` to filter out spans for certain urls (healthchecks).  The package private modifier on the `TraceService` makes it unnecessarily difficult as I can't easily extend `OpenTelemetryTraceReporter` - can the trace service be public or am I missing something?

ivantopo:
Hey @MaciejSzewczyszyn, I don't see any particular reason for those objects to be package-private. I guess it is just how it got implemented. That said, if you want to disable tracing for particular URLs that's something you can do via configuration alone! Try adding something like this to your configuration:

```
kamon.trace.adaptive-sampler.groups {
  health-checks {
    operations = [ "/health" ]
    rules {
      sample = never
    }
  }
}
```

Just make sure that the operation name matches exactly the Span operation name. Does this solve the problem you are having?

MaciejSzewczyszyn:
Thanks @ivantopo. The traces are still being sent though - this is how this bit is currently configured:
```
kamon {
  (...)
  trace {
    span-metrics {
      scope-spans-to-parent = yes
    }

    join-remote-parents-with-same-span-id = true
    identifier-scheme = "double"
    sampler = "always"

    adaptive-sampler {
      groups {
        health-checks {
          operations = [ "/app_status" ]
          rules {
            sample = never
          }
        }
      }
    }
  }
}
```
We're on 2.2.0. Am I doing something wrong? Is there a doc reference for this specific part that I could take a look at?

ivantopo:
There is not much documentation about sampling right now, hopefully I'll update it these days since we are releasing a new local tail-based sampler feature in Kamon!

In the meantime, the issue here is that you are not using the adaptive sampler but rather the "always" sampler with that `sampler = "always"` setting. If you revert it to `sampler = "adaptive"` then the rule you defined should work as expected.

MaciejSzewczyszyn:
[...]

PR #1078, opened 2021-11-15
+ Removing private package modifier from the otel trace service

MaciejSzewczyszyn:
PR to address https://github.com/kamon-io/Kamon/issues/1077


ivantopo:
Hey @MaciejSzewczyszyn, I'm almost sure that [this answer](https://github.com/kamon-io/Kamon/issues/1077#issuecomment-969338942) should make this PR unnecessary. Let me know if that worked for you!

PR #1079, opened 2021-11-15
Make it easier to expose Prometheus metrics on custom endpints

ivantopo:
It often happens that people want to expose the Prometheus scrape data on their own HTTP server and the only way to do it so far is to:
1. Prevent Kamon from auto-starting the Prometheus reporter via settings with `kamon.modules.prometheus-reporter.enabled=false`
2. Create an instance of the reporter programmatically and add it to Kamon via `Kamon.addReporter(...)`, and keep a reference to the reporter
3. Use the reference to call `prometheusReporter.scrapeData()` and return that on the HTTP server

It is kind of annoying. Since we know that in most cases there is only one PrometheusReporter instance created automatically by Kamon, this PR ensures that we keep a reference to that instance and expose it's scrape data via `PrometheusReporter.latestScrapeData()`, so that users can skip all the above when they need to.

### DIFF
--- a/reporters/kamon-prometheus/src/main/scala/kamon/prometheus/PrometheusReporter.scala
+++ b/reporters/kamon-prometheus/src/main/scala/kamon/prometheus/PrometheusReporter.scala
@@ -45,8 +45,17 @@ class PrometheusReporter(configPath: String = DefaultConfigPath, initialConfig: 
     startEmbeddedServerIfEnabled()
   }
 
-  override def stop(): Unit =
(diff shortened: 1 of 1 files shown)


ivantopo:
Let me modify this. Now that I think about it, there is no reason for users to access the entire PrometheusReporter instance, since what they want all the time is the scrape data. Let's expose that alone.

ivantopo:
Validated this locally. Works fine.

PR #1080, opened 2021-11-16
add automatic tail sampling based on latency and error count thresholds

ivantopo:
I originally shared this idea on our [Discord server](https://discord.com/channels/866301994074243132/866652723810140191/909891800938999810) and I'm copy/pasting it here:

> There is an idea that came to mind after chatting with one of our users and related to trying to keep "interesting" traces only.. where "interesting" means that they are either high latency or contain errors. We already got a span-reporting-delay setting (see https://github.com/kamon-io/Kamon/blob/master/core/kamon-core/src/main/resources/reference.conf#L148-L156) a few versions ago that can keep Spans in memory for a few extra seconds, so that users can add a bit of extra logic in their controllers and decide whether to keep or drop a trace. The issue there is that users MUST add some extra logic by hand.
> 
> I would like to add three things here:
> - An error counter on our Trace object. This counter will be incremented when any Span for a trace is marked as failed. 
> - A small piece of logic that runs after Spans finish and figures out whether the trace should be marked for keeping or not. 
> - A configuration format that allows users to set this up with configuration only. Something like saying "try to keep all traces with at least 2 errors, or with latency of 1+ seconds"
> 
> For starters I think we can apply this logic only when Local Root spans are finished, and then we can try to expand this with a bit more functionality, like setting different rules for different operations, or maybe even trying to keep traces for different latency buckets. Many ideas come to mind but I think it would be good to keep it simple and see where it goes 😄 
> 
> This would work really well for people using Kamon for monolith-like applications, but the "forced" traces would be partial if you are doing distributed traces. Still, I think it is better to have a partial trace than no trace at all. If you have the trace and trace ID maybe you can gather some extra data from logs if they are correlated. And, not everybody is doing microservices!

This PR is implementing the idea from above, allowing to force keep traces with a minimum error count or latency threshold. I still need to write some tests and ensure everything works fine. Fortunately the implementation is pretty simple.

### DIFF
--- a/core/kamon-core/src/main/resources/reference.conf
+++ b/core/kamon-core/src/main/resources/reference.conf
@@ -241,6 +241,30 @@ kamon {
       }
     }
 
+    # Automatically change the sampling decision for a trace based on the either the number of errors in a trace, or a
+    # fixed latency threshold. The local tail sampler will only work when the `span-reporting-delay` setting is set to a
+    # value greater than zero.
+    #
+    # Important things to keep in mind:
+    #   - The sampling decision override happens only in the current process. This means that any requests sent to other
+    #     services before the sampling decision override happened might be missing from the resulting trace.
+
+    #   - Try to set `span-reporting-delay` to a value slightly higher than your application's timeout to ensure all the
+    #     local spans related to the trace will still be kept in memory when the sampling override happens
+    #
+    local-tail-sampler {
+
+      # Turns the local tail sampler logic on or off.
+      enabled = no
+
+      # Overrides the sampling decision to "Sample" when a trace contains at least this number of errors when the local
+      # Root Span finishes processing.
+      error-count-threshold = 1
+
+      # Overrides the sampling decision to "Sample" when the local Root Span took more than this time to complete.
+      latency-threshold = 1 second
+    }
+
     # Settings that influence the tags applied to the "span.processing-time" metric for all finished spans with metric
     # tracking enabled.
     #
--- a/core/kamon-core/src/main/scala/kamon/trace/Trace.scala
+++ b/core/kamon-core/src/main/scala/kamon/trace/Trace.scala
@@ -17,6 +17,8 @@
 package kamon
 package trace
 
+import java.util.concurrent.atomic.AtomicInteger
+
 /**
   * Holds information shared across all Spans from the same Trace. It might seem like too little information but all in
   * all, a trace is just a bunch of Spans that share the same trace identifier ;).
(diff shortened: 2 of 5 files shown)


core/kamon-core/src/main/scala/kamon/trace/Trace.scala
```diff
@@ -56,6 +58,17 @@ trait Trace {
...
+  /**
+    * Signals that a Span belonging to this trace has failed.
+    */
+  def spanFailed(): Unit
+
+  /**
+    * Returns the number of failed Spans for this trace in this process. This error count does not reflect errors that
+    * might have happened on other services participating in the same trace.
+    */
+  def failedSpanCount(): Int
```
ihostage, line 70: WDYT, Ivan? 🤔 
```suggestion
  def failedSpansCount(): Int
```

ivantopo: Thanks @ihostage, got it done and pushed!

ivantopo:
Got the tests in and rebased from master. If nothing special comes up I'll merge this tomorrow and put out a minor release.

core/kamon-core/src/main/resources/reference.conf
```diff
@@ -241,6 +241,29 @@ kamon {
...
+    # fixed latency threshold. The tail sampler will only work when the `span-reporting-delay` setting is set to a value
+    # greater than zero.
+    #
+    # Important things to keep in mind:
+    #   - The sampling decision override happens only in the current process. This means that any requests sent to other
+    #     services before the sampling decision override happened might be missing from the resulting trace.
+    #   - Try to set `span-reporting-delay` to a value slightly higher than your application's timeout to ensure all the
+    #     local spans related to the trace will still be kept in memory.
+    #
+    tail-sampler {
```
ivantopo, line 254: I'm wondering if we should call this `local-tail-sampler` or maybe `sampling-decision-overwrite`. I guess tail sampler is going to be easier to find because it is the industry name for this kind of behavior, but I would like to emphasize that this is a local-only feature.

core/kamon-core/src/main/resources/reference.conf
```diff
@@ -241,6 +241,29 @@ kamon {
...
+
+      # Turns the tail sampler logic on or off.
+      enabled = no
```
ivantopo, line 257: We should add a warning log if users enabled tail sampling but didn't set a `span-reporting-delay` value. The features are independent, but it is very unlikely to catch full traces without the reporting delay.

core/kamon-core/src/main/resources/reference.conf
dpsoft: +1 to `local-tail-sampler` instead of  `tail-sampler`

dpsoft:
@ivantopo awesome! and the most important, is really simple and useful

core/kamon-core/src/main/resources/reference.conf
ivantopo: Got it done.

Issue #238, opened 2015-07-30
Finagle tracing


benjumanji:
Hi!

I'd really like to use kamon with some of the twitter stack tools. I have taken a quick look at the tracing docs and I think to add finagle tracing info to kamon all I really need to do is find a way to convert finagle trace records into TraceInfo objects, and stick them _somewhere_. I can see how I can subscribe to tracing info, I just don't think there is a hook yet where I can feed trace info into kamon?

Would this be something that you were willing to support?

ivantopo:
Hey Benjamin!

Can you explain in a bit more detail what you want to achieve? I'm not very familiar with Finagle, but as far as I know it's tracing capabilities are aligned with the capabilities of Zipkin which goes beyond what we currently have in Kamon... what we do now can be considered as a very lightweight tracing that lacks many of the things you can get from Finagle/Zipkin, like real distributed tracing and preserving causal relation between segments. 

Improving the tracing capabilities to get to those levels of detail is something we plan to do, but I'm not sure of how long will it take to do so. In the meantime, you can explore the awesome work [these guys] are doing and maybe even give it a try!

If you want to explore integrating Kamon with Finagle yourself I would recommend starting with the very basic stuff like getting some metrics out of it and then build on top of that to get more complex features working. We can certainly provide as much help as we can to build such a module that will certainly be a good thing for the project, but at the moment we don't have the capacity to support such module ourselves. Hope you decide to give it a try and share your work with the community!



[these guys]: https://github.com/ebuzzing/Kamon/tree/zipkin/kamon-zipkin

benjumanji:
So effectively with a finagle client you have a stack of modules, one of these handles tracing. You can insert your own tracer, and the trait is effectively `trait Tracer(trace: Record)`, where `Record` looks something like a `TraceInfo`. So I was thinking, if you have cool plans to make cool reporting / dashboards or whatnot out of your trace info, it would be nice if I could make a finagle tracer that kicked out data into a kamon collector. In theory this should be that hard (he says). I ask basically because zipkin OSS seems to have crumpled a little of late, though I do note this project: http://openzipkin.github.io/zipkin/ has sprung up. I'll have a look at the project you referenced.

benjumanji:
It should be noted that zipkin and tracing in finagle are not the same thing. The zipkin project really covers the collection and display of the tracing, but finagle will merrily allow you to send it's trace records anywhere your heart desires.

jtjeferreira:
[...]

PR #1096, opened 2021-12-30
Add finagle client instrumentation

seglo:
Instrumentation to manage Kamon spans for finagle client request & response lifecycle.

Relates to #238

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -547,7 +547,7 @@ lazy val `kamon-finagle` = (project in file("instrumentation/kamon-finagle"))
     libraryDependencies ++= Seq(
       kanelaAgent % "provided",
       "com.twitter" %% "finagle-http" % "21.12.0" % "provided",
-      "com.twitter" %% "bijection-util" % "0.9.5" % "provided",
+      "com.twitter" %% "bijection-util" % "0.9.5" % "test",
       scalatest % "test",
       logbackClassic % "test",
     )
--- a/build.sbt
+++ b/build.sbt
@@ -539,6 +539,20 @@ lazy val `kamon-caffeine` = (project in file("instrumentation/kamon-caffeine"))
     )
   ).dependsOn(`kamon-core`, `kamon-testkit` % "test")
 
+lazy val `kamon-finagle` = (project in file("instrumentation/kamon-finagle"))
+  .disablePlugins(AssemblyPlugin)
+  .enablePlugins(JavaAgent)
+  .settings(instrumentationSettings)
+  .settings(
+    libraryDependencies ++= Seq(
+      kanelaAgent % "provided",
+      "com.twitter" %% "finagle-http" % "21.12.0" % "provided",
+      "com.twitter" %% "bijection-util" % "0.9.5" % "provided",
+      scalatest % "test",
+      logbackClassic % "test",
+    )
+  ).dependsOn(`kamon-core`, `kamon-instrumentation-common`, `kamon-testkit` % "test")
+
 /**
  * Reporters
  */
--- a/build.sbt
+++ b/build.sbt
@@ -544,6 +544,7 @@ lazy val `kamon-finagle` = (project in file("instrumentation/kamon-finagle"))
   .enablePlugins(JavaAgent)
   .settings(instrumentationSettings)
   .settings(
+    crossScalaVersions := Seq("2.12.11", "2.13.1"),
     libraryDependencies ++= Seq(
       kanelaAgent % "provided",
       "com.twitter" %% "finagle-http" % "21.12.0" % "provided",
@@ -811,6 +812,7 @@ val `kamon-bundle` = (project in file("bundle/kamon-bundle"))
     `kamon-redis` % "shaded",
     `kamon-okhttp` % "shaded",
     `kamon-caffeine` % "shaded",
+    `kamon-finagle` % "shaded",
 )
 
 lazy val `bill-of-materials` = (project in file("bill-of-materials"))
--- a/build.sbt
+++ b/build.sbt
@@ -549,7 +549,7 @@ lazy val `kamon-finagle` = (project in file("instrumentation/kamon-finagle"))
     libraryDependencies ++= Seq(
       kanelaAgent % "provided",
       "com.twitter" %% "finagle-http" % "21.12.0" % "provided",
-      "com.twitter" %% "bijection-util" % "0.9.5" % "test",
+      "com.twitter" %% "bijection-util" % "0.9.7" % "test",
       scalatest % "test",
       logbackClassic % "test",
     )
--- a/build.sbt
+++ b/build.sbt
@@ -140,6 +140,7 @@ val instrumentationProjects = Seq[ProjectReference](
   `kamon-tapir`,
   `kamon-redis`,
   `kamon-caffeine`,
+  `kamon-finagle`,
 )
 
 lazy val instrumentation = (project in file("instrumentation"))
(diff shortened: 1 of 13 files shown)


seglo:
I think the project is referenced in the right places of the `build.sbt`, but I don't see its test results in the CI build.

ivantopo:
Hey @seglo, awesome that you are contributing this instrumentation! I'm going to be away until Monday, then I'll take a proper look :eyes:

ivantopo:
Hey @seglo, I'm on it now :smile: 

Got covid last week and it really kept me busy, but now I'm back. Here are a few questions that came up during my first look at it:
  1. Is there any particular reason why only the client side is instrumented? We can definitely start just with the client, but I was wondering if there is anything big preventing from bringing full support.
  2. Is there anything preventing us from adding the Kamon filter automatically via automatic instrumentation? so that people don't need to do anything by hand to get tracing.

build.sbt
```diff
@@ -539,6 +540,21 @@ lazy val `kamon-caffeine` = (project in file("instrumentation/kamon-caffeine"))
     )
   ).dependsOn(`kamon-core`, `kamon-testkit` % "test")
 
+lazy val `kamon-finagle` = (project in file("instrumentation/kamon-finagle"))
```
ivantopo, line 543: Please run `headerCreate` on the project to get all the license-related headers.

instrumentation/kamon-finagle/src/main/scala/kamon/instrumentation/finagle/client/Tags.scala
```diff
@@ -0,0 +1,63 @@
...
+    val ClientSpanKind = "client"
+    val HttpSpanType = "http"
+  }
+
+  import Keys._
+  import Values._
+
+  def setHttpClientTags(span: Span): Unit = {
+    span.tag(SpanKind, ClientSpanKind)
+    span.tag(SpanType, HttpSpanType)
```
ivantopo, line 37: It shouldn't be necessary to add the span kind as a tag. The kind has its own field on the spans and it is set automatically by the HTTP Client instrumentation, and then the reporters turn it into a tag when the backend needs it. Here I don't think it is necessary.

instrumentation/kamon-finagle/src/main/scala/kamon/instrumentation/finagle/client/Tags.scala
```diff
@@ -0,0 +1,63 @@
...
+import java.net.InetSocketAddress
+
+/**
+ * Unique tags and value constants added to Kamon Finagle spans.
+ *
+ * As a general rule metric tags should only be for tags with low cardinality.
+ */
+private[finagle] object Tags {
+
+  object Keys {
```
ivantopo, line 16: There are a few tags that I don't recall using in the past, could you please share what made you include them? I'm talking about:
  - `span.type`
  - `error.type`
  - `http.status_category`

instrumentation/kamon-finagle/src/main/scala/kamon/instrumentation/finagle/client/Tags.scala
seglo: Got it. Yes, I see `span.kind` being added to tags.

The span type is an interesting case. We use the `kamon-datadog` span reporter and I noticed that the `DDSpan` that's serialized and sent to DataDog hardcodes its span kind field to `custom`. [There are a handful of supported options](https://docs.datadoghq.com/tracing/visualization/services_list/#services-types). They have subtle effects on how the APM frontends are rendered. We've since overridden this by implementing our own `KamonDataDogTranslator` to send it as `web`. I was thinking about opening a Kamon issue for this to make it more configurable, but our solution seems to work fine for now.

I'll remove this tagging from the implementation. I can sort out what our infrastructure needs outside of this instrumentation.

instrumentation/kamon-finagle/src/main/scala/kamon/instrumentation/finagle/client/Tags.scala
seglo: I addressed `span.type` in the other comment.

`error.type` this is something we use internally, but now that I think about it it's something we can instead define as a `mark` when errors occur for certain finagle operation events, so I'll make that change and drop the tag.

`http.status_category` another internal concern I'll remove from this implementation.

seglo:
Thanks for the review @ivantopo !

> Got covid last week and it really kept me busy, but now I'm back. Here are a few questions that came up during my first look at it:

Oh my goodness. I hope you're recovering well!

> 
>     1. Is there any particular reason why only the client side is instrumented? We can definitely start just with the client, but I was wondering if there is anything big preventing from bringing full support.

My team uses Akka for most of our needs, which already has good coverage by Kamon. However, much of the rest of our infrastructure uses finagle http and gRPC services that we integrate with. Those services have a separate custom Open API finagle tracing instrumentation and they use DataDog libraries instead of Kamon.

A server implementation is totally do-able, but I'll have to discuss it with the team to see if we can spare the time to work on it since we don't have a ready made solution. That said, many parts of the implementation required for adding server support can be reused from what's already in this PR so it may not be too much work and something I can tackle in my free time.

> 
>     2. Is there anything preventing us from adding the Kamon filter automatically via automatic instrumentation? so that people don't need to do anything by hand to get tracing.

It may be possible to set a default Finagle tracer. I'll have to do some digging.

ivantopo:
Hey @seglo, never mind about the server part then. I thought that you were using the server as well but maybe it was complicated to instrument and I could help you push it forward :smile:, if you are not really using it then better to wait until someone using the server-side of Finagle comes along and wants to give it a try.

instrumentation/kamon-finagle/src/main/scala/kamon/instrumentation/finagle/client/Tags.scala
ivantopo: Great! please ping me when you add those changes to look at this again.

instrumentation/kamon-finagle/src/main/scala/kamon/instrumentation/finagle/client/Tags.scala
ivantopo: I think that this is something we definitely should address in the datadog reporter itself. It wouldn't be too complicated to check if a span has `SpanKind=Client|Server` and some `http.*` tag, we would automatically add the span.type tag there so it works nicely in Datadog's UI. 

Also, I remember an old conversation about showing DB spans properly in Datadog. It might be a similar thing we need to do. Let's track those changes in a separate issue/pr: #1098

ivantopo:
Hey @seglo, I'm really lacking on any Finagle-specific knowledge to comment on that part of the work, but Kamon-wise this seems good to go. Do you have any additional comments or changes to include before merging?

seglo:
> better to wait until someone using the server-side of Finagle comes along and wants to give it a try.

Agreed! If the situation changes and we need to maintain one of our finagle services then I might follow up then if someone else hasn't already implemented server-side support.

> Hey @seglo, I'm really lacking on any Finagle-specific knowledge to comment on that part of the work, but Kamon-wise this seems good to go. Do you have any additional comments or changes to include before merging?

@ivantopo I think it's ready to go, but I'll update with master and do one more pass first. Are you cutting a release soon? I can probably get to this later today.

ivantopo:
I have one more tiny change to include and then ship a release, so probably it will all go out tomorrow morning if you manage to do those last checks today. And if it doesn't go tomorrow we can do another release with this PR alone!

seglo:
@ivantopo Ok. IMO it's good to go!

PR #1097, opened 2022-01-07
Update version of typesafe config to latest

ryanb93:


dpsoft:
@ryanb93 thanks for proposing this change!

PR #1106, opened 2022-02-09
Http protobuf for otel take 2

hughsimpson:
Alternative implementation of opentelemetry http/protobuf reporting; includes more of the standard configuration options from https://opentelemetry.io/docs/reference/specification/protocol/exporter/#specifying-headers-via-environment-variables than were previously supported


ivantopo:
I'm closing this PR in favor of #1102.

PR #1112, opened 2022-02-17
add error.type to traces. Includes feature flag (on by default)

hughsimpson:
This is for some extra compatibility with opentelemetry libs. The tests fail locally with some OOMs, which is odd to me but I guess if the limits are tightly scoped, then adding an extra boolean field to the traces might explain it...

### DIFF
--- a/core/kamon-core/src/main/resources/reference.conf
+++ b/core/kamon-core/src/main/resources/reference.conf
@@ -178,7 +178,7 @@ kamon {
 
     # Decides whether to include the class name of a Throwable as the "error.type" Span tag when a Span is marked
     # as failed.
-    include-error-type = no
+    include-error-type = yes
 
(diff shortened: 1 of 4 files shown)


core/kamon-core/src/main/resources/reference.conf
```diff
@@ -176,6 +176,10 @@ kamon {
     # as failed.
     include-error-stacktrace = yes
 
+    # Decides whether to include the class name of a Throwable as the "error.type" Span tag when a Span is marked
+    # as failed.
+    include-error-type = no
```
ivantopo, line 181: Why not have it on by default? It feels like a good thing to have for filtering failed spans

ivantopo (changes requested):
Thanks for this addition, @hughsimpson! Just a tiny config defaults change and we are good to go :smile:

PR #1113, opened 2022-02-17
WIP: Moved kamon-netty

LeonDaniel:


instrumentation/kamon-netty/src/main/scala/kamon/netty/Netty.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  private def nameGeneratorFromConfig(config: Config): HttpOperationNameGenerator = {
+    val dynamic = new DynamicAccess(getClass.getClassLoader)
+    val nameGeneratorFQCN = config.getString("kamon.instrumentation.netty.client.tracing.operations.name-generator")
+    dynamic.createInstanceFor[HttpOperationNameGenerator](nameGeneratorFQCN, Nil)
+  }
+
+  Kamon.onReconfigure { newConfig =>
+    nameGenerator = nameGeneratorFromConfig(newConfig)
+  }
```
ivantopo, line 65: I think none of this code is required. The `HttpServerInstrumentation` class is already able to generate operation names and accept custom name generators so this should probably be gone (couldn't find any evidence that it is actually being used)

instrumentation/kamon-netty/src/main/scala/kamon/netty/Netty.scala
```diff
@@ -0,0 +1,94 @@
...
+}
+
+class DefaultNameGenerator extends HttpOperationNameGenerator {
```
ivantopo, line 68: Remove this class for the same reasons as mentioned above

instrumentation/kamon-netty/src/main/scala/kamon/netty/PathOperationNameGenerator.scala
```diff
@@ -0,0 +1,8 @@
+package kamon.netty
```
ivantopo, line 1: Remove this file

instrumentation/kamon-netty/src/main/scala/kamon/netty/instrumentation/HttpServerInstrumentation.scala
```diff
@@ -0,0 +1,57 @@
...
+
+class HttpServerInstrumentation extends InstrumentationBuilder {
+
+  onSubTypesOf("io.netty.handler.codec.http.HttpObjectDecoder")
+    .advise(method("decode").and(takesArguments(3)), classOf[ServerDecodeMethodAdvisor])
+
+  onSubTypesOf("io.netty.handler.codec.http.HttpObjectEncoder")
+    .advise(method("encode").and(takesArguments(3)), classOf[ServerEncodeMethodAdvisor])
+
+//  @After("execution(* io.netty.handler.codec.http.HttpObjectDecoder+.decode(..)) && args(ctx, *, out)")
```
ivantopo, line 30: Remove the commented code

instrumentation/kamon-netty/src/main/scala/kamon/netty/instrumentation/ServerBootstrapInstrumentation.scala
```diff
@@ -0,0 +1,67 @@
...
+  onSubTypesOf("io.netty.channel.EventLoopGroup")
+    .mixin(classOf[EventLoopMixin])
+
+  onType("io.netty.bootstrap.ServerBootstrap")
+    .advise(method("group").and(takesArguments(2)), classOf[ServerGroupMethodAdvisor])
+
+  onType("io.netty.bootstrap.ServerBootstrap$ServerBootstrapAcceptor")
+    .advise(method("channelRead").and(takesArguments(2)), classOf[ServerChannelReadMethodAdvisor])
+
+//  @Before("execution(* io.netty.bootstrap.ServerBootstrap.group(..)) && args(bossGroup, workerGroup)")
```
ivantopo, line 34: Remove please

instrumentation/kamon-netty/src/main/scala/kamon/netty/instrumentation/mixin/HttpMessageMixin.scala
```diff
@@ -0,0 +1,26 @@
+package kamon.netty.instrumentation.mixin
```
ivantopo, line 1: This class should be removed and use `HasContext` from instrumentation-common instead

ivantopo (changes requested):
There are many tiny details to figure out here and there, and three bigger questions to address:
- Is there a way this instrumentation can work with the latest Netty version? It will most certainly happen that some people will have newer versions in their classpath and as far as I understand, that would break this instrumentation. Some ideas that come to mind:
  - Find a different way of extracting the event loop name for the metrics
  - Add some conditions to ensure that event loop metrics will only be collected on supported versions and not fail on newer ones
- There are no tests for epoll-based loops. I saw the helper methods in the tests side but they are not being used. Is it because this doesn't support epoll or the tests should be duplicated for epoll and that should be it?
- What happens when someone is using a higher-level framework that builds on top of Netty (e.g. Armeria). Do we get double spans in that case? :thinking:

PR #1117, opened 2022-02-23
measure the entire Jedis roundtrip time, fixes #1082

### LINKED ISSUE #1082: The Jedis instrumentation is not accounting for the entire round trip to Redis
I had a quick chat with @jtjeferreira about it. He discovered that the Jedis instrumentation is only measuring the time it takes to send the requests to Redis and leaving out the time it might take to receive the response. I'll look into this later this week.

ivantopo:
Turns out that our current Jedis instrumentation is only measuring the time it takes to send a command to Redis, without taking into account how long will it take to get an answer back. This PR fixes that. Thanks to @jtjeferreira for bringing this up!

TODO:
  - [x] Check if we should change the span names and tags to something that matches the otel spec
  - [x] Try it out with Jedis 4.x
  - [x] Ignore other possibly unwanted methods from generating spans in the Jedis class (currently only `toString`, `hashCode`, and `close` are ignored)

### DIFF
--- a/instrumentation/kamon-redis/src/main/scala/kamon/instrumentation/jedis/JedisInstrumentation.scala
+++ b/instrumentation/kamon-redis/src/main/scala/kamon/instrumentation/jedis/JedisInstrumentation.scala
@@ -24,10 +24,14 @@ import kamon.trace.Span
 import kanela.agent.api.instrumentation.InstrumentationBuilder
 import kanela.agent.libs.net.bytebuddy.asm.Advice
 import kanela.agent.libs.net.bytebuddy.description.method.MethodDescription
-import kanela.agent.libs.net.bytebuddy.matcher.ElementMatchers
-import kanela.agent.libs.net.bytebuddy.matcher.ElementMatchers.{isMethod, isPublic, isStatic, namedOneOf, not, takesArgument, takesArguments}
+import kanela.agent.libs.net.bytebuddy.matcher.ElementMatchers.{isMethod, isPublic, isStatic, namedOneOf, not}
 
 class JedisInstrumentation extends InstrumentationBuilder {
+
+  /**
+    * Most methods in `Jedis` and `BinaryJedis` end up sending calls to the Redis server, so we are
+    * targeting all methods in those classes, except for the ones excluded below.
+    */
   onTypes("redis.clients.jedis.Jedis", "redis.clients.jedis.BinaryJedis")
     .advise(
       isMethod[MethodDescription]()
@@ -51,6 +55,9 @@ class JedisInstrumentation extends InstrumentationBuilder {
       classOf[ClientOperationsAdvice])
 
 
+  /**
+    * For Jedis 3.x. This advice ensures we get the right command name in the Span.
+    */
   onType("redis.clients.jedis.Protocol")
     .advise(
       method("sendCommand")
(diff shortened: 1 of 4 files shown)


instrumentation/kamon-redis/src/main/scala/kamon/instrumentation/jedis/JedisInstrumentation.scala
```diff
@@ -14,36 +14,47 @@
...
 import kanela.agent.libs.net.bytebuddy.asm.Advice
 import kanela.agent.libs.net.bytebuddy.description.method.MethodDescription
-import kanela.agent.libs.net.bytebuddy.matcher.ElementMatchers.isPublic
+import kanela.agent.libs.net.bytebuddy.matcher.ElementMatchers.{isMethod, isPublic, namedOneOf, not, whereAny}
+import redis.clients.jedis.Protocol
 
 class JedisInstrumentation extends InstrumentationBuilder {
-  onType("redis.clients.jedis.Protocol")
-    .advise(method("sendCommand").and(isPublic[MethodDescription]), classOf[SendCommandAdvice])
+  onType("redis.clients.jedis.Jedis")
```
jtjeferreira, line 30: `redis.clients.jedis.Jedis` is only one of the implementations of `redis.clients.jedis.JedisCommands`. I guess we should add tests for the other classes...

ivantopo: I noticed that there is `Jedis` and `ShardedJedis`, but the sharded version uses a regular Jedis under the hood so it should work fine. There are other variants too, like `JedisCluster` and in 4.x there is a `JedisPool` too. I'll leave all of those for a separate PR.

ivantopo: More details here: https://github.com/kamon-io/Kamon/issues/1121

ivantopo:
I managed to test it locally and add a workaround for this to work with Jedis 4.1, at least at a very basic level. We still need to figure out support for some 4.x-only components (see https://github.com/kamon-io/Kamon/issues/1121 for follow ups). This is ready to review.

instrumentation/kamon-redis/src/main/scala/kamon/instrumentation/jedis/JedisInstrumentation.scala
```diff
@@ -14,36 +14,125 @@
...
+
   onType("redis.clients.jedis.Protocol")
-    .advise(method("sendCommand").and(isPublic[MethodDescription]), classOf[SendCommandAdvice])
+    .when(classIsPresent("redis.clients.jedis.CommandArguments"))
+    .advise(
+      method("sendCommand")
+        .and(isPublic[MethodDescription])
+        .and(isStatic[MethodDescription])
+        .and(takesArguments(2)),
+      classOf[SendCommandAdviceForJedis4])
```
jtjeferreira, line 69: I guess this is for supporting different Jedis versions, but a few comments explaining, will help the next developer who looks at this :)

ivantopo: Pushed a change for it. Thanks for the hint Joao!

PR #1119, opened 2022-02-24
apply gRPC tags and take a sampling decision on Akka gRPC server requests, fixes #1118

### LINKED ISSUE #1118: Server-side Akka gRPC requests are not handled properly
Noticed that server requests are not finishing their traces because `.takeSamplingDecision()` is never called. This means we end up without traces, and metrics only show one `http.server.request` operation. This is only on the server side, I didn't get into reviewing the client side yet.

ivantopo:
Adds a few [rpc tags](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/rpc/#attributes) and ensures that Kamon will take a sampling decision on matched gRPC services. I'm not including tests because Akka gRPC doesn't work on Scaka 2.11 and the entire process of tweaking the build would be considerable more work than the tiny instrumentation addition.


jtjeferreira (approved):
LGTM, but would be nice to have some tests

ivantopo:
Yeah, I'm considering creating a new subproject `kamon-akka-grpc` and putting it all there, only for scala 2.12+. I was also looking at the client side of it and I think it wouldn't be super complicated to implement. If I find time for it over the weekend I'll put it all together, otherwise we ship only the server side and do the subproject+client later

Issue #1115, opened 2022-02-22
Consider a clear way to avoid starting traces on the client/producer side of requests


ivantopo:
I just noticed a situation where traces would be started automatically by the Kafka producer instrumentation and then followed on all consumers, but there was no actual current trace on the service producing the messages. The Kafka producer instrumentation, and all client instrumentations for that matter, should not start a trace automatically if there wasn't a trace already available. Probably controllable with a config setting as we already have for HTTP clients: https://github.com/kamon-io/Kamon/pull/1006

ivantopo:
Added a small related control for producers via #1125.

PR #1125, opened 2022-02-24
do not start start traces in the producer instrumentation, related to #1115

ivantopo:
Introduces a new setting `start-trace-on-producer` that determines whether the producer instrumentation should start a new Span (and trace) when producing a message without any current trace on the producing application. In most cases you don't want that, so we are defaulting it to `no`. 

This will prevent apps from creating unnecessary spans on the producer side that can in turn generate A LOT of spans on the consumer side.

### DIFF
--- a/instrumentation/kamon-kafka/src/main/resources/reference.conf
+++ b/instrumentation/kamon-kafka/src/main/resources/reference.conf
@@ -6,6 +6,11 @@ kamon.instrumentation.kafka {
   client {
     tracing {
 
+      # Decides whether the Producer instrumentation should start a new trace when sending records to Kafka and no
+      # trace is currently running on the application. You almost never will the Producer to start new traces, but
+      # we have this setting for backwards compatibility reasons.
+      start-trace-on-producer = no
+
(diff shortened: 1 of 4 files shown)


instrumentation/kamon-kafka/src/main/resources/reference.conf
```diff
@@ -6,6 +6,11 @@ kamon.instrumentation.kafka {
   client {
     tracing {
 
+      # Decides whether the Producer instrumentation should start a new trace when sending records to Kafka and no
+      # trace is currently running on the application. You almost never will the Producer to start new traces, but
```
OriolMunoz, line 10:
```suggestion
      # trace is currently running on the application. You almost never want the Producer to start new traces, but
```

OriolMunoz (commented):
just a tiny change :D

instrumentation/kamon-kafka/src/main/resources/reference.conf
ivantopo: Thanks, Oriol!

Issue #1127, opened 2022-02-25
Cats 3.x support 


cmcmteixeira:
This is not yet mergeable for obvious reason (too many comments, printlns and debug code)

As previously discussed on discord, I attempted to instrument the internals of cats-effect 3.x .

This lead me through a bit of a journey through the internals of cats-effect which I'm yet to fully understand.

My findings should therefore be taken with a pinch of :salt:

###Instrumenting the runLoop

This was my initial approach to this problem.

Initially it seemed to work :+1:  but later I found that it suffer from the same issues as the solutions bellow.

It also has the downside of inserting a `Kamon.storeContext / Kamon.currentContext` on every enter/exit of this loop which I fear is quite taxing performance wise.

###Instrumenting run() and "forks"

This methodology proved to work a bit better. It also avoided the previous performance downside altogether since most of the calls will be done around the `run` method and not the `runLoop`.

Unfortunately it doesn't work 100% of the times.

**Current Problem**

This brings us to the current state of affairs and the problem at hand.

Running the tests in the `kamon.instrumentation.futures.cats.CatsIoInstrumentationSpec` will show something very similar to this on the console:
```
run(enter)     | cFiber: 1696169892 | sFiber: NA         | FiberCtx: {}              | ThreadCtx {}              | Thread pool-4-thread-1
runLoop(Enter) | cFiber: 1696169892 | sFiber: NA         | FiberCtx: {}              | ThreadCtx {}              | Thread pool-4-thread-1
Context will be set
ScheduleNew    | cFiber: 1696169892 | sFiber: 1696169892 | FiberCtx: {}              | ThreadCtx {key=value}     | Thread pool-4-thread-1
runLoop(Exit)  | cFiber: 1696169892 | sFiber: NA         | FiberCtx: {key=value}     | ThreadCtx {key=value}     | Thread pool-4-thread-1
run(enter)     | cFiber: 1696169892 | sFiber: NA         | FiberCtx: {key=value}     | ThreadCtx {}              | Thread scala-execution-context-global-33
run(exit)      | cFiber: 1696169892 | sFiber: NA         | FiberCtx: {key=value}     | ThreadCtx {}              | Thread pool-4-thread-1
runLoop(Enter) | cFiber: 1696169892 | sFiber: NA         | FiberCtx: {key=value}     | ThreadCtx {key=value}     | Thread scala-execution-context-global-33
[...]

PR #1128, opened 2022-02-25
Cats Effect 3.x initial implementation

cmcmteixeira:
Details can be found [here](https://github.com/kamon-io/Kamon/issues/1127)


ivantopo:
@cmcmteixeira please take a look at this [issue](https://github.com/kamon-io/Kamon/runs/5336397196?check_suite_focus=true#step:5:6991) when trying to compile for Scala 2.11

dpsoft:
There is not support for scala 2.11. `All current Cats Effect releases are published for Scala 2.12, 2.13, 3.0`

build.sbt
```diff
@@ -242,9 +242,9 @@ lazy val `kamon-cats-io` = (project in file("instrumentation/kamon-cats-io"))
       kanelaAgent % "provided",
       {
         if(scalaBinaryVersion.value == "2.11")
-          "org.typelevel" %% "cats-effect" % "2.0.0" % "provided"
+          "org.typelevel" %% "cats-effect" % "3.3.5" % "provided"
```
hughsimpson, line 245: I think it might be saner to split this into kamon-cats-io/kamon-cats-io_3 ?

cmcmteixeira: Probably the best way to go indeed.. Just gotta make it work first :wink:

PR #1130, opened 2022-02-28
Fix path to authentication token.

### LINKED ISSUE #1129: Authentication token for module kamon-influxdb is not recognized
As of PR #1110 kamon-influxdb should now properly support InfluxDB authentication tokens. Unfortunately the [respective code](https://github.com/kamon-io/Kamon/blob/master/reporters/kamon-influxdb/src/main/scala/kamon/influxdb/InfluxDBReporter.scala#L233) looks at `kamon.influxdb.token` and not at `kamon.influxdb.authentication.token` as stated in the [documentation of kamon-influxdb](https://kamon.io/docs/latest/reporters/influxdb/#authentication).

duese:
Fixes #1129


ivantopo:
Thanks @duese! This was such a dumb mistake on my side :facepalm:

PR #1145, opened 2022-03-19
DataDog Trace HTTP Propagation

seglo:
ref: https://github.com/kamon-io/Kamon/discussions/1031

Basic DataDog HTTP Propagation. This does not include support for [other headers the `dd-trace-java` library supports](https://github.com/DataDog/dd-trace-java/blob/76e41d3d51314ed2814d0c8deac78c17bb87638e/dd-trace-core/src/main/java/datadog/trace/core/propagation/DatadogHttpCodec.java#L24..L27) including `ot-baggage-` prefixed headers and `x-datadog-origin`. I think it could be added simply, but we don't use either header in our solutions ATM.

I also want to give credit to my colleague for his contributions @linkleonard

### DIFF
--- a/core/kamon-core-tests/src/test/scala/kamon/trace/DataDogSpanPropagationSpec.scala
+++ b/core/kamon-core-tests/src/test/scala/kamon/trace/DataDogSpanPropagationSpec.scala
@@ -1,0 +1,157 @@
+package kamon.trace
+
+import kamon.context.{Context, HttpPropagation}
+import kamon.trace.Trace.SamplingDecision
+import org.scalatest.OptionValues
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpec
+
+import scala.collection.mutable
+
+class DataDogSpanPropagationSpec extends AnyWordSpec with Matchers with OptionValues {
+  val dataDogPropagation = SpanPropagation.DataDog()
+
+  "The DataDog Span propagation for HTTP" should {
+    "write the Span data into headers" in {
+      val headersMap = mutable.Map.empty[String, String]
+      dataDogPropagation.write(testContext(), headerWriterFromMap(headersMap))
+
+      headersMap.get("x-datadog-trace-id").value shouldBe unsignedLongString("1234")
+      // not a typo, span id should be set as `x-datadog-parent-id`
+      headersMap.get("x-datadog-parent-id").value shouldBe unsignedLongString("4321")
+      headersMap.get("x-datadog-sampling-priority").value shouldBe "1"
+    }
+
+    "not inject anything if there is no Span in the Context" in {
+      val headersMap = mutable.Map.empty[String, String]
+      dataDogPropagation.write(Context.Empty, headerWriterFromMap(headersMap))
+      headersMap.values shouldBe empty
+    }
+
+    "extract a RemoteSpan from incoming headers when all fields are set" in {
+      val headersMap = Map(
+        "x-datadog-trace-id" -> unsignedLongString("1234"),
+        "x-datadog-parent-id" -> unsignedLongString("4321"),
+        "x-datadog-sampling-priority" -> "1"
+      )
+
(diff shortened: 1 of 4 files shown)


core/kamon-core/src/main/scala/kamon/trace/SpanPropagation.scala
```diff
@@ -417,4 +417,85 @@ object W3CTraceContext {
...
+      val identifierScheme = Kamon.identifierScheme
+      val traceId = reader.read(Headers.TraceId)
+        .map(id => identifierScheme.traceIdFactory.from(decodeUnsignedLongToHex(id)))
+        .getOrElse(Identifier.Empty)
+
+      // To preserve the parent/child relationship of spans in DataDog the `x-parent-parent-id` is used as the span id
+      // of the new span. See `DatadogHttpCodec` implementation in `dd-trace-java` for details.
+      val spanId = reader.read(Headers.ParentSpanId)
+        .map(id => identifierScheme.spanIdFactory.from(decodeUnsignedLongToHex(id)))
+        .getOrElse(Identifier.Empty)
```
ivantopo, line 469: I'm not super sure if this is the place to apply this logic. Have you looked at the `kamon.trace.join-remote-parents-with-same-span-id` setting? It sounds like that setting should roughly have the same effect, although parentId will be set in that case.

ivantopo: Actually, I just noticed that the client side is only including the span id as the parent id. Weird, but if that's how DD does it :shrug:

core/kamon-core-tests/src/test/scala/kamon/trace/DataDogSpanPropagationSpec.scala
```diff
@@ -0,0 +1,157 @@
...
+        .get(Span.Key).trace.samplingDecision shouldBe SamplingDecision.Sample
+
+      dataDogPropagation.read(headerReaderFromMap(notSampledHeaders), Context.Empty)
+        .get(Span.Key).trace.samplingDecision shouldBe SamplingDecision.DoNotSample
+
+      dataDogPropagation.read(headerReaderFromMap(noSamplingHeaders), Context.Empty)
+        .get(Span.Key).trace.samplingDecision shouldBe SamplingDecision.Unknown
+    }
+
+    "not include the x-datadog-sampling-priorit header if the sampling decision is unknown" in {
```
ivantopo, line 73: Tiny typo in priority

core/kamon-core/src/main/scala/kamon/trace/SpanPropagation.scala
ivantopo: @seglo so, Datadog ends up merging the data sent by the client-side service and the server-side service into the same Span? Asking out of curiosity

ivantopo:
Hey @seglo, thanks for implementing this! 

Overall it looks great. If you have a chance to fix the typo before Friday please do, otherwise I'll do it and merge the PR so it makes it to the next release :muscle:

core/kamon-core/src/main/scala/kamon/trace/SpanPropagation.scala
seglo: I agree, it's confusing.

> Datadog ends up merging the data sent by the client-side service and the server-side service into the same Span?

Intuitively this is what I thought would happen as well, but the DataDog trace view shows them as separate spans.

My first thought was to assign the id passed as `x-datadog-parent-id` to the Parent ID of the Kamon trace and generate a new ID for the span itself, but this broke the parent/child association in the DataDog trace view.

Here's what we see, the top span is from our frontend and the 2nd is an akka-http service using this implementation.

datadog-trace

seglo:
Excellent. Thanks for the review @ivantopo.

core/kamon-core/src/main/scala/kamon/trace/SpanPropagation.scala
ivantopo: Thanks for sharing!

PR #1147, opened 2022-03-23
Fix twitter future context propagation

ivantopo:
This PR changes the way we instrument Twitter Futures to ensure we always do context propagation. It was motivated by this discussion: https://github.com/kamon-io/Kamon/discussions/1005


dpsoft (approved):

dispalt:
Confirmed this is working for us, so a big +1 from me.

PR #1148, opened 2022-03-23
allow filtering metrics at a reporter level, fixes #1146

### LINKED ISSUE #1146: Ability to filter metrics exposed in prometheus EP
Hello,

Is it possible to filter metrics that will be exposed by prometheus reporter?
E.g. I would like to blacklist some of the collected metrics so they won't be returned in Prometheus metrics endpoint response

Regards

ivantopo:
This filters metric snapshots before they get posted to a reporter. We already had the the settings in place since Kamon 2.3.0 but for some reason I didn't actually implement the filters :sob:. This PR adds an implementation. Fixes #1146.


dpsoft (approved):
awesome!

PR #1152, opened 2022-04-02
trace with http2->http2

mofei100:
fix http2 enable,but http1 endpoint can not trace


ivantopo:
Hey @mofei100!

I'm wondering, what is happening differently on this PR that was not fixed already with #1094? Is it just about tracing the Upgrade request or is there something else missing?

PR #1180, opened 2022-06-15
Adds metrics-excludes configuration to summary setttings

Scarysize:
Allows to exclude certain metrics from being reported as summaries. This is especially helpful when using "**" to report all metrics as summaries.


TjarkoG:
Any Feedback on this?

Issue #1177, opened 2022-06-06
Kamon tracing supports Redis Radisson client? (Not jedis/lettuce/redisscala)


nmadhavi33:
I am looking for kamon.conf file configuration for redis Redisson client? (https://mvnrepository.com/artifact/org.redisson/redisson) client.  Client is not jedis/lettuce/redisscala

Whitilied:
waiting for my pr 
https://github.com/kamon-io/Kamon/pull/1190

PR #1190, opened 2022-07-19
kamon-redis supports redisson client

Whitilied:


Whitilied:
link to
https://github.com/kamon-io/Kamon/issues/1177

Whitilied:
@ivantopo

Whitilied:
@timcharper

instrumentation/kamon-redis/src/main/scala/kamon/instrumentation/redisson/RedissonInstrumentation.scala
```diff
@@ -0,0 +1,68 @@
...
+        (spanName, statements)
+      case command: CommandData[_, _] =>
+        val spanName = s"redis.command.${command.getCommand.getName}"
+        val statement = parseCommand(command)
+        (spanName, statement)
+    }
+
+    Kamon.clientSpanBuilder(commandName, "redisson")
+      .tag("db.operation", commandName)
+      .tagMetrics("db.system", "redis")
```
ivantopo, line 46: Hey @Whitilied, do you think it would be possible to ass the `db.statement` tag here with the contents of that local `statements` variable? I'm just looking at the OpenTelemetry [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#redis) (screenshot below) and thinking it would be nice to set that field as well, if we have it!

image

Whitilied: gree with u

Whitilied: @ivantopo change to the OpenTelemetry's Semantic Conventions

anujprashar:
Hi @Whitilied @ivantopo When can we expect this feature to be released?

Issue #1064, opened 2021-10-21
Publish instrumentation artifacts for Scala 3


ivantopo:
We went through upgrading the core libraries and reporters for Scala 3, but all the instrumentation is still pending. I got this exception at my very first attempt at running the `instrumentation-common` tests with Scala 3 and thought that instead of delaying a new release even more we should publish the parts that are already supported, and create this issue to track instrumentation updates.

```
[pool-1-thread-1-ScalaTest-running-ContextInstrumentationSpec] ERROR 2021-10-21 09:55:12  Logger : Error => kamon.instrumentation.context.ContextInstrumentationSpec$Target with message public kamon.context.Context kamon.instrumentation.context.ContextInstrumentationSpec$Target.doWork() does not define an index 0. Class loader: sun.misc.Launcher$AppClassLoader@18b4aac2: java.lang.IllegalStateException: public kamon.context.Context kamon.instrumentation.context.ContextInstrumentationSpec$Target.doWork() does not define an index 0
	at kanela.agent.libs.net.bytebuddy.asm.Advice$OffsetMapping$ForArgument$Unresolved.resolve(Advice.java:1664)
	at kanela.agent.libs.net.bytebuddy.asm.Advice$OffsetMapping$ForArgument.resolve(Advice.java:1575)
	at kanela.agent.libs.net.bytebuddy.asm.Advice$OffsetMapping$ForArgument$Unresolved.resolve(Advice.java:1683)
	at kanela.agent.libs.net.bytebuddy.asm.Advice$Dispatcher$Inlining$Resolved$ForMethodEnter.doApply(Advice.java:8305)
	at kanela.agent.libs.net.bytebuddy.asm.Advice$Dispatcher$Inlining$Resolved$ForMethodEnter.apply(Advice.java:8266)
	at kanela.agent.libs.net.bytebuddy.asm.Advice$Dispatcher$Inlining$Resolved$AdviceMethodInliner.visitMethod(Advice.java:7970)
	at kanela.agent.libs.net.bytebuddy.jar.asm.ClassReader.readMethod(ClassReader.java:1353)
	at kanela.agent.libs.net.bytebuddy.jar.asm.ClassReader.accept(ClassReader.java:744)
	at kanela.agent.libs.net.bytebuddy.jar.asm.ClassReader.accept(ClassReader.java:424)
[...]

PR #1191, opened 2022-07-19
Cross compile kamon-instrumentation-common with scala 3

vaslabs:
Hello,
Thank you for your work on Kamon.

We are recently looking to upgrade to scala 3 and stumbled upon some kamon modules not having scala 3 builds. One in particular is kamon-http4s which depends on kamon-core and kamon-instrumentation-common .

~~It seems easy enough to add scala 3 to the latter. I have already done the work for kamon-http4s , just need an official release if you can and send you a PR on that project too.~~

Please let me know what you think.

Kind regards

EDIT:
just saw this https://github.com/kamon-io/Kamon/issues/1064

I'll have a deeper look

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -160,6 +160,8 @@ lazy val `kamon-instrumentation-common` = (project in file("instrumentation/kamo
   .settings(instrumentationSettings)
   .settings(
         crossScalaVersions += `scala_3_version`,
+  ).settings(
+    scalacOptions ++= { if(scalaBinaryVersion.value.startsWith("3.")) Seq("-Xtarget:8") else Seq.empty }
   )
   .settings(
     resolvers += Resolver.mavenLocal,
--- a/build.sbt
+++ b/build.sbt
@@ -160,8 +160,6 @@ lazy val `kamon-instrumentation-common` = (project in file("instrumentation/kamo
   .settings(instrumentationSettings)
   .settings(
         crossScalaVersions += `scala_3_version`,
-  ).settings(
-    scalacOptions ++= { if(scalaBinaryVersion.value.startsWith("3.")) Seq("-Xtarget:8") else Seq.empty }
   )
   .settings(
     resolvers += Resolver.mavenLocal,
@@ -826,7 +824,6 @@ lazy val bundle = (project in file("bundle"))
 
 import com.lightbend.sbt.javaagent.Modules
 import BundleKeys._
-import sbt.Keys.scalacOptions
 
 lazy val commonBundleSettings = Seq(
   moduleName := "kamon-bundle",
--- a/build.sbt
+++ b/build.sbt
@@ -159,6 +159,9 @@ lazy val `kamon-instrumentation-common` = (project in file("instrumentation/kamo
   .enablePlugins(JavaAgent)
   .settings(instrumentationSettings)
   .settings(
+        crossScalaVersions += `scala_3_version`,
+  )
+  .settings(
     resolvers += Resolver.mavenLocal,
     libraryDependencies ++= Seq(
       slf4jApi % "test",
--- a/build.sbt
+++ b/build.sbt
@@ -824,6 +824,7 @@ lazy val bundle = (project in file("bundle"))
 
 import com.lightbend.sbt.javaagent.Modules
 import BundleKeys._
+import sbt.Keys.scalacOptions
 
 lazy val commonBundleSettings = Seq(
   moduleName := "kamon-bundle",
--- a/instrumentation/kamon-instrumentation-common/src/main/scala-2.11/scala/annotation/static.scala
+++ b/instrumentation/kamon-instrumentation-common/src/main/scala-2.11/scala/annotation/static.scala
@@ -1,0 +1,4 @@
+package scala.annotation
+
+import scala.annotation.meta._
+final class static extends StaticAnnotation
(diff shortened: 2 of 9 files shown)


vaslabs:
I'll put my notes here

The issue seems to be originating from byte buddy:
```java
@Override
                protected ParameterDescription resolve(MethodDescription instrumentedMethod) {
                    ParameterList<?> parameters = instrumentedMethod.getParameters();
                    if (parameters.size() <= index) {
                        throw new IllegalStateException(instrumentedMethod + " does not define an index " + index);
                    } else {
                        return parameters.get(index);
                    }
                }
```

The bytecode (javap -v) output of scala 3 doWork method is
```
  public kamon.context.Context doWork();
    descriptor: ()Lkamon/context/Context;
    flags: ACC_PUBLIC
    Code:
      stack=1, locals=1, args_size=1
         0: getstatic     #22                 // Field kamon/Kamon$.MODULE$:Lkamon/Kamon$;
         3: invokevirtual #25                 // Method kamon/Kamon$.currentContext:()Lkamon/context/Context;
         6: areturn
      LineNumberTable:
        line 77: 0
      LocalVariableTable:
        Start  Length  Slot  Name   Signature
            0       7     0  this   Lkamon/instrumentation/context/ContextInstrumentationSpec$Target;

```

The scala 2.13 is
```
  public kamon.context.Context doWork();
    descriptor: ()Lkamon/context/Context;
    flags: ACC_PUBLIC
    Code:
      stack=1, locals=1, args_size=1
         0: getstatic     #16                 // Field kamon/Kamon$.MODULE$:Lkamon/Kamon$;
         3: invokevirtual #19                 // Method kamon/Kamon$.currentContext:()Lkamon/context/Context;
         6: areturn
      LineNumberTable:
        line 77: 0
      LocalVariableTable:
        Start  Length  Slot  Name   Signature
            0       7     0  this   Lkamon/instrumentation/context/ContextInstrumentationSpec$Target;


```

vaslabs:
Found a difference in the bytecode which looks to me like debug symbols (from -g:vars) which is ignored in the scala 3 compilation.
So decompiling with FernFlower we have

### Scala 2.13

```java
public final class InvokeWithCapturedContext {
   @OnMethodExit(
      onThrowable = Throwable.class
   )
   public static void exit(@Enter final Storage.Scope scope) {
      .MODULE$.exit(scope);
   }

   @OnMethodEnter
   public static Storage.Scope enter(@This final HasContext hasContext) {
      return .MODULE$.enter(hasContext);
   }
}
```

## Scala 3

```java
public final class InvokeWithCapturedContext {
   @OnMethodEnter
   public static Storage.Scope enter(HasContext var0) {
      return .MODULE$.enter(var0);
   }

   @OnMethodExit(
      onThrowable = Throwable.class
   )
   public static void exit(Storage.Scope var0) {
      .MODULE$.exit(var0);
   }
}

```

vaslabs:
It also looks like that parameter annotations are kept but:

The accompanying scala 3 `InvokeWithCapturedContext$` that has the singleton implementation:
```java

public final class InvokeWithCapturedContext$ implements Serializable {
   public static final InvokeWithCapturedContext$ MODULE$ = new InvokeWithCapturedContext$();

   private InvokeWithCapturedContext$() {
   }

   private Object writeReplace() {
      return new ModuleSerializationProxy(InvokeWithCapturedContext$.class);
   }

   @OnMethodEnter
   public Storage.Scope enter(@This final HasContext hasContext) {
      return .MODULE$.storeContext(hasContext.context());
   }

   @OnMethodExit(
      onThrowable = Throwable.class
   )
   public void exit(@Enter final Storage.Scope scope) {
      scope.close();
   }
}
```

and the scala 2.13


```java
public final class InvokeWithCapturedContext$ {
   public static final InvokeWithCapturedContext$ MODULE$ = new InvokeWithCapturedContext$();

   @OnMethodEnter
   public Storage.Scope enter(@This final HasContext hasContext) {
      return .MODULE$.storeContext(hasContext.context());
   }

   @OnMethodExit(
      onThrowable = Throwable.class
   )
   public void exit(@Enter final Storage.Scope scope) {
      scope.close();
   }

   private InvokeWithCapturedContext$() {
   }
}

```


The annotations go to the underlying singleton type but not to the top level with the static methods

vaslabs:
Ok I found a fix for this

For scala 3 we need to use `@static` which requires a companion class and gives full java compatibility with static methods.

But, this annotation does not seem to be available for scala 2.13 so we'll need other version specific implementations

vaslabs:
This should be ready to be tested for #1191

build.sbt
```diff
@@ -158,6 +158,11 @@ lazy val `kamon-instrumentation-common` = (project in file("instrumentation/kamo
   .disablePlugins(AssemblyPlugin)
   .enablePlugins(JavaAgent)
   .settings(instrumentationSettings)
+  .settings(
+        crossScalaVersions += `scala_3_version`,
+  ).settings(
+    scalacOptions ++= { if(scalaBinaryVersion.value.startsWith("3.")) Seq("-Xtarget:8") else Seq.empty }
```
ivantopo, line 164: Can you please move this setting to this other file: https://github.com/kamon-io/Kamon/blob/master/project/Build.scala#L167

build.sbt
```diff
@@ -821,6 +826,7 @@ lazy val bundle = (project in file("bundle"))
 
 import com.lightbend.sbt.javaagent.Modules
 import BundleKeys._
+import sbt.Keys.scalacOptions
```
ivantopo, line 829: This will go away when the previous line gets moved

ivantopo (changes requested):
Thank you so much @vaslabs!

There is a tiny change in the `scalacOptions` setting and the syntax of annotations, just to make them easier to read.

ivantopo (approved):

ivantopo:
Thank you so much for the effort @vaslabs!

PR #1204, opened 2022-09-07
increase the default range of http.server.connection.lifetime metric, fixes #1201

### LINKED ISSUE #1201: Can't record big values on `http.server.connection.lifetime`
One of our users is getting this error with long-lived HTTP connections (web sockets):

```
Failed to record value [7774197398489] on [http.server.connection.lifetime,{port=9000,component=play.server.akka-http,interface=0.0.0.0}] because the value is outside of the configured range. The recorded value was adjusted to the highest trackable value [3600000000000]
```

By default our histograms can cover up to 1h latencies and in this case, a connection was alive for over 2 hours at the time it closed. If it was a more commonly used metric I would suggest increasing the range in the metric but in this case we are not sure of how big this number can get as it could go all the way to days. I think we could:
  - Increase the lower discernible value for this metric so we can pack a much higher bound in the range while keeping the same memory footprint
  - Use a lower precision histogram for this metric, with recorded values up to 10% away from the originally recorded value. This can massively increase range and probably reduce footprint.

Whatever we decide, it should be the default for this metric so nobody faces the same issue in the future!

ivantopo:
Details in #1201.

### DIFF
--- a/instrumentation/kamon-instrumentation-common/src/main/resources/reference.conf
+++ b/instrumentation/kamon-instrumentation-common/src/main/resources/reference.conf
@@ -251,4 +251,15 @@ kamon {
       }
     }
   }
+
+  metric.factory.custom-settings {
+
+    # Increases the range of this metric to cover up to ~7 days while still saving 30% of space in
+    # allocated histograms. Should prevent the most common warnings about out-of-range updates.
+    #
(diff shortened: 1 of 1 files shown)


Dichotomia:
+1

PR #1213, opened 2022-10-10
tweak cats-effect-3 instrumentation

hughsimpson:
- add new test case for cats-io-3 instrumentation
- rm 'RestoreContextOnSuccessfulResume' advice

I've been running into some funny behaviour when executing code using dispatchers and was able to minimise it in the tests. It seems like the 'resume' advice actually pollutes context under some circumstances, and it's not obvious that it ever actually does the right thing.

Some of the other instrumentation seemed redundant in light of this, so I removed that too. Tests still pass, as does the new one which was consistently failing before - however I'm conscious that I may be missing something here.

I'll think about this a bit more, and add new tests if inspiration strikes as to why we thought we needed those advice points in the first place, but thought it sensible to open up a pr for further discussion in the meantime

### DIFF
--- a/instrumentation/kamon-cats-io-3/src/test/scala/kamon/instrumentation/futures/cats3/CatsIoInstrumentationSpec.scala
+++ b/instrumentation/kamon-cats-io-3/src/test/scala/kamon/instrumentation/futures/cats3/CatsIoInstrumentationSpec.scala
@@ -1,7 +1,7 @@
 package kamon.instrumentation.futures.cats3
 
 import cats.effect.unsafe.{IORuntime, IORuntimeConfig, Scheduler}
-import cats.effect.{IO, Resource, Spawn}
+import cats.effect.{Deferred, IO, Resource, Spawn}
 import kamon.Kamon
 import kamon.context.Context
 import kamon.tag.Lookups.plain
@@ -11,16 +11,18 @@ import org.scalatest.matchers.should.Matchers
 import org.scalatest.wordspec.AnyWordSpec
 import java.util.concurrent.Executors
 import java.util.UUID
+
 import scala.concurrent.{Await, ExecutionContext}
 import scala.concurrent.ExecutionContext.global
 import scala.concurrent.duration._
+
 import cats.effect.std.Dispatcher
 import cats.implicits._
 import kamon.trace.Identifier.Scheme
 import kamon.trace.{Identifier, Span, Trace}
 
 class CatsIoInstrumentationSpec extends AnyWordSpec with Matchers with ScalaFutures with PatienceConfiguration
-    with OptionValues with Eventually with BeforeAndAfterEach {
+  with OptionValues with Eventually with BeforeAndAfterEach {
 
   System.setProperty("kamon.context.debug", "true")
 
@@ -86,7 +88,7 @@ class CatsIoInstrumentationSpec extends AnyWordSpec with Matchers with ScalaFutu
         test.unsafeRunSync()(runtime)
       }
 
-     "must allow complex Span topologies to be created" in {
+      "must allow complex Span topologies to be created" in {
         val parentSpan = Span.Remote(
           Scheme.Single.spanIdFactory.generate(),
           Identifier.Empty,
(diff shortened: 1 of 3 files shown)


hughsimpson:
@m50d @cmcmteixeira @ivantopo @dispalt @dpsoft - since you were involved in the initial effort, if any of you can spot something egregious with this, would appreciate eyes and feedback

m50d:
From vague memory there are two cases that are worth checking:

 - A fiber gets picked up by a scheduler thread after sleeping
 - Resuming from a future - note that in this case the cats-effect code deliberately races, so we need to do the right thing for both possible winners, which can be difficult to test

hughsimpson:
Ah yes. That helps. I'm now able to write a test that passed before and fails with these changes 😭

instrumentation/kamon-cats-io-3/src/test/scala/kamon/instrumentation/futures/cats3/CatsIoInstrumentationSpec.scala
```diff
@@ -125,6 +128,67 @@ class CatsIoInstrumentationSpec extends AnyWordSpec with Matchers with ScalaFutu
...
+        _ <- IO.sleep(100.millis)
+        _ <- d.complete(())
+        r <- t.joinWith(IO("" -> "nomatch"))
+        _ <- IO(s2.close())
+        _ <- IO(s1.close())
+//        _ <- IO(println(r._1 + ":"))
+//        _ <- IO(println(txId1, txId2, r._2))
+        // before: r._1 == r._2
+        // now: r._1 == txId2
+      } yield r._1 shouldEqual r._2
```
hughsimpson, line 188: I'm not completely certain that the new behaviour here is really so wrong... When resuming from a deferred completion, there's a sense in which the lexical scope has a claim to be 'the' context of execution, but the fiber that completes the deferred value also seems to have a justifiable claim - the desired semantics pobably differ significantly in different use cases...

Of course, consistency would be nice, and if I remove the `IO.sleep(100.millis)` in this for-comprehension, then I've found that the context introduced within the lexical scope of `withDeferred` wins and the tests pass

m50d: I would think that the vast majority of the time we would want the lexical scope to be the context? The most common use case I know of for things like this is where the deferred is completed by e.g. a database connection pool, or a "send outgoing webhooks" executor, whereas the fiber that suspended is the one that belongs to a specific request.

ivantopo:
@hughsimpson there is one case I would like to include on these possible tweaks:

```scala
      "must capture the current context when creating and running fibers" in {
        val runtime = IORuntime.global
        val anotherExecutionContext: ExecutionContext = ExecutionContext.fromExecutorService(Executors.newFixedThreadPool(10))
        val context = Context.of("tool", "kamon")


        val effect = for {
          contextOnAnotherThread <- IO.delay(Kamon.currentContext())
          _ <- IO.delay(Seq("hello", "world"))
          _ <- Spawn[IO].evalOn(IO.sleep(10.millis), anotherExecutionContext)
          _ <- IO.cede
          _ <- IO.cede
        } yield {
          val currentContext = Kamon.currentContext()
          currentContext shouldBe context
          contextOnAnotherThread shouldBe context
          currentContext
        }

        val contextInsideYield = Kamon.runWithContext(context) {
          // This is what would happen at the edges of the system, when Kamon has already
          // started a Span in an outer layer (usually the HTTP server instrumentation) and
          // when processing gets to user-level code, the users want to run their business
          // logic as an effect. We should always propagate the context that was available
          // at this point to the moment when the effect runs.
          
          Await.result(effect.unsafeToFuture()(runtime), 100.seconds)
        }

        context shouldBe contextInsideYield
      }
```

Hopefully, the comment in the test makes it clear. This test doesn't pass with these tweaks and I think it is a pretty common use case. What are your thoughts on this?

PS: I'm really excited to see you guys getting deep on this!

ivantopo:
I PR'd the test via https://github.com/kamon-io/Kamon/pull/1214, along with a tiny change to make the instrumentation compatible with 3.4.0-RC2

instrumentation/kamon-cats-io-3/src/test/scala/kamon/instrumentation/futures/cats3/CatsIoInstrumentationSpec.scala
hughsimpson: Yeah you're probably right, I don't know what I was thinking. Following exactly what's going on in the cats-effect code was twisting my melon

PR #1221, opened 2022-10-28
caffeine and annotation instrumentation modules for scala 3

hughsimpson:
Enables two more instrumentation modules for scala 3
- kamon-caffeine
- kamon-annotation

The scala 3 version bump is not actually required for this to work.

I was going to include kamon-redis in this, but since it includes a dep bump have stuck it in https://github.com/kamon-io/Kamon/pull/1223 instead - although AFAICT should be binary compatible so hopefully not a big deal


ivantopo:
Once again, thanks @hughsimpson!

PR #1223, opened 2022-10-28
kamon-redis bump + scala 3 compatibility 

hughsimpson:
Makes kamon-redis module work for scala 3. Bumps rediscala to new organization name for all scala versions except for scala 2.11 to enable this


ivantopo:
Thanks for the upgrade @hughsimpson!

PR #1226, opened 2022-11-03
Update okhttp3 to 4.10.0

### LINKED ISSUE #1183: Upgrade okhttp3 to 4.x for `kamon-prometheus`
Kamon currently depends on an old version of okhttp3 (see `Build.scala`):

```scala
val okHttp            = "com.squareup.okhttp3"  %  "okhttp"          % "3.14.7"
```

This creates issues with other packages such as `com.softwaremill.sttp.client3:okhttp-backend` who have upgraded to `okhttp3` version `4.9.3`: adding a dependency to `kamon-prometheus` effectively downgrades the `okhttp` package in my class path, which breaks my application.

It would be really nice to upgrade Kamon to the latest version of `okhttp3`, is this something you could consider?

azolotko:
Resolves #1183

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -531,7 +531,7 @@ lazy val `kamon-okhttp` = (project in file("instrumentation/kamon-okhttp"))
   .settings(
     libraryDependencies ++= Seq(
       kanelaAgent % "provided",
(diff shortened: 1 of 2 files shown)


ivantopo:
Thanks for the contribution @azolotko! I deployed a snapshot locally with this OkHttp version and everything is working fine :tada:

Issue #803, opened 2020-07-04
Support for W3C Correlation Context


ivantopo:
Thanks to @jmendesky for bring this up [on Gitter](https://gitter.im/kamon-io/Kamon?at=5efb474afa0c9221fc6cc209), I wasn't really aware that this was a thing already!

Indeed, there is a draft of the spec available [online](https://w3c.github.io/correlation-context/) and the format is very similar to what Kamon is already using for transferring context tags over HTTP. The implementation should be pretty straight forward. I think that once they settle on the [header name](https://github.com/w3c/correlation-context/issues/17) we can introduce this as an experimental feature.

A couple things that we need to figure out before start implementing:

### Storage and Encoding of Properties
Our context tags are simple key/value pairs. In the correlation context spec, each key/value pair could have several properties, which are some sort of nested key/value pairs. The concept of properties doesn't match our model and will need to figure out how to transfer them in Kamon. We could:
  - Ignore them? I'm not really sure of why someone might want to use properties instead a flat key/value pairs structure. If anyone can dig out a discussion with the rationale for this I'll be grateful!
  - Transfer them as a context entry (most likely scenario).

### Migration
This change will break context propagation through HTTP as soon as deployed in one service. We have several users with several services spread on different Kamon versions and this won't be a pleasant change for anyone. Maybe folks are not using context tags that much and this wont be such a big problem. At a minimum, we have to warn users about the breaking change.

ivantopo:
I'm bringing this comment made by @jmendesky on #979:

> :D we were thinking about that.
I suppose w3c baggage would be an alternative serialisation format to what is present at the moment (context-tags)?
Also, would it make sense to not support properties in the beginning? Its usage is not 100% clear and the w3c spec doesn't even show them in their examples. Without properties, the spec would be compatible with Kamon's context.

And also bringing @SimunKaracic and @duxet in.

Now, regarding this particular section: 

> would it make sense to not support properties in the beginning?

Yes, I think that would be totally fine. If we ever get to a point when we really need to carry those properties we could add them as a context entry instead of tags, but I would prefer to avoid that until the spec gets to a stable point.

@jmendesky I hear you are interested in giving this a try :smile:, how can we help you get there?

PR #1237, opened 2022-12-22
Allow W3C TraceContext propagation in kamon-kafka

### LINKED ISSUE #1235: Kafka instrumentation is not using w3c for context propagation
Hello,
We're using Kamon to instrument our akka services and it works really well! 
Because we use different languages across all the projects we decided to go with W3C trace-context for context propagation.

This works really great on Kamon with HTTP:
```
kamon.propagation.http.default.entries {
  incoming {
    span = "w3c"
  }
  outgoing {
    span = "w3c"
  }
}
```

However, we can't find a way to use the trace-context headers in kafka messages. Is there a config we can set for it? 

If it's not available but wanted, I could make a PR for it.

Symbianx:
Fixes #1235 by adding a new configuration key `kamon.instrumentation.kafka.client.tracing.propagator` that can take the value `w3c` to propagate the trace context using the W3C TraceContext specification.

Please excuse any bad scala code, I'm still a newb in this ecosystem.

### DIFF
--- a/instrumentation/kamon-kafka/src/main/scala/kamon/instrumentation/kafka/client/KafkaInstrumentation.scala
+++ b/instrumentation/kamon-kafka/src/main/scala/kamon/instrumentation/kafka/client/KafkaInstrumentation.scala
@@ -22,6 +22,7 @@ import kamon.context.Context
 import kamon.instrumentation.context.HasContext
 import kamon.trace.Span
 import org.apache.kafka.clients.consumer.ConsumerRecord
+import org.slf4j.LoggerFactory
 
 import scala.util.control.NonFatal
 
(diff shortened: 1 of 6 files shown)


Symbianx:
Haven't done tests yet because I'm having issues compiling `kamon-core` which fails with unresolved dependencies for `javax.annotation.Generated` and `sun.misc.Contended`. 

I have I feeling I'm using the wrong sbt or java versions but can't find any information on what settings I should be using to compile the project.

Any tips would be handy!

Symbianx:
Added a tests to assert w3c is being used and rebased the PR into 1 commit.

instrumentation/kamon-kafka/src/main/scala/kamon/instrumentation/kafka/client/KafkaInstrumentation.scala
```diff
@@ -194,12 +207,13 @@ object KafkaInstrumentation {
...
-    val ContextHeader = "kctx"
   }
 
-  case class Settings (
+  case class Settings(
     startTraceOnProducer: Boolean,
     continueTraceOnConsumer: Boolean,
-    useDelayedSpans: Boolean
+    useDelayedSpans: Boolean,
+    propagator: KafkaPropagator,
```
ivantopo, line 216: Please remove this trailing comma so it compiles on earlier Scala versions

ivantopo (changes requested):
Hey @Symbianx, thanks so much for the PR! There is a tiny change required for this to compile on early Scala versions and other than that, it looks good to me.

It would be nice to propagate context tags as [W3C baggage](https://www.w3.org/TR/baggage/) so that they don't get lost at this point, but I think that's something that should be addressed on a separate PR.

instrumentation/kamon-kafka/src/main/scala/kamon/instrumentation/kafka/client/KafkaInstrumentation.scala
Symbianx: 👍  did the change

ivantopo:
Thanks @Symbianx!

Turns out I already opened an issue regarding w3c-baggage: https://github.com/kamon-io/Kamon/issues/803. Let's follow up on that over there.

Symbianx:
Thank you for merging @ivantopo! 🎉  
When can we expect this change to be released? Is there a release cadence we should be aware of?

PR #1250, opened 2023-01-13
Adds source thread to mdc

nvollmar:
We found this very useful when analyzing dispatcher related issues

### DIFF
--- a/instrumentation/kamon-logback/src/main/resources/reference.conf
+++ b/instrumentation/kamon-logback/src/main/resources/reference.conf
@@ -14,6 +14,9 @@ kamon.instrumentation.logback {
     span-id-key = "kamonSpanId"
     span-operation-name-key = "kamonSpanName"
 
+    # MDC key used to store the source thread from which the log is written.
+    source-thread-key = "kamonSourceThread"
+
     # Enables copying of Context information into the MDC. Please note that if you only want to include certain Context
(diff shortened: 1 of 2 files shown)


instrumentation/kamon-logback/src/main/resources/reference.conf
```diff
@@ -13,6 +13,7 @@ kamon.instrumentation.logback {
     trace-id-key = "kamonTraceId"
     span-id-key = "kamonSpanId"
     span-operation-name-key = "kamonSpanName"
+    source-thread-key = "kamonSourceThread"
```
ivantopo, line 16: Hey @nvollmar! Can you please add a separate comment to this setting stating that this MDC key contains the name of the thread where the log event was created? I think the name is obvious, but having it right there next to the span-related keys makes it look as if the thread name won't be available if there isn't a span, but according to the code this will always be there.

ivantopo (changes requested):
Nice and useful contribution @nvollmar! Just a tiny change on the config and we are good to go.

instrumentation/kamon-logback/src/main/resources/reference.conf
nvollmar: Make sense, will add it.

ivantopo (approved):

ivantopo:
Thanks @nvollmar :tada:

PR #1266, opened 2023-03-15
kamon-prometheus: Add default buckets for Dimension.Percentage

jatcwang:


reporters/kamon-prometheus/src/main/resources/reference.conf
```diff
@@ -52,6 +52,16 @@ kamon.prometheus {
       1048576
     ]
 
+    percentage-buckets = [
```
ivantopo, line 55: It saddens me that there is such a wide spread across those buckets but well, that's how it has to be :shrug:

ivantopo (changes requested):
Hey @jatcwang thanks for working on this!

There is a compilation issue on the tests, please see here: https://github.com/kamon-io/Kamon/actions/runs/4429816623/jobs/8037791330#step:6:18336

Other than that it looks good to me.

jatcwang:
@ivantopo Thanks for the review. Can you tell me what's wrong with the latest test failure? Seems like a flaky test and unrelated to my changes? https://github.com/kamon-io/Kamon/actions/runs/4566168676/jobs/8058292358?pr=1266#step:6:4421

ivantopo:
@jatcwang just a random test failure, I'm re-running the tests now but as long as this compiles we should be good to go :smile:

ivantopo (approved):

PR #1280, opened 2023-05-25
 Automatic instrumentation for Scala's global execution context

### LINKED ISSUE #1271: Automatic instrumentation for Scala's global execution context
This came up on a conversation with a customer: can we instrument Scala's execution context? I think we can, and should be a pretty simple thing to do too. 

The use case was basically to ensure that this execution context is NOT used, since most big-ish applications will usually have dedicated thread pools for IO and blocking operations, but there is always a rouge Future here and there that ends up using the global EC. Sometimes it amounts to something, sometimes it doesn't. We should provide enough info for users to figure that out.

solnaranu:
Resolves  #1271

### DIFF
--- a/instrumentation/kamon-executors/src/main/scala/kamon/instrumentation/executor/ScalaGlobalExecutionContextInstrumentation.scala
+++ b/instrumentation/kamon-executors/src/main/scala/kamon/instrumentation/executor/ScalaGlobalExecutionContextInstrumentation.scala
@@ -17,12 +17,6 @@
 package kamon.instrumentation.executor
 
 import kanela.agent.api.instrumentation.InstrumentationBuilder
-
-/**
-  * Captures the current Context upon creation of all Runnable/Callable implementations and sets that Context as current
-  * while their run/call methods are executed. See the module's exclude configuration for more info on what packages and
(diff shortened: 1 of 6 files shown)


instrumentation/kamon-executors/src/main/scala/kamon/instrumentation/executor/ScalaGlobalExecutionContextInstrumentation.scala
```diff
@@ -0,0 +1,30 @@
...
+
+package kamon.instrumentation.executor
+
+import kanela.agent.api.instrumentation.InstrumentationBuilder
+
+/**
+  * Captures the current Context upon creation of all Runnable/Callable implementations and sets that Context as current
+  * while their run/call methods are executed. See the module's exclude configuration for more info on what packages and
+  * implementations will not be targeted by this instrumentation (e.g. it does not target any java.* class by default).
+  */
```
ivantopo, line 25: I think this comment should be removed because it does not apply to this particular instrumentation

ivantopo (changes requested):
Hey Bojan! Just a minor change on the comments and please add the sources for Scala 3 so it compiles and passes tests :smile:

ivantopo (approved):

PR #1289, opened 2023-08-04
+ opentelemetry: Add metrics support

hughsimpson:
Adds support for opentelemetry metrics reporting

~Main point of discussion from my perspective is whether and how to group by 'component' before sending so that that's accurately captured.. generally I think it's a safe assumption that metrics with the same name will be from the same component, so I'm loath to spend the cycles on verifying that, but OTOH maybe I'm wrong? It could perhaps be a feature toggle down the road if necessary...~ Removed that bit and just tagging everything with component=kamon-metrics

There's also maybe a little more refactoring that could have been done to share a bit more of the code logic between traces and metrics, but is it really worth it? I reckon we could save about 50 lines or something at most, and it's not like there a whole whack of other telemetry facets out there waiting to be implemented. I think this is ok...

Further Edit: I haven't implemented any sort of automatic adjustment of the offsets for exponential histograms - the value for positive buckets is currently always 1, and negative always -bucketCount. I _believe_ it should be relatively easy to implement automatic shifting of that ( half max scale minus min scale, adjust scales accordingly?) but haven't done so. I also have questions in my mind about the edge case where the original kamon metric bucket 'covers' two or more target buckets. Started to implement some distribution-of-average logic there but have held myself back for now - it should be easy enough to add, though.

There may be some inconstancy in terms of which bucket we allocate frequencies to at boundaries. I'm personally happy however with whatever error term that might introduce, although I admit to not having given it a thorough analysis 

I don't know whether the spec would advocate for or against it, but it also seems reasonable to me - whether or not support for balancing base offset is added - to drop any leading 0's in the 'negative' bucket (changing its offset accordingly) and any trailing 0's in the 'positive' bucket.


reporters/kamon-opentelemetry/src/main/scala/kamon/otel/MetricsConverter.scala
```diff
@@ -0,0 +1,121 @@
...
+      boundaries += el.value.toDouble
+    }
+    ImmutableHistogramPointData.create(
+      fromNs,
+      toNs,
+      SpanConverter.toAttributes(s.tags),
+      JDouble valueOf s.value.sum.toDouble,
+      JDouble valueOf s.value.min.toDouble,
+      JDouble valueOf s.value.max.toDouble,
+      boundaries.result().dropRight(1).asJava,
```
hughsimpson, line 71: Otel just wants bucket boundaries, so we drop the 'max' here

reporters/kamon-opentelemetry/src/main/scala/kamon/otel/MetricsConverter.scala
```diff
@@ -0,0 +1,121 @@
...
+      counts.result().asJava
+    )
+  }
+
+  def toHistogramData(distributions: Seq[Snapshot[Distribution]]): Option[HistogramData] =
+    distributions.filter(_.value.buckets.nonEmpty) match {
+      case Nil => None
+      case nonEmpty => Some(ImmutableHistogramData.create(AggregationTemporality.CUMULATIVE, nonEmpty.map(toHistogramDatum).asJava))
```
ivantopo, line 79: Shouldn't this and the counter have `AggregationTemporality.DELTA`?

reporters/kamon-opentelemetry/src/main/scala/kamon/otel/MetricsConverter.scala
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+  def convertHistogram(histogram: MetricSnapshot.Distributions): Option[MetricData] =
```
ivantopo, line 82: Have you considered using exponential histograms instead of the plain base histogram here? Exponential histograms would provide a much closer representation of Kamon's HDR histogram data

ivantopo (changes requested):

ivantopo:
Regarding this part:

> Main point of discussion from my perspective is whether and how to group by 'component' before sending so that that's accurately captured.. generally I think it's a safe assumption that metrics with the same name will be from the same component, so I'm loath to spend the cycles on verifying that, but OTOH maybe I'm wrong? It could perhaps be a feature toggle down the road if necessary...

I'm my limited experience with OTel, I'm not really sure this brings any additional value to end users or makes anything easier/better along the pipeline so I would say lets group it all together as "kamon metrics" and then bring grouping per component if we have a clear use case for it. If it comes to that, we might want Kamon itself to provide a bit more information than just the `component`. That's another relic of the opentracing times we should kill :)

reporters/kamon-opentelemetry/src/main/scala/kamon/otel/MetricsConverter.scala
hughsimpson: I had not... Hadn't even looked at it, in fact. Will take a look later on today

reporters/kamon-opentelemetry/src/main/scala/kamon/otel/MetricsConverter.scala
ivantopo: Adding some comments I dropped on Discord for visibility:

Kamon's buckets are just HDR histogram buckets, which are pretty "small" buckets in comparison to what a usual Prometheus-like histogram expects. On Prom systems you would expect to expose 10-20 buckets per histogram, but Kamon's HDR can have thousands of buckets and that buckets iterator will only give you the buckets have values on the current tick, which might not be the same as in the previous tick

we do a bit of work on the prometheus exporter to re-bucket the data into a smaller resolution histogram based on the configured bucket boundaries and I think here we should do something similar BUT instead of doing that, I would rather go with the exponential histogram

I'm not sure if the exponential histograms are already part of the official spec and supported everywhere, but a bit of googling should clear that out

reporters/kamon-opentelemetry/src/main/scala/kamon/otel/MetricsConverter.scala
```diff
@@ -0,0 +1,158 @@
...
+      case Nil => None
+      case nonEmpty =>
+        val mapped = nonEmpty.flatMap { s =>
+          s.value match {
+            case zigZag: Distribution.ZigZagCounts =>
+              logger.error("Unable to construct exponential histogram data - Unimplemented")
+              None
+//              Some(ExponentialHistogramPointData.create(
+//                ???, zigZag.sum, ???, ???, ???, fromNs, toNs, SpanConverter.toAttributes(s.tags), new JArrayList[DoubleExemplarData]()
+//              ))
```
hughsimpson, line 103: Not got around to this yet...

reporters/kamon-opentelemetry/src/main/scala/kamon/otel/MetricsConverter.scala
hughsimpson: Refactored this pr to support `OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION` (in theory)

reporters/kamon-opentelemetry/src/main/scala/kamon/otel/MetricsConverter.scala
ivantopo: I think you probably don't need to understand anything about the ZigZagCounts' implementation to get this going. You will probably need to add a setting for the exponential histogram scale and then go through Kamon's bucketsIterator and map them to the appropriate bucket on the exponential histogram. Sounds weird, but once you get to it you will see it :)

hughsimpson: Ok, I get how that'd work now! Makes sense

PR #1291, opened 2023-08-17
build.sbt: Publish kamon-pekko and kamon-pekko-http artifacts

pronovic:
Thanks for adding support for Pekko and Pekko HTTP.   I'm in the process of converting some of my own code from Akka to Pekko, and I wanted to do some local testing with the new Kamon functionality before v2.7.0 was released.  

However, `build.sbt` does not seem to publish the new artifacts.  I'm not 100% confident that I understand the builld process here, but I think these adjustments are what is needed.  With this in place, I can do `sbt +clean +compile +publishM2` and get the Pekko artifacts in `~/.m2` so I can use them to test my other libraries and services.

If I've misunderstood something here, I'm happy to make changes and resubmit.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -137,6 +137,8 @@ val instrumentationProjects = Seq[ProjectReference](
   `kamon-akka`,
   `kamon-akka-http`,
   `kamon-akka-grpc`,
+  `kamon-pekko`,
+  `kamon-pekko-http`,
   `kamon-play`,
   `kamon-okhttp`,
   `kamon-tapir`,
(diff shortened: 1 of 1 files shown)


pronovic:
@DieBauer just wanted you to be aware of this change.

TjarkoG:
we'd love a release for kamon-pekko

nvollmar:
@ivantopo would be great to get that in

hughsimpson (approved):
lgtm

PR #1292, opened 2023-08-21
+kafka: permit specifying custom propagation implementation

hughsimpson:


yarosman:
relates to

hughsimpson:
@yarosman I think b3 should be pretty similar to the custom implementation I made for the test -- if you're up for implementing it, I'll happily accept a pr to my branch (although I don't need it myself)

yarosman:
@ivantopo What do you think about merging it ?

hughsimpson:
Think Ivan's a bit busy with other stuff at the moment so might take a while to get to this - I've been overriding the class locally in the meantime

hughsimpson:
@yarosman released as 2.6.4, lemme know how it goes

yarosman:
> @yarosman released as 2.6.4, lemme know how it goes

I hope I will try to integrate it on next week ))

PR #1300, opened 2023-10-16
upgrades kanela agent to 1.0.18 which includes support for jdk21

danischroeter:


danischroeter:
implements #1302

pk044:
would be awesome to have this - we want to move to JDK21 and this is the last thing we need to complete the migration.

danischroeter:
@dpsoft @ivantopo would be nice if you guys could merge upgrade to the latest kanela

PR #1301, opened 2023-10-16
Respect include-host configuration in DatadogAPIReporter

rafaavc:
Currently, `DatadogAPIReporter` sends `host` tag information regardless of what's specified in `kamon.datadog.environment-tags.include-host`. By removing it from the request payload, we can rely on `Kamon.environment` and `EnvironmentTags.from` to filter and send the `host` tag while respecting the configuration.

The `DatadogAPIReporter` uses v1 of the [API](https://docs.datadoghq.com/api/latest/metrics/#submit-metrics), which contains the optional `host` field. Sending the host through this field and sending it as a tag seems to result in the same behavior.

An alternative to replacing the `host` field for the tag would be sending that field only when `include-host` is true.

### DIFF
--- a/reporters/kamon-datadog/src/main/scala/kamon/datadog/DatadogAPIReporter.scala
+++ b/reporters/kamon-datadog/src/main/scala/kamon/datadog/DatadogAPIReporter.scala
@@ -16,7 +16,6 @@
 
 package kamon.datadog
 
-import java.lang.StringBuilder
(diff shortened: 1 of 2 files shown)


rafaavc:
Closing this PR since I no longer think this is a good idea. This could have unintended consequences for other users. Furthermore, `kamon.datadog.environment-tags.include-host` should only be applied to the environment tags.

PR #1303, opened 2023-10-17
Use ConfigUtil.splitPath to build environment tag keys from a config path

thyandrecardoso:
Config paths are different from config keys, and their difference is explained in the documentation from `com.typesafe.config.Config`.

The existing implementation uses the config path of kamon environment tags as the keys of a `TagSet`. This has the drawback of quoting some tags when it is not necessary. I am not sure if there a reason to assume a tag key corresponds to Config path, I am assuming there is not. Please correct me if I am wrong.

This PR tries to avoid quoting of tags by using `ConfigUtil.splitPath()` to break and unquote the path.


hughsimpson:
Will look at this tomorrow but seems reasonable

hughsimpson (approved):

PR #1307, opened 2023-10-20
feat: add pekko-grpc instrumentation

inobu:
### Summary
Add instrumentation for pekko-grpc

### Example
I did publishLocal and use it
In our environment, It works like this
スクリーンショット 2023-10-20 18 51 05

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -139,6 +139,7 @@ val instrumentationProjects = Seq[ProjectReference](
   `kamon-akka-grpc`,
   `kamon-pekko`,
   `kamon-pekko-http`,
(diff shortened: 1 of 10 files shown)


inobu:
cc @seglo @linkleonard @dvgica @hughsimpson

dvgica:
@inobu, of that list only Hugh is a maintainer. That said, the rest of us would love to see support for Pekko gRPC. We're currently using an internal fork of https://github.com/nezasa/kamon-akka-grpc with Akka swapped for Pekko.

hughsimpson (approved):
Seems legit 👍

PR #1295, opened 2023-09-04
Release scala-3 versions of kamon-akka, kamon-pekko and kamon-pekko-http

TjarkoG:
release scala3 releases of:
  * kamon-akka
  * kamon-pekko
  * kamon-pekko-http

since Akka.actor >=2.6.17 was released with scala3 we can release them as scala3 packages as well
Pekko was released for scala3 from v1.0.0

PR #1311, opened 2023-11-09
Update Akka, pekko, and pekko-http to support scala 3

hughsimpson:


instrumentation/kamon-pekko/src/main/resources/reference.conf
```diff
@@ -153,7 +153,6 @@ kanela.modules {
       "kamon.instrumentation.pekko.instrumentations.EventStreamInstrumentation",
       "kamon.instrumentation.pekko.instrumentations.ActorRefInstrumentation",
       "kamon.instrumentation.pekko.instrumentations.DispatcherInstrumentation",
-      "kamon.instrumentation.pekko.instrumentations.ActorMonitorInstrumentation",
```
hughsimpson, line 156: This doesn't seem to fix any tests, and breaks things with scala 3

hughsimpson: Actually I've figured this one out. Accepting the argument to the advice as an Envelope causes the scala 3 compiler to do some sort of magic to strip the HasContext mixin out. Mysterious, but resolvable.

TjarkoG:
Hey
I've just now managed to have a look at the kamon-akka project as suggested in #1295 .
sry that i wasn't able to do it earlier but i had some private stress...
When i run the tests locally all tests for:
* kamon-pekko
* kamon-pekko-http
* kamon-akka
are succeeding. Am i missing something or did you already managed to fix the issue?

PR #1319, opened 2024-02-16
Add kamon http4s

vaslabs:
# Moves [kamon-http4s](https://github.com/kamon-io/kamon-http4s) here.

## Noteable events:

- I've followed the same approach as kamon cats io and adapted the same structure as the kamon-http4s project with 2 separate modules.
- Publishing will be identical to the kamon-http4s project, not sure if you want to namespace it

~~- We have left a common configuration because we identified common code, but haven't changed anything from the original source. My suggestion is we release as is and then we can refactor the common code to avoid investigation of what changed for any weird regressions .~~
~~We have added the `Y-partial-unification` flag to the common compilation settings of 2.12, it seemed easier, we can move to http4s only if deemed necessary~~
~~We followed the kamon-akka as a template for the setup, but we have not changed the original kamon-http4s code, only the sbt setup~~

Please let us know if we missed anything, I've tested +test, +publishLocal in the kamon-http4s project

#### Credits and special thanks

This PR is co-authored with @irodotos7 and our time was funded by @vaslabs-ltd


vaslabs:
there's a flaky test, potentially needs some investigation so it doesn't become a problem

vaslabs:
I've addressed the flakiness and changed the approach on structuring the modules + publishing, ready to be reviewed!

build.sbt
```diff
@@ -143,7 +143,9 @@ val instrumentationProjects = Seq[ProjectReference](
   `kamon-lagom`,
   `kamon-finagle`,
   `kamon-aws-sdk`,
-  `kamon-alpakka-kafka`
+  `kamon-alpakka-kafka`,
+  `kamon-http4s-1_0`,
+  `kamon-http4s-0_23`
```
hughsimpson, line 148: I'd add this to `kamon-bundle-dependencies-2-12-and-up` and `kamon-bundle-dependencies-3` too TBH. Seems like we don't currently have a way of including support in the build file for 2.13 bundle and not 2.12 though, so whilst you could also include  `kamon-http4s-1_0` in the scala 3 bundle, don't think we can get it in the 2.13 one. Alas.

hughsimpson (approved):

PR #1324, opened 2024-02-27
Add kamon-opensearch

rgprajeen:
As there is no support for instrumenting OpenSearch 1.x client even though the clients are vastly similar. I have duplicated the instrumentation for ElasticSearch and made the appropriate changes to support the OpenSearch client.

The changes can be summarized as below,
```
Change package names from com.elasticsearch.client to org.opensearch.client
Change all identifiers from elasticsearch to opensearch
Include necessary libraries
```

All test cases are green.


prajeenrg:
Hi @hughsimpson , Could you please review this?

PR #1327, opened 2024-03-02
Added and applied scalafmt

hughsimpson:
I figured maybe it was time to format everything 😅 

I've tried to pick a set of configurations that best reflect the predominant coding style across the project, although there is (inevitably!) a lot of inconsistency in the existing codebase.

As long as this pr is merged by a standard merge (rather than a squash) we _should_ ignore the result of the initial scalafmtAll in git blame, because of the inclusion of `.git-blame-ignore-revs` with a reference to the relevant commit.

I've also added a linting check to the CI jobs to ~catch out the unwary~ preserve this beautiful new world.

Over to you, @ivantopo 😄


Philippus:
Maybe also add `scalafmtSbt` to format the `*.sbt` and `project/*.scala` files.

Issue #1331, opened 2024-03-21
ClassCastException when using Slick and Kamon Jdbc 


alexmihailov:
Kamon version 2.6.1
Kanela version 1.0.17
Slick version 3.3.2

When using Slick and Kamon Jdbc with setting:
`kanela.modules.executor-service-capture-on-submit setup.enabled = true`

I get an exception:
```
java.lang.ClassCastException: class kamon.instrumentation.executor.CaptureContextOnSubmitInstrumentation$ContextAwareRunnable cannot be cast to class slick.util.AsyncExecutor$PrioritizedRunnable (kamon.instrumentation.executor.CaptureContextOnSubmitInstrumentation$ContextAwareRunnable and slick.util.AsyncExecutor$PrioritizedRunnable are in unnamed module of loader 'app')
	at slick.util.ManagedArrayBlockingQueue.offer(ManagedArrayBlockingQueue.scala:17)
	at java.base/java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1347)
	at slick.util.AsyncExecutor$$anon$1$$anon$4.execute(AsyncExecutor.scala:163)
	at scala.concurrent.impl.Promise$Transformation.submitWithValue(Promise.scala:429)
	at scala.concurrent.impl.Promise$DefaultPromise.submitWithValue(Promise.scala:338)
	at scala.concurrent.impl.Promise$DefaultPromise.dispatchOrAddCallbacks(Promise.scala:312)
	at scala.concurrent.impl.Promise$DefaultPromise.map(Promise.scala:182)
	at scala.concurrent.Future$.apply(Future.scala:678)
	at slick.jdbc.JdbcBackend$DatabaseDef.io(JdbcBackend.scala:65)
```
Also, to reproduce the error, it is necessary that HicariCP did not immediately process the tasks, but queued them by calling the java.util.concurrent method.BlockingQueue.offer(E).

I created an example in which I described the error in more detail and reproduced it - https://github.com/alexmihailov/kamon-slick-issue

PR #1333, opened 2024-04-02
Providers for Runnable and Callable wrappers when using Kanela "executor-service-capture-on-submit" module

alexmihailov:
While working with the Kanela  "executor-service-capture-on-submit" module, I caught the ClassCastException, which I described in https://github.com/kamon-io/Kamon/issues/1331. In this PR, I propose a solution to this problem by adding a configurable list of providers that will be able to provide the necessary wrappers over Runnable and Callable. This will help to avoid ClassCastException errors.

### DIFF
--- a/instrumentation/kamon-executors/src/main/java/kamon/instrumentation/executor/CaptureContextOnSubmitInstrumentation.java
+++ b/instrumentation/kamon-executors/src/main/java/kamon/instrumentation/executor/CaptureContextOnSubmitInstrumentation.java
@@ -16,20 +16,35 @@
 
 package kamon.instrumentation.executor;
 
+import com.typesafe.config.Config;
 import kamon.Kamon;
-import kamon.context.Context;
-import kamon.context.Storage.Scope;
 import kamon.instrumentation.executor.CaptureContextOnSubmitAdvices.CallableCollectionWrapperAdvisor;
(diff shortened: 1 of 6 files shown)


hughsimpson:
Sorry for letting this fall by the wayside. Taking a look now...

hughsimpson:
This is causing some issues with `kamon-twitter-future`

alexmihailov:
Hello! Thanks for watching my PR. I will try to fix the problems with the `kamon-twitter-future module`

alexmihailov:
Hi, @hughsimpson! The mistake was that in Scala 3, the `PrioritizedRunnable` interface in the Slick library was sealed. Therefore, I have removed the edits from the `kamon-jdbc` module. In my laptop, all tests for `kamon-jdbc` passed.

alexmihailov:
Thus, it is possible to write providers, and their implementation will be left to the user for the time being. In my case, I am using Scala 2.13 and am able to extend `PrioritizedRunnable`.

hughsimpson (approved):

hughsimpson:
Nice, thanks for fixing the tests

hughsimpson:
I did like the idea of the slick instrumentation, though. Would be good to get that in at some point

alexmihailov:
I will try to do PR with edits for Slick and think about how I can do them for Scala 3.

PR #1347, opened 2024-06-12
Cats-effect-3: Attempt to handle the Dispatcher case

dispalt:
This slightly unfinished, I'd like to make sure the 3.3.x version still works...  I more wanted to highlight the test where I discovered it wasn't working correctly, which is `"must correctly use dispatchers"`


hughsimpson:
I'm very excited about this

PR #1356, opened 2024-08-06
fix apache http client and cxf tests on self hosted runners

ivantopo:
I'll be pushing things to this branch to test what's wrong with the tests


ivantopo:
The issue was that within our self-hosted runner the test containers are not listening on localhost but in a dedicated interface like `172.17.0.1`. The test container gives the right host so the fix was quick. I notice some other build failures related to the Cassandra tests but that is completely separate from this.

PR #1358, opened 2024-08-30
Allow failure log level configuration

thyandrecardoso:
Datadog APIs sometimes give us timeouts or 502s, which seems to [happen to others as well](https://github.com/micrometer-metrics/micrometer/issues/1499).

The API reporter currently logs those failures in `ERROR` level which can be a bit frustrating for transient failures like those.

Here I try to add a configuration mechanism to control that log level. What do you think?


ivantopo:
Thanks @thyandrecardoso!

PR #1360, opened 2024-09-01
Allow configuration of the percentile metrics to submit in Datadog API reporter

thyandrecardoso:
Currently, only p95 gets reported. Looking at p99 seems a fairly common and important use case. I tested this by making sure that these changes work against our Datadog account and I see the new metric.

Can we report p99 as well?

### DIFF
--- a/reporters/kamon-datadog/src/main/resources/reference.conf
+++ b/reporters/kamon-datadog/src/main/resources/reference.conf
@@ -70,6 +70,11 @@ kamon {
     # The log level in which to log failures to submit metrics.
     failure-log-level = "error"
 
+    # For histograms, which percentiles to submit.
+    # Each value configured here will correspond to a different custom metric submitted to Datadog.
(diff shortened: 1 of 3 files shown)


ivantopo:
I think that adding the p99 would be a good addition, but I wouldn't just add it as a default because that one additional metric could have a significant effect on people's bills because they are counted as custom metrics.

Could you please change the implementation so that we read a list of percentiles from config (similar to how the influxdb reporter does here: https://github.com/kamon-io/Kamon/blob/master/reporters/kamon-influxdb/src/main/resources/reference.conf#L16) and default to only the p95 so that it behaves the same after people upgrade?

thyandrecardoso:
Makes sense. Will try to improve this in the upcoming days...

thyandrecardoso:
@ivantopo Hi, can you take a look at this again please?

hughsimpson (approved):
lgtm

Issue #1352, opened 2024-07-25
Nullpointer exception for pekko-instrumentation in version 1.1.0-M1 onward


TjarkoG:
when testing pekko with the version 1.1.0-M1 which will be released soon we've ran into a nullpointer error when running with Kamon + kanela-instrumentation.

to Reproduce:
build.sbt
``` sbt
name := "pekko-quickstart-scala"

version := "1.0"

scalaVersion := "2.13.14"

lazy val pekkoVersion = "1.1.0-M1"

run/fork := true

libraryDependencies ++= Seq(
  "org.apache.pekko" %% "pekko-actor-typed" % pekkoVersion,
  "org.apache.pekko" %% "pekko-http" % pekkoVersion,
  "org.apache.pekko" %% "pekko-stream" % pekkoVersion,
  "ch.qos.logback" % "logback-classic" % "1.5.6",
  "io.kamon" %% "kamon-pekko" % "2.7.3",
  "io.kamon" %% "kamon-system-metrics" % "2.7.3",
  "io.kamon" %% "kamon-prometheus" % "2.7.3"
)
```

Main.scala:
``` scala

//#full-example
package com.example

import com.typesafe.config.ConfigFactory
import org.apache.pekko.actor.ActorSystem
import org.apache.pekko.http.scaladsl.Http
import org.apache.pekko.http.scaladsl.client.RequestBuilding.Get
import org.apache.pekko.http.scaladsl.server.Directives

object Main extends Directives {
  def main(args: Array[String]): Unit = {
    val config = ConfigFactory.load()
    implicit val untyped: ActorSystem = ActorSystem.create("Test", config)
    Http()(untyped).singleRequest(Get("https://www.google.com")).onComplete(rsp => println(rsp.map(_.status)))(untyped.dispatcher)
  }
}
```

when starting with `-javaagent:javaagent/kanela-agent.jar` (version 1.0.18)

i'm receiving the exception:
```
org.apache.pekko.actor.ActorInitializationException: pekko://Test/system/IO-TCP/selectors/$a: exception during creation, root cause message: [Cannot invoke "org.apache.pekko.dispatch.DispatcherPrerequisites.settings()" because the return value of "kamon.instrumentation.pekko.instrumentations.DispatcherInfo$HasDispatcherPrerequisites.dispatcherPrerequisites()" is null]
	at org.apache.pekko.actor.ActorCell.create(ActorCell.scala:679)
[...]

PR #1361, opened 2024-09-10
make kamon-pekko tolerant of null dispatcherPrerequisites

pjfanning:
relates to #1352 

unit tests pass and tested with https://github.com/pjfanning/kamon-pekko-http-test

The NPE may pop up in other use cases but this sorts the one seen in #1352 

I know a fuller fix is needed - one that ensures the dispatcherPrerequisites are found but that can be done independently.


hughsimpson:
I've been spending some time today trying to figure out what's causing this (cf https://github.com/kamon-io/Kamon/pull/1362) and honestly I'm not sure, but it seems like various pieces of instrumentation aren't working... Especially in scala 2.13. Scala 3 doesn't seem to have any issues.. so that's odd. Would rather the instrumentation tests passed for pekko 1.1, because otherwise we're probably hiding errors...

pjfanning:
@hughsimpson feel free to do a new PR that sorts out the instrumentation issue. This PR is useful as far as I am concerned because it stops Kamon from crashing the JVM.

To repeat what I said in the PR description:
I know a fuller fix is needed - one that ensures the dispatcherPrerequisites are found but that can be done independently.

PR #1365, opened 2024-09-19
ThreadPoolConfigCopyAdvice often sets null prereqs

pjfanning:
Found when testing #1352 

Even when testing with Scala 2.12, I see that the set in this PR often sets null
1352 affects Scala 2.13 far more than Scala 2.12.


hughsimpson:
I'm not sure that this would work... the copy will be instantiated with null values on those fields anyway, so not sure this actually defends against anything?

hughsimpson:
I can't see any intervention pathway where they would be possible short of constructor advice pulling it from context, and then how do you get it into the constructor context if not here?

hughsimpson:
I'm gonna close this, I think it's the wrong approach

PR #1367, opened 2024-09-28
add kamon-pekko-connectors-kafka

pjfanning:
based on alpakka-kafka


pjfanning:
CI test failures were in kamon-http4s

hughsimpson:
I'm not even sure this really does much useful, since a context will live directly on the producer record on creation, and that's what's used for propagation (assuming the kamon-kafka instrumentation is switched on). I tried to write a test for this, but couldn't make one that failed without the code. I suppose this should at least fix any mdc fields on logs emitted during a commit, though. Life's too short. Merging.

PR #1368, opened 2024-09-30
add test for context propagation for Pekko scheduled events

pjfanning:
Related to https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/11755

I thought there might be a risk that Kamon might not support context propagation when the Pekko scheduler is used. The test shows that it works ok but I think the test would be useful for regression purposes.


pjfanning:
There is already a test that I missed so this new test is not useful.

PR #1380, opened 2025-02-07
Build on all jdks

ivantopo:


ivantopo:
I somehow thought these changes were never merged :facepalm:

PR #1386, opened 2025-03-04
expose text Content-Type in embedded http server

xela85:
As of today, the Kamon embedded server does not set the `Content-Type` in its response headers. 
This breaks Prometheus 3.0 which requires this header to be properly set : 
https://prometheus.io/docs/prometheus/3.0/migration/#scrape-protocols

To fix this issue, I have modified `SunHttpServerSpecSuite`

### DIFF
--- a/reporters/kamon-prometheus/src/test/scala/kamon/prometheus/EmbeddedHttpServerSpec.scala
+++ b/reporters/kamon-prometheus/src/test/scala/kamon/prometheus/EmbeddedHttpServerSpec.scala
@@ -7,8 +7,9 @@ import org.scalatest.matchers.should.Matchers
 import org.scalatest.wordspec.AnyWordSpec
 
 import java.io.FileNotFoundException
-import java.net.URL
+import java.net.{URL, URLConnection}
 import java.util.zip.GZIPInputStream
+import scala.jdk.CollectionConverters._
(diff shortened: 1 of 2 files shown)


reporters/kamon-prometheus/src/test/scala/kamon/prometheus/EmbeddedHttpServerSpec.scala
```diff
@@ -7,8 +7,9 @@ import org.scalatest.matchers.should.Matchers
 import org.scalatest.wordspec.AnyWordSpec
 
 import java.io.FileNotFoundException
-import java.net.URL
+import java.net.{URL, URLConnection}
 import java.util.zip.GZIPInputStream
+import scala.jdk.CollectionConverters._
```
ivantopo, line 12: This import does not work with earlier Scala versions, but I'm about to merge removing support for Scala 2.11 and 2.12. Probably there is no need to change this but just wait until #1387 gets merged

TidyMaze:
We encountered the same issue, would be great to change that content type.
Thanks guys!

xela85:
Hi! Just wanted to follow up on this PR.  
Is there anything blocking it, or anything I can do to help move it forward?

PR #1387, opened 2025-04-07
upgrade to Kanela 2.0.0-beta.1

ivantopo:
For the most part this is the exact same functionality we had already but now using Kanela 2.0 instead. A few important things to mention:
  - This PR drops support for Scala 2.11 and 2.12
  - This PR also drops support for Akka 2.5 and Play 2.6. Some instrumentation might work but we are not going to be testing it or providing support for it anymore
  - We are now using slf4j 2.0, which for almost everyone will just mean upgrading their logback versions as well. This will only be a problem for people using Spring Boot 2.7 or earlier with Kamon who can try adding the `org.springframework.boot.logging.LoggingSystem=none` system property as we did in our tests, or try to upgrade to Spring Boot 3, or just keep using Kamon 2.7.x. I don't think there will be a lot of effort from our side on the Spring instrumentation because Micrometer already does a much better work with the instrumentation. Any future efforts will probably be focused on interop rather than rewriting instrumentation in Kamon
  - All the `within` and `exclude` patters are now matched using a simple string prefix instead of regex, so they had to be updated all over the place
  - The executors instrumentation no longer has the ability to use custom wrappers for `Runnable/Callable`s submitted to the executors. I'm not sure if anyone ever used this feature but we were not using it in Kamon so it got removed
  - This release still works with Java 8, but it is likely that we will drop support for Java 8 and require at least Java 17 in the soon to come Kamon 3.0


mkurz:
@ivantopo I think you can remove the `upgrade-to-kanela-2` branch now ;)

PR #1391, opened 2025-04-10
Support datadog metrics submission v2 endpoint

thyandrecardoso:
This proposes adding support for version v2 of the metric submission endpoint on the Datadog API reporter. https://docs.datadoghq.com/api/latest/metrics/#submit-metrics

This also makes some other opportunistic changes to the HttpClient class in hopes of making it easier to reason about its logic, and make the reporter configuration more inline with the Span reporter.

My main motivation for these changes was to understand if sending data to that _newer_ Datadog endpoint version has any impact on the sporadic connection timeouts I was seeing while using this reporter (which was also the reason for https://github.com/kamon-io/Kamon/pull/1359).

I added a new configuration knob to indicate if one should use the new or old JSON payload format. Not really sure if there is a better way to do this.


thyandrecardoso:
@hughsimpson any chance you could take a look at this? 🥺

Issue #1392, opened 2025-04-15
BuildInfo kanela agent jar name lacks version


hughsimpson:
Something funny happened on the latest release to cause the BuildInfo to refer to an unversioned kanela agent jar target. It didn't look like this was intentional. I have been unable to reproduce locally but inspection of the jar yields that the info refers to `kanela-agent.jar` instead of `kanela-agent-$version.jar`. Net result is failed instrumentation loads which previously worked.
`kamon/runtime/attacher/BuildInfo$.class`:
v2.7.5:

<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/6dfe7d20-a501-4a9d-9f79-8de5e7f78f6d" />

v2.7.6:

<img width="1508" alt="Image" src="https://github.com/user-attachments/assets/1f4ef709-8c3c-4484-92b9-0a509352965b" />

 Will try to chase up asap...

ivantopo:
A bit more weirdness into this: this only happened with the Scala 2.13 version of the attacher. I see that all others have the right values.

PR #1393, opened 2025-04-16
Throw more helpful error if kanela-agent jar is undiscoverable

hughsimpson:
Also log errors if we somehow manage to find an unversioned kanela-agent jar on the path

I haven't managed to replicate any aspect of https://github.com/kamon-io/Kamon/issues/1392 locally and quite frankly can't really see how it could've happened. Adding some logs to the jar discovery process in case it happens again...


ivantopo:
LGTM. I'm almost sure that the root of the issue has to be connected with me rebuilding all the stuff over and over again while troubleshooting the sonatype upload issue.

PR #1395, opened 2025-05-13
Fix executor instrumentation by unordered config merging

### LINKED ISSUE #1294: Context propagation is broken from Akka 2.8.0 onwards
We have observed when upgrading to Akka 2.8.x that context propagation is not working. We use it for logging context and also http header propagation but neither of these work.
Everything works fine for Akka 2.7.

Forbiddensequence:
Fixes #1294


ivantopo:
Sorry guys, I know this looks like extremely low hanging fruit that should be handled faster. I didn't merge it yet because I remember having a similar issue many year ago and trying to use the `+=` operator but that didn't work out. I vaguely remember having situations where the config wouldn't resolve because of adding to an array that doesn't exist, but that was so long ago and I can't find any traces of it. 

So I thought I'll first try a few things in different types of applications and then merge it if I can't reproduce whatever issue I had in the past. I'll try to do that now while the tests run :D

ivantopo:
I couldn't find any obvious way to break this so we are merging it and hoping it will all be fine :smile:

PR #1397, opened 2025-05-18
pekko-connectors-kafka works on Scala 3

pjfanning:
I think it is omitted because the Akka equivalent does not work on Scala 3 (at least not any the Apache licensed versions)


ivantopo:
Thanks @pjfanning!

Issue #1281, opened 2023-06-05
Trace reporting not working when setting reporter-queue-size to any number different than the default.


vvasilevbosch:
When setting different than default '4096', Tracer.reconfigure is called, which reinstantiates **kamon.trace.Tracer#_spanBuffer** but kamon.trace.Tracer#_onSpanFinish keeps reference to its old value, which is what causes this issue. The **_onSpanFinish** lambda should be updated with the new reference of **_spanBuffer**

https://github.com/kamon-io/Kamon/blob/c3f065517f046f6feffb47652a9ac91d664f160b/core/kamon-core/src/main/scala/kamon/trace/Tracer.scala#L464

` @volatile private var _spanBuffer = new MpscArrayQueue[Span.Finished](_traceReporterQueueSize)
  private val _onSpanFinish: Span.Finished => Unit = _spanBuffer.offer
  //.........
  def reconfigure(newConfig: Config): Unit = synchronized {
    //......
      if(_traceReporterQueueSize != traceReporterQueueSize) {
        // By simply changing the buffer we might be dropping Spans that have not been collected yet by the reporters.
        // Since reconfigures are very unlikely to happen beyond application startup this might not be a problem.
        // If we eventually decide to keep those possible Spans around then we will need to change the queue type to
        // multiple consumer as the reconfiguring thread will need to drain the contents before replacing.
        _spanBuffer = new MpscArrayQueue[Span.Finished](traceReporterQueueSize)
      }
    //......
  }
`

abhinavk96:
@vvasilevbosch did you find a workaround for this?

PR #1399, opened 2025-05-23
fix: Update _onSpanFinish when _spanBuffer is updated

ChrisMcD1:
Without _onSpanFinish updated, then whenever a span is closed, it will will be offered to the old queue, that nothing is reading from anymore. This causes all traces to not be sent.

fixes https://github.com/kamon-io/Kamon/issues/1281


ivantopo:
Hey @ChrisMcD1, great catch!

